PR #71, opened 2018-02-28
Update next example

timneutkens:
Implements pagesExtensions allowing `pages/index.md` to be a valid page.
Also removes the `.babelrc` and uses the same babel-loader config as next itself does :pray:


johnotander:
Awesome! ❤️ ❤️ ❤️

PR #72, opened 2018-03-04
Allow multi-line imports

### LINKED ISSUE #69: Add support for multiple imports in the same paragraph

timneutkens:
Fixes #69

Verified against our app + added to the fixture.

In the future the parsing should probably be done using babel/babylon. But this suffices for now 🙏 

Result:

```
import Foo from './foo'
import Bar from './bar'

text
```

Instead of:

```
import Foo from './foo'

import Bar from './bar'

text
```


johnotander:
Thanks!

PR #85, opened 2018-03-12
Stringify arrays

johno:
When exporting an array, remark-parse parses
them was linkReferences. This ensures they're
properly stringified.


timneutkens:
:100:

PR #96, opened 2018-04-02
[WIP] Implement async api

johno:
In order to allow async plugins we need to use
the async unified api. This will allow plugins
to return promises that are awaited before moving
on to the next.

This is pretty powerful because we can use it at
compile time to do pretty much anything. The example
test checks images for their size and adds that as
props.

This approach can also be used to use image recognition
in order to dynamically populate alt attributes at
compile time.


rauchg:
😍

PR #137, opened 2018-05-07
WIP: Fix Markdown after JSX bug

Flaque:
This is a WIP for issue #136, but to start it adds a failing test to show the issue.

PR #140, opened 2018-05-09
fix(mdxast): render props parse case

### LINKED ISSUE #136: Bug: A header directly after a JSX opening tag will not render as an <h1>
# Overview

Markdown headers don't render correctly when there's no space in between the JSX opening tag.

## Example failure

Take the following code:

```jsx
import Styles from "./styles.js";

<Styles>
# hello, world!

</Styles>
```

This will render without an H1:
bad code

## Example success

If we add a space, like in the following:

```jsx
import Styles from "./styles.js";

<Styles>

# hello, world!

</Styles>
```

This will render an h1 version:

good code

### LINKED ISSUE #139: Bug: working with render props
Isn't possible to reproduce render props component like that:

```
import { Foo } from './Foo'

# Hello Foo

<Foo>
  {() => {
    const bar = 'bar'

    return (
      <Bar bar={bar} />
    )
  }}
</Foo>
```

I noticed that in the `MDAST => MDXAST` process you just turn `HTML` into `JSX` nodes, but in this case the line breaks between `const` and `return` is a trade off, because the parse is a `Paragraph` instead of the same scope of `HTML`.

So, I was trying to figure out how fix this and I think that what can be a possible solution is create something like a merge between nodes when the `HTML` node that has the opening tag is invalid (without enclosing tag). I don't know if this can generate some trade offs or is a solution that fit in all cases. So, I didn't think yet how this can affect performance, but was the only thing that came in mind in that case 😞

pedronauck:
This pr just add [mergeNodeWithoutCloseTag](https://github.com/pedronauck/mdx/blob/03761d59f30571f1df674570bf2c945fccb35415/packages/mdxast/index.js#L124) method in the `mdxast` parser that check if some node has just the element open tag without close it, then merging values into the correct node making a reverse iteration over the tree and deleting incorreting nodes.

I stressed some cases with [a new snapshot test](https://github.com/pedronauck/mdx/blob/03761d59f30571f1df674570bf2c945fccb35415/packages/mdxast/test/test.js#L40) and looking over it is everything going fine!

This closes #139, closes  #136 and closes #137


packages/mdxast/test/test.js
```diff
@@ -30,7 +33,11 @@ const parseFixture = str => {
 }
 
 test('it parses a file', () => {
-  const result = parseFixture(fixture)
+  const result = parseFixture(fixture.basic)
+  expect(result).toMatchSnapshot()
```
timneutkens: Could you use babel to parse the result? It's much more reliable than having a snapshot test that will be guaranteed to fail on every little change.

pedronauck: good, I can do that :v:

timneutkens: 👌

pedronauck:
@timneutkens do you think that just this case is fine or maybe write some more?

wooorm:
@timneutkens It does “block” HTML if it’s a [known HTML block element](https://github.com/remarkjs/remark/blob/e2f4841e6402b4b9fd784cf9a7f35c32da31a538/packages/remark-parse/lib/tokenize/html-block.js#L31). Would it be possible to detect the imported components, and pass their names to the processor as “blocks”?

pedronauck:
I think that doesn't do that because it's just a html, that no reason to use a function inside it like in jsx! But render a component with some render callbacks is something that the community are embracing a lot, limit or create some *workarounds* in think that isn't a good option. IMHO, the better choice would be the language works like expect!

timneutkens:
@pedronauck the thing is that anything inside jsx tags is not some special case, it should just be 1 big jsx block meaning it'd work fine if the parser turns it into that. But I'll have to check the remark code again, specifically the part @wooorm was pointing out, since in that case we don't even have to change that much on our side of the parser 😄

> doesn't parse multi-paragraph because it assumes that document is just a html, that no reason to use a function inside it like in jsx!

Note that remark just says: everything inside html is text, so it has nothing to do with it having to know what the content is, since mdx doesn't parse jsx tags, it just puts it into the final document without touching it.

pedronauck:
I agree in some points, but I think that `mdxast` is assuming the jsx parser work, that is what remark does for html. Assuming this parser position has a lot of trade-offs that you need to consider because jsx is totally different and has a lot of implications that a simple html parser doesn't need to care about. If you Just replace an ast node that is `html` to `jsx` you're assuming they work in the same way, and both we now that isn't the reality.

pedronauck:
Nothing more about that @timneutkens ?

renatorib:
What is missing to merge this PR?

pedronauck:
Close in favor of the new algorithms that [extends remark parser](https://github.com/mdx-js/mdx/blob/jsx-parsing/packages/mdx/index.js#L7-L58)

PR #142, opened 2018-05-16
Add badge to display install size

styfle:
This adds a badge to the `README.md` file to display the npm install size


johno:
I don't think this is necessary since MDX dependencies aren't used during runtime. If anything it could be slightly misleading.

Appreciate the PR though, thanks!

PR #155, opened 2018-06-05
fix(mdx): add "components" prop to wrapper

aputinski:
Currently, the "wrapper" element does not get the "components" prop which prevents it from being customized.

https://github.com/mdx-js/mdx/blob/master/packages/mdx/mdx-hast-to-jsx.js#L27


timneutkens:
You're totally right, this will be fixed in #156. Thank you very much for your contribution! I'd merge this PR first but I've removed ava from @mdx/tag in favor of jest, for consistency between the packages.

PR #156, opened 2018-06-06
add `export default` support

timneutkens:
As per @pranaygp's idea. This PR adds support for layout defined in the mdx file.

- `export default` should export a React component or function

```

some text

export default ({children}) => <div id="some-wrapper">
  {children}
</div>

```


packages/tag/test/mdx-tag.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+  expect(result).toMatch(/style="color:tomato"/)
+})
+
+it('Should render the layout component', async () => {
+  const Layout = ({children}) => <div id="layout">{children}</div>
+  const components = { h1: H1 }
+  const result = renderToString(
+    <MDXTag Layout={Layout} name="wrapper" components={components}>
+      <MDXTag Layout={Layout} name="h1" components={{ h1: H1 }} />
```
johno: Not sure why this line also has a `Layout` prop.

pranaygp:
doing a regex search for `export default` doesn't always work btw @timneutkens:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

Specifically, look at the Syntax examples there for:

```javascript
export { name1 as default, … };

export { default } from …;
```

packages/tag/test/mdx-tag.test.js
timneutkens: Because I copy-pasted the previous line and edited 🤣 Will update

timneutkens: Fixed

timneutkens:
As we talked about on slack yesterday this is not a problem for the initial implementation.

PR #158, opened 2018-06-11
Convert array of class names to a string

devongovett:
For example, when using `@mapbox/rehype-prism`, the `className` property is an array instead of a string, which meant the classes were stringified with commas between them instead of spaces.


timneutkens:
@devongovett actually looking at it again I'm not sure how this works, as we stringify the `props` right before you repurpose the className property. I'm thinking your newly added logic should be above it to work. Can you make another PR? And lets add an integration test for this 🕵️

devongovett:
ah right I copied and pasted wrong. one sec

devongovett:
https://github.com/mdx-js/mdx/pull/159

Issue #162, opened 2018-06-14
Exporting layout component as function fails


l3wi:
When following the `pages/hello` example in `example/next` changing the export from component to function causes the MDX not to render.

### Reproduce
Take the next example:
```javascript
import Layout from '../components/Layout'

export default Layout

# Hello, world!
```
Modify to export as function as described in the docs:
```javascript
import Layout from '../components/Layout'

export default () => <Layout some='metadata' />

# Hello, world!
```

l3wi:
No issue: just incorrect example in readme.

PR #163, opened 2018-06-14
Fix component wrapper example

l3wi:
Fixes example that leads to this: #162


timneutkens:
There's a few more cases in the readme where `export default` is shown incorrectly.

Issue #166, opened 2018-06-22
Using mdx without React


TheKnarf:
```
./node_modules/@mdx-js/tag/dist/mdx-tag.js
  0:0  error  Module not found: Error: Can't resolve 'react' in
              '/Users/fl-macbook-retina/Dropbox/Projects/node_js_server/node_modules/@mdx-js/tag/dist'
```

I wanna use Mdx for a project where I'm not using React but something similar. In my Webpack config I've configured babel the following way to not transform Jsx into `React.createElement` but into just `dom`. Can I get MDX to threat Jsx the same way?

```
{
        loader: 'babel-loader',
        options: {
                presets: ['env', 'react'],
                plugins: [
                        ["transform-react-jsx", {
                                "pragma": "dom" // default pragma is React.createElement
                        }]
                ]
        }
},
```

Seems like it's not currently possible?

I believe it would be if one could make the following line a configuration option:
https://github.com/mdx-js/mdx/blob/master/packages/loader/index.js#L10

TheKnarf:
Seems like it's not currently possible?

I believe it would be if one could make the following line a configuration option:

https://github.com/mdx-js/mdx/blob/master/packages/loader/index.js#L10

timneutkens:
Right, I'm open to a pull request making it an options. Regarding @mdx/tag, that's something you'll have to implement yourself if you're not using React.

naomiHauret:
Facing the same issue, I'm trying to use mdx with Hyperapp.

johno:
This should now work with v1 alpha since it uses a custom pragma-based approach.

TheKnarf:
So how do I change the Pragma?

mdings:
> So how do I change the Pragma?

create a .babelrc file with the following configuration (in the case of Hyperapp which is using h)

```javascript
{
    "presets": [
        "@babel/env",
        "@babel/react"
    ],
    "plugins": [
        [
            "transform-react-jsx",
            {
                "pragma": "h" // default pragma is React.createElement
            }
        ]
    ]
}
```

PR #169, opened 2018-06-23
Work towards issue #166

TheKnarf:
This allows you to change out imports for the mdx-loader.

It doesn't fix everything, as it still forces you to reimplement @mdx-js/tag to get a version that works without React, but it's a step towards that goal.


TheKnarf:
@timneutkens This is an pull request for issue #166.

Does it look good? Anything you'd like me to change before merging it in?

johno:
Thanks for opening up a PR, and apologies for the belated response. Rather than overriding the loader for different use cases, I typically recommend writing a new loader to keep things as simple as possible.

I'd love to document in `examples` or support a loader that can be used outside the React ecosystem if you'd like to pursue that. Going to close this for now and keep the conversation going in #166 where we can discuss the best way to add support for this functionality.

PR #176, opened 2018-06-29
Improve parsing for imports and exports

johno:
Before we relied 100% on remark's default parser which got us to an MVP but doesn't quite cut it due to the added syntax that imports, exports, and JSX. In many areas that syntax directly conflicts with MD syntax.

For example, consider the following export:

```js
export const frontMatter = {
  heroImg: `${process.env.ASSET_PATH}/space-cats.png`,
  editUrl: 'https://github.com/mdx-js/mdx/edit/master/readme.md',
  authors: [
    'Wilma Flintstone',
    'Fred Flintstone'
  ]
}
```

The template string, url, _and_ array conflict with Markdown parsing. This expected/intended behavior from an MD perspective, but not from an ES export. This occurs because the paragraph tokenizer then passes the parsed paragraph text on to inline tokenizers in order to parse links, inline code, emphasis, etc.

So we're now introducing our own [block parsers](https://github.com/remarkjs/remark/tree/master/packages/remark-parse#parserblocktokenizers) that inject themselves before the HTML/paragraph block tokenizers that skip inline tokenization and pass along the raw values for imports/exports. Note that this is still a WIP

Also, I'd like to explicitly thank @wooorm for providing a wonderfully extensible library that allows us to do whacky things 💟.

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -9,7 +9,8 @@ const {
   isImport,
   isExport,
   isExportDefault,
-  BLOCKS_REGEX
+  BLOCKS_REGEX,
+  EMPTY_NEWLINE
 } = require('./util')
 
 const DEFAULT_OPTIONS = {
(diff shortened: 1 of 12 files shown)


ChristopherBiscardi:
It looks like this is something we'd need to [support page queries](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/11) in gatsby-mdx since right now the graphql tagged template literal is getting mangled as an inline-code element.

ChristopherBiscardi:
This totally worked for us :D thanks!

PR #179, opened 2018-07-05
Use create-react-context in @mdx-js/tag

### LINKED ISSUE #178: Suggestion: Use `create-react-context` for 16.0+ compatibility
Currently, mdx-tag requires React 16.3+. I believe this is mostly due to the user of the new `createContext`, a feature landed in React 16.3. For my project, I need to use React 16.2 because Enzyme, the testing library I'm using, doesn't yet support the new features added in 16.3, so I'm using React 16.2 in my project to run my tests.

If mdx swapped out its usage of createContext for the [`create-react-context`](https://github.com/jamiebuilds/create-react-context) library, I'd be able to continue to use 16.2 in my tests.

I'd also be willing to do this work, if this is something you'd accept.

mAAdhaTTah:
Fixes #178.

### DIFF
--- a/packages/tag/package.json
+++ b/packages/tag/package.json
@@ -21,6 +21,7 @@
     "mdx"
   ],
   "dependencies": {
+    "create-react-context": "^0.2.2",
     "prop-types": "^15.6.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


johno:
Thank you!

PR #189, opened 2018-07-15
Layout props

silvenon:
I'm sure that this solution for layout props isn't ideal, but it's a crucial piece of the puzzle for integrating MDX into some tools. The API can always be changed later. Also, this is a breaking change to anyone who decided to use the `components` prop in their MDX file, because now that should become `props.components`.

With this change we'll be able to do stuff like this:

`posts/layout-props.mdx`:
```
import Layout from '../layouts/post'

# Layout Props!

Passing additional props to custom layouts now works!

export default ({ children, timeStamp }) => <Layout timeStamp={timeStamp}>{children}</Layout>
```
`pages/index.js`:
```jsx
import Post from '../posts/layout-props.mdx'

export default () => (
  <Post
    comonents={{}}
    timeStamp={Date.now()}
  />
)
```

Would resolve:

- gatsbyjs/gatsby#312
- nhducit/gatsby-plugin-mdx#13
- nhducit/gatsby-plugin-mdx#14
- nhducit/gatsby-plugin-mdx#19
- silvenon/gatsby-transformer-mdx#1

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -36,7 +36,7 @@ function toJSX(node, parentNode = {}) {
       '\n' +
       exportNodes.map(childNode => toJSX(childNode, node)).join('\n') +
       '\n' +
-      `export default ({components}) => <MDXTag name="wrapper" ${layout ? `Layout={${layout}}` : ''} components={components}>${jsxNodes
+      `export default ({components, ...props}) => <MDXTag name="wrapper" ${layout ? `Layout={${layout}} layoutProps={props}` : ''} components={components}>${jsxNodes
         .map(childNode => toJSX(childNode, node))
         .join('')}</MDXTag>`
     )
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 bundle.js
 .next
 *-error.log
+*-debug.log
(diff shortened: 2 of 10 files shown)


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -29,8 +29,10 @@ function toJSX(node, parentNode = {}) {
       importNodes.map(childNode => toJSX(childNode, node)).join('\n') +
       '\n' +
       exportNodes.map(childNode => toJSX(childNode, node)).join('\n') +
+      (layout ? `function objectWithoutProperties(obj, keys) { var target = {}; for (var i in obj) { if (keys.indexOf(i) >= 0) continue; if (!Object.prototype.hasOwnProperty.call(obj, i)) continue; target[i] = obj[i]; } return target; }` +
+      '\n' : '') +
       '\n' +
-      `export default ({components}) => <MDXTag name="wrapper" ${layout ? `Layout={${layout}}` : ''} components={components}>${jsxNodes
+      `export default (props) => <MDXTag name="wrapper" ${layout ? `Layout={${layout}} layoutProps={objectWithoutProperties(props, ['components'])}` : ''} components={props.components}>${jsxNodes
```
johno: Since the output JSX still requires babel parsing maybe it makes sense to use destructuring?

```js
export default ({ components, ...props }) =>
```

silvenon: Great idea! That way I won't be breaking anything.

silvenon:
Done. I added that test to @mdx-js/runtime tests, I hope that's fine. There's also a @mdx-js/tag test which tests if `layoutProps` are applied to the layout component.

PR #200, opened 2018-07-26
mdx docs and pointer to gatsby-mdx

ChristopherBiscardi:
Added some docs for the gatsby-mdx plugin. I chose to put basic setup here and leave detailed docs in the plugin repo but if you want more here just let me know.


johno:
This is great, thanks!

Issue #143, opened 2018-05-17
Specify plugins options in mdPlugins array?


madiodio:
Given a remark plugin, let's say [`remark-emoji`](https://github.com/rhysd/remark-emoji/): how do I specify his options using the `mdPlugins` array.

Sorry if I'm missing something so obvious.

johno:
Looks like we have a docs bug 😸, I've opened up #144 to improve our documentation for applying plugins to your MDX.

Plugins need to be passed to the MDX loader via webpack options. This will usually live in your `webpack.config.js`:

```js
const images = require('remark-images')
const emoji = require('remark-emoji')

module.exports = {
  module: {
    rules: [
      {
        test: /\.mdx?$/,
        use: [
          {
            loader: 'babel-loader'
          },
          {
            loader: '@mdx-js/loader',
            options: {
              mdPlugins: [images, emoji]
            }
          }
        ]
      }
    ]
  }
}
```

If you're using Next.js, you can specify options in your `next.config.js`:

```js
const images = require('remark-images')
const emoji = require('remark-emoji')

module.exports = {
  pageExtensions: ['js', 'jsx', 'md', 'mdx'],
  webpack: (config, { defaultLoaders }) => {
    config.module.rules.push({
      test: /\.mdx?$/,
      use: [
        defaultLoaders.babel,
        {
          loader: '@mdx-js/loader',
          options: {
            mdPlugins: [images, emoji]
          }
        }
      ]
    })

    return config
  }
}
```

Hope this helps! Let us know if you have any other questions.

madiodio:
@johno Thanks for your prompt reaction. I have already chimed in the examples folder to see how one can pass plugins to the webpack loader options 😅 but it's nice that you added that to the docs. But my issue was specifically do you pass options to the `remark-emoji` plugin like it can be done [here](https://github.com/rhysd/remark-emoji/#options).

johno:
My apologies on misunderstanding you. We don't support providing options yet, unfortunately. Though it's definitely something we should add functionality for.

madiodio:
No worries @johno, it's kind of a edge case but in the meantime I've got to shim trough the code and discover that this line [mdx/packages/mdx/index.js#L24](mdx/packages/mdx/index.js#L24) actually allowed me to specify the plugin options like this

```js
...
options: {
  mdPlugins: [images, emoji],
  padSpaceAfter: true // this is a `remark-emoji` option
}
...
```

[...]

PR #201, opened 2018-07-26
handle [plugin, pluginOptions] syntax for mdPlugins and hastPlugins

revolunet:
Hi,

this handle `[plugin, pluginOptions]` syntax for mdPlugins and hastPlugins

related to https://github.com/mdx-js/mdx/issues/143

```js
const options = {
  mdPlugins: [
    [remark.toc, { heading: "Sommaire" }],
    [
       remark.frontmatter,
       {
         type: "yaml",
         marker: "-"
       }
    ],
    remark.emoji,
    remark.autolinkHeadings,
    remark.slug
  ]
}
```

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -56,7 +56,14 @@ function createMdxAstCompiler(options) {
     .use(esSyntax)
     .use(squeeze, options)
 
-  mdPlugins.forEach(plugin => fn.use(plugin, options))
+  mdPlugins.forEach(plugin => {
+    // handle [plugin, pluginOptions] syntax
+    if (Array.isArray(plugin) && plugin.length > 1) {
+      fn.use(plugin[0], plugin[1])
+    } else {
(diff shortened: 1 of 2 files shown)


packages/mdx/index.js
```diff
@@ -56,7 +56,13 @@ function createMdxAstCompiler(options) {
     .use(esSyntax)
     .use(squeeze, options)
 
-  mdPlugins.forEach(plugin => fn.use(plugin, options))
+  mdPlugins.forEach(plugin => {
+    // handle [plugin, pluginOptions] syntax
+    if (Array.isArray(plugin) && plugin.length > 1) {
```
johno: We should avoid the `return`s and do an if/else:

```js
if (Array.isArray(plugin) && plugin.length > 1) {
  fn.use(plugin[0], plugin[1])
} else {
  fn.use(plugin, options)
}
```

PR #202, opened 2018-07-27
Add example with razzle

jeetiss:
this example should work after @jaredpalmer publish plugin on npm


johno:
Thanks for the PR, looks great! Do you mind running prettier as a separate PR to keep things isolated?

jeetiss:
@johno done 🙌🏻

johno:
Thanks!

PR #204, opened 2018-07-28
Fix tests

jeetiss:
change test environment to node

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -26,5 +26,8 @@
     "@mdx-js/mdx": "^0.15.0-1",
     "@mdx-js/tag": "^0.15.0-1",
     "buble": "^0.19.3"
+  },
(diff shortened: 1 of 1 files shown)


packages/runtime/package.json
```diff
@@ -26,5 +26,8 @@
     "@mdx-js/mdx": "^0.15.0-1",
     "@mdx-js/tag": "^0.15.0-1",
     "buble": "^0.19.3"
+  },
+  "jest": {
+    "testEnvironment": "node"
   }
-}
+}
```
silvenon: I suggest setting your text editor to append newline at the end.

johno:
Thanks @jeetiss!

> This fixes the tests, but why? I don't understand how they got broken in the first place.

Yeah this had me scratching my head, too. 🤔

Issue #205, opened 2018-07-28
Export syntax creates duplicate default exports


jxnblk:
Using the following syntax for layouts causes a syntax error (`SyntaxError: Only one default export allowed per module.`):

```js
export { default } from './Layout'
```

silvenon:
Just to expand upon this a bit.

The code snippet above is a way to make a default export using named exports. There are other variants of this, like:

```js
// exporting directly from a module
export { Layout as default } from './components'
```
```js
// exporting a variable
import Layout from './components'
export { Layout as default }
```
```js
// bundled with other exports
export { Layout as default, meta } from './post-components'
```

etc. AFAICT, this isn't a simple regex fix because we need convert things like this:

```js
export { default } from './Layout'
```

into something like this:

```jsx
import Layout from './Layout'

export default ({ components, ...props }) =>
  <MDXTag name="wrapper" layout={Layout} ...
```

So we need to pick a name which doesn't collide with existing variable names etc. I think we need to bring in Babel parser for this.

silvenon:
As a maintainer I'm against is as well, but I can see why people would expect this to work. I can submit a PR which raises an error upon encountering this syntax.

jxnblk:
I agree with keeping Babel out of mdx. I mostly just wanted to share that I noticed that that particular syntax doesn't seem to work

PR #209, opened 2018-07-31
Implement option to opt out of export wrapper

johno:
In certain cases, like mdx-js/runtime or mdx-deck
we want to avoid the wrapping export default so that
it doesn't need to be manually stripped off.


silvenon:
Will this fix #205?

johno:
It won't unfortunately, we will need to update the export default regex to also account for this type of export.

PR #213, opened 2018-08-06
fix create-react-app example

kentcdodds:
I tried this out in a devtip video: https://www.youtube.com/watch?v=d2sQiI5NFAM&list=PLV5CVI1eNcJgCrPH_e6d57KRUTiDZgs0u

Some deps were outdated/missing so I added those to make it work. I also made the example cooler :)

screen shot 2018-08-06 at 9 43 32 am


johno:
Thanks!

PR #215, opened 2018-08-08
Fix importing .md files.

treshugart:
In the README under [markdown file transclusion](https://github.com/mdx-js/mdx#markdown-file-transclusion) it says:

> You can transclude Markdown files by importing one .md file into another:

This doesn't work, though, because you need to add the extension and asset type handler, explicitly, to the Parcel bundler instance.


johno:
Thanks!

Issue #221, opened 2018-08-12
New lines on code blocks get stripped when using rehype-prism


hugmanrique:
New lines sometimes get stripped from the [`@mapbox/rehype-prism`](https://github.com/mapbox/rehype-prism) output when highlighting a code block.

I'm using the following Webpack loader:

```js
{
  test: /\.mdx$/,
  include: path.resolve(__dirname, 'posts'),
  loader: [
    'babel-loader',
    {
      loader: '@mdx-js/loader',
      options: { hastPlugins: [[rehypePrism, { ignoreMissing: true }]] }
    }
  ]
}
```

The following `.mdx` file:

```
\`\`\`dockerfile
# Add main script
COPY start.sh /home/start.sh
COPY build-log.xml /home/playpen/

# GitLab CI doesn't have access to /home
RUN chmod -R 700 /home/
\`\`\`
```

And here's the result:
![mdx-loader output](https://i.imgur.com/ddinosR.png)

React source (produced by `@mdx-js/loader`): https://paste.ubuntu.com/p/C3cyx5ZYjV/

Which React renders as:

```html
<pre><code class="language-dockerfile"><span class="token comment"># Add main script</span><span class="token keyword">COPY</span> start.sh /home/start.sh
<span class="token keyword">COPY</span> build<span class="token punctuation">-</span>log.xml /home/playpen/

<span class="token comment"># GitLab CI doesn't have access to /home</span><span class="token keyword">RUN</span> chmod <span class="token punctuation">-</span>R 700 /home/
</code></pre>
```

As you can see, there should be a line break between the first `.token .comment` (`# Add main script`) and the following `.token .keyword` (`COPY`).

One of the reasons why this could be happening is the `@mdx-js/loader` output is an array of react components, so newlines produced by rehype plugins might not be preserved. I haven't looked to much into it, so I might be wrong.

I also didn't know whether to report this on the [`rehype-prism`](https://github.com/mapbox/rehype-prism) repo or here, but I couldn't reproduce it with a barebones remark setup.

silvenon:
Hey! I've been away on holidays for a while so I need some time to catch up, but I just wanted to say thanks for being so excellent at reporting issues! ⭐️

hugmanrique:
Thank you! Also, don't worry about it, this is for a personal project so I'm not on a deadline 😀

hugmanrique:
Thank you for the quick fix! 😄

phil-lgr:
[...]

PR #180, opened 2018-07-06
fix(tables) remove new line backtick wrapping

ifyoumakeit:
Removes warning...
```bash
Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <thead>. Make sure you don't have any extra whitespace between tags on each line of your source code.
```
- Remove wrap of new lines
- Add table to fixture
- Create test to fail on single wrapped new lines
- Update test with wrapped line and extra 's'
- Add table to next.js example

closes mdx-js/mdx#151

PR #226, opened 2018-08-18
Preserve whitespace in code blocks

silvenon:
I'm trying to write a test that reproduces #221. The idea is to mount the component exported by the fixture and use `textContent` to check if whitespace in the code block is preserved.

I'm having difficulties trying to load the module compiled by webpack as a React component and mounting it. Webpack gives me an object with a source text, but I'm not sure how to parse it as a Node module. If someone has an idea, let me know. 😅 

P. S. I'm not 100% certain that this is really an MDX issue, but the fact that I can't reproduce it without it suggests that it is.

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -1,4 +1,9 @@
 function toJSX(node, parentNode = {}, options = {}) {
+  const {
+    // default options
+    skipExport = false,
+    preserveNewlines = false,
+  } = options
   let children = ''
 
   if (node.type === 'root') {
@@ -36,7 +41,7 @@ function toJSX(node, parentNode = {}, options = {}) {
       '\n' +
       exportNodes.map(childNode => toJSX(childNode, node)).join('\n') +
       '\n' +
-      (options.skipExport ? '' : 'export default ({components, ...props}) => ') +
+      (skipExport ? '' : 'export default ({components, ...props}) => ') +
       `<MDXTag name="wrapper" ${
         layout ? `Layout={${layout}} layoutProps={props}` : ''
       } components={components}>${jsxNodes
@@ -47,7 +52,14 @@ function toJSX(node, parentNode = {}, options = {}) {
 
   // recursively walk through children
   if (node.children) {
-    children = node.children.map(childNode => toJSX(childNode, node)).join('')
+    children = node.children.map(childNode => {
+      const childOptions = {
+        ...options,
+        // tell all children inside <pre> tags to preserve newlines as text nodes
+        preserveNewlines: preserveNewlines || node.tagName === 'pre',
+      }
+      return toJSX(childNode, node, childOptions)
+    }).join('')
   }
 
   if (node.type === 'element') {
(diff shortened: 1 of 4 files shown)


silvenon:
I realized I could test this easily with runtime, so nvm, I think I got it. 😉

silvenon:
Ok, I managed to reproduce this in a test. 🎉

hugmanrique:
So would moving the prism plugin to compilers fix the issue? What's the main difference between mdPlugins and this new option?

silvenon:
`hastPlugins` operate on rehype tree, while `compilers` operate on JSX strings. Notice that between applying these two sets of plugins the tree is converted to JSX. `@mapbox/rehype-prism` can only work on rehype, so I think it has to stay in `hastPlugins`.

silvenon:
I (hopefully) figured it out! In the process of converting to JSX there was an exception with newlines introduced by #180. I added an exception to that exception (exceptionception) and converted newlines to JSX text nodes only for code inside `<pre>` tags.

@ifyoumakeit because you have a deeper understanding of that part of the code, could you take a look at this?

ifyoumakeit:
I can definitely can take a look!

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -47,7 +52,13 @@ function toJSX(node, parentNode = {}, options = {}) {
 
   // recursively walk through children
   if (node.children) {
-    children = node.children.map(childNode => toJSX(childNode, node)).join('')
+    children = node.children.map(childNode => {
+      const childOptions = {
+        ...options,
+        preserveNewlines: preserveNewlines || node.tagName === 'pre',
```
ifyoumakeit: 👍 I think this makes sense, as long as there's information about this default behavior somewhere.

alexkrolick:
Nice, this was super annoying: https://github.com/alexkrolick/mdx-observable/pull/1

silvenon:
@ifyoumakeit thanks a lot! I’ll make sure to at least add comments.

@alexkrolick could you be more specific? What problem would these changes solve for you?

alexkrolick:
Actually I might be me misinterpreting the problem. I thought this would fix the fact that newlines inside render functions are interpreted as MD blocks instead of as JS code to evaluate. You can see how the formatting in my link requires adding comments or semicolons anywhere you want a blank line.

silvenon:
This PR unfortunately won't take care of that, it's for fenced code blocks only. We're painfully aware of the issue you mentioned, though.

silvenon:
Done! Polished it up a bit and added more comments to clarify what's going on.

@johno ready when you are. 😉

johno:
Thanks all, this is amazing!

packages/mdx/test/index.test.js
```diff
@@ -97,7 +103,7 @@ it('Should render elements without wrapping blank new lines', async () => {
   | :--- | :---- |
   | Col1 | Col2  |`)
 
-  expect(result.includes('{`\n`}')).toBe(false)
+  expect(result).not.toContain('{`\n`}')
```
ifyoumakeit: 💅

silvenon: tenor

devongovett:
Can we get this released?

johno:
Now published as `0.15.1`, apologies on the delay.

PR #234, opened 2018-08-23
Absolute path to the @mdx-js/tag

jeetiss:
The `@mdx-js/tag` dependency is not resolving if loader used with such builders as `razzle`, `next` etc.
I've added the `absolutePath` option to loader to make it able to resolve `@mdx-js/tag` in this case.

### DIFF
--- a/packages/loader/index.js
+++ b/packages/loader/index.js
@@ -3,13 +3,19 @@ const mdx = require('@mdx-js/mdx')
 
 module.exports = async function(content) {
   const callback = this.async()
-  const options = getOptions(this)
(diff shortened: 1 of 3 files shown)


timneutkens:
It works fine with Next.js though. Maybe razzle doesn't add the extra node_modules directory to the `resolve` key in webpack.

jeetiss:
@timneutkens yep :(
can you review my pr to the razzle?
https://github.com/jaredpalmer/razzle/pull/732

timneutkens:
👍 great!

PR #249, opened 2018-09-03
Add filename info to mdx loader

CompuIves:
For a rehype plugin I'm working on I need the path of the current file being transpiled, I use it to resolve imports. This adds the filename info to the `process` function which allows plugins to access it.

### DIFF
--- a/packages/loader/index.js
+++ b/packages/loader/index.js
@@ -3,9 +3,9 @@ const mdx = require('@mdx-js/mdx')
 
 module.exports = async function(content) {
   const callback = this.async()
-  const options = getOptions(this)
+  const options = Object.assign({}, getOptions(this), {filepath: this.resourcePath});
 
-  const result = await mdx(content, options || {})
+  const result = await mdx(content, options)
 
   const code = `
   import React from 'react'
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -103,7 +103,12 @@ function sync(mdx, options) {
   const opts = Object.assign({}, DEFAULT_OPTIONS, options)
   const compiler = createCompiler(opts)
 
-  const { contents } = compiler.processSync(mdx)
+  const fileOpts = { contents: mdx };
+  if (opts.filepath) {
+    fileOpts.path = opts.filepath;
+  }
+
+  const { contents } = compiler.processSync(fileOpts)
 
   return contents
 }
@@ -112,7 +117,12 @@ async function compile(mdx, options = {}) {
   const opts = Object.assign({}, DEFAULT_OPTIONS, options)
   const compiler = createCompiler(opts)
 
-  const { contents } = await compiler.process(mdx)
+  const fileOpts = { contents: mdx };
+  if (opts.filepath) {
+    fileOpts.path = opts.filepath;
+  }
+
+  const { contents } = await compiler.process(fileOpts)
 
   return contents
 }
--- a/packages/mdx/test/index.test.js
+++ b/packages/mdx/test/index.test.js
@@ -122,6 +122,22 @@ test('Should await and render async plugins', async () => {
   expect(result).toMatch(/HELLO, WORLD!/)
 })
 
+test('Should process filepath and pass it to the plugins', async () => {
+  const result = await mdx(fixtureBlogPost, {
+    filepath: 'hello.mdx',
+    hastPlugins: [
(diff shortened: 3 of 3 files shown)


packages/loader/index.js
```diff
@@ -5,7 +5,12 @@ module.exports = async function(content) {
   const callback = this.async()
   const options = getOptions(this)
 
-  const result = await mdx(content, options || {})
+  const filename = this.resourcePath;
```
wooorm: For the name, I’d personally suggest `options.filePath`, as you’re setting `file.path`, but it doesn’t really matter!

Re the style, maybe use this, looks cleaner to me:

```js
options = Object.assign({}, getOptions(this), {filename: this.resourcePath})
```

Also: is there a specific cwd somehow set by webpack? we could use that for [file.cwd](https://github.com/vfile/vfile#vfilecwd)

packages/mdx/index.js
```diff
@@ -103,7 +104,13 @@ function sync(mdx, options) {
   const opts = Object.assign({}, DEFAULT_OPTIONS, options)
   const compiler = createCompiler(opts)
 
-  const { contents } = compiler.processSync(mdx)
+  const fileOpts = { contents: mdx };
+  if (opts.filename) {
```
wooorm: Is there ever no filename (or `resourcePath` in the loader)?

packages/mdx/index.js
```diff
@@ -103,7 +104,13 @@ function sync(mdx, options) {
...
+    fileOpts.path = opts.filename;
+  }
+  const file = vfile(fileOpts);
```
wooorm: On thing that may make this cleaner, and remove the `vfile` dependency, is that you pass `fileOpts` directly into `.process` / `.processSync`, no need to wrap yourself!

packages/loader/index.js
CompuIves: Ohh that is much cleaner! The reason that I put it in an `if` was that when I tested it with `filename: undefined` it hanged for some reason. I can test a bit more, maybe it was something else.

I also like `filepath` much more, I will change it!

packages/mdx/index.js
CompuIves: I was expecting that `mdx` is not always called from the webpack loader, so I wanted to make the `filename` optional, a bit like [`babel-core` does](https://www.npmjs.com/package/babel-core#options).

packages/mdx/index.js
CompuIves: Hmmmmm that's true, that would make the API not backward compatible though because it would need to be wrapped in `{content}`. It feels a bit less straightforward, but I can change it!

CompuIves:
Bedankt voor de review!

I'll process the comments this evening 😄

packages/mdx/index.js
wooorm: totes, thought we were still in the loader!

packages/mdx/index.js
wooorm: would it not? `.process` and `.processSync` accept both, and create a vfile from either? Maybe I’m missing something!

CompuIves: Whoops, you're right!

CompuIves:
Hey! I processed the parameter to `filepath`, but when I changed the `fileOpts` to not use `vfile` the test seemed to fail. I'm not sure if it takes the path info if it's not a `vfile`. I put the `vfile` constructor back and it seemed to work.

packages/mdx/index.js
```diff
@@ -103,7 +103,12 @@ function sync(mdx, options) {
   const opts = Object.assign({}, DEFAULT_OPTIONS, options)
   const compiler = createCompiler(opts)
 
-  const { contents } = compiler.processSync(mdx)
+  const fileOpts = { contents: mdx };
```
wooorm: This looks amazing! You *could* pass {contents: mdx, path: opts.filepath}, but that's just nitpicking.

Looks great, lekker bezig!

CompuIves: Hmm yes, I didn't do it in this case since the mdx transpiler didn't work for me when `path: undefined`. I can still change it since that's probably something that won't often happen, but this one is more correct methinks.

Also, dankjewel :D! Leuk om zoveel Nederlanders hier te zien!

johno:
Thanks, this looks great!

Issue #141, opened 2018-05-14
Passing props to a custom code block


alanbsmith:
Hey there!

I finally got to dig into this lib this weekend, and it's amazing! I really love how it's designed. Thanks for creating and maintaining it. 😄

I looked through the open and closed issues and saw some related topics, but not anything identical. But I apologize if this is a repeat issue I missed.

I'm wondering if it's possible pass props to a custom code block component. I'm using this lib with [`react-live`](https://github.com/FormidableLabs/react-live) which is great for making my code blocks editable.

I was thinking it would be nice to have an API similar to [`react-styleguidist`](https://react-styleguidist.js.org/docs/documenting.html#usage-examples-and-readme-files)

````jsx
```jsx { props: { code: code, scope: { Button, styled } } 
<Button>Hello, World!</Button>
```
````

Currently, I'm achieving this with the following:

* Passing a custom Demo component to the MDX component:

```jsx
<MDX components={{ code: Demo }} />
```
```jsx
// Demo.js
...
class Demo extends Component {
  render() {
    return (
      <LiveProvider code={this.props.code} scope={this.props.scope} noInline={true}>
        <Editor />
        <LiveError />
        <LivePreview />
      </LiveProvider>
    )
  }
}
...
```

* Creating a separate raw file for my code examples:

```jsx
function Button(props) {
  return  <button>{props.children}<button>;
}

render(<Button />)
```

* And then importing it into my `.md` file:

```jsx
import styled from 'styled-components';
import Button from './Button';
import code from './Demo/Button.example';
import Demo from './Demo';

# Button

## Description

This is our general-purpose Button component.

## Demo

<Demo code={code} scope={{styled, Button}} />
```

And that's fine, but it's a lot of steps. If there's a way to currently do this, that would be great! If not, I'd be happy to work on it if that's something you'd like to support. I reviewed the source code, but I didn't see anything immediately pop out at me for where to add that logic. But I only started looking this weekend. If you can tell me where to start, I'd be happy to take that on. 😄

I know you're really busy, so thanks for your time and feedback! And thanks again for this lib!

johno:
One approach I've been using for a project is frontmatter inside code blocks:

```jsx
---
responsive: true
---

<Box>Hello!</Box>
```

[...]

PR #257, opened 2018-09-12
enable infostring props

ChristopherBiscardi:
This PR enables the passing of the full infostring content as props to react components, enabling things like repls to be enabled as such:

carbon 2

The PR mostly copies from the underlying [mdast handler](https://github.com/syntax-tree/mdast-util-to-hast/blob/1f8591e0b6163c437999e842774f66c9027cfc47/lib/handlers/code.js), but adds basic handling for potential meta strings. I think the handling is "good enough" right now even though it has some quirks like requiring a single space between infostring parameters. We can expand the algorithm in the future if things like array support are desired.

based on the conversation in https://github.com/syntax-tree/mdast-util-to-hast/issues/24

closes https://github.com/mdx-js/mdx/issues/141 and https://github.com/ChristopherBiscardi/gatsby-mdx/issues/76

### DIFF
--- a/packages/mdx/mdx-ast-to-mdx-hast.js
+++ b/packages/mdx/mdx-ast-to-mdx-hast.js
@@ -1,4 +1,6 @@
 const toHAST = require('mdast-util-to-hast')
+const detab = require('detab')
+const u = require('unist-builder')
 
 module.exports = function mdxAstToMdxHast() {
   return (tree, file) => {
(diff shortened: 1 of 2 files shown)


ChristopherBiscardi:
node 8 failed. Maybe because of Object.entries? I could replace it with something else if necessary.

packages/mdx/mdx-ast-to-mdx-hast.js
```diff
@@ -18,6 +20,38 @@ module.exports = function mdxAstToMdxHast() {
           ]
         })
       },
+      code(h, node) {
+        var value = node.value ? detab(node.value + '\n') : ''
```
johno: Can we use `const` here?

packages/mdx/mdx-ast-to-mdx-hast.js
```diff
@@ -18,6 +20,38 @@ module.exports = function mdxAstToMdxHast() {
...
+        var lang = node.lang && node.lang.match(/^[^ \t]+(?=[ \t]|$)/)
```
johno: Let's make this regex a variable.

packages/mdx/mdx-ast-to-mdx-hast.js
```diff
@@ -18,6 +20,38 @@ module.exports = function mdxAstToMdxHast() {
...
+              acc[t[0]] = t[1]
+              return acc
+            } else {
+              acc[cur] = true
+              return acc
+            }
+          }, {})
+
+        meta &&
+          Object.entries(meta).forEach(([key, value]) => (props[key] = value))
```
johno: Not sure we should drop node 8 support until we go v1, maybe we can make this a util or use `Object.keys`?

PR #260, opened 2018-09-15
Fix support for comments

silvenon:
Previous implementation of comments was simply converting all `<!--` and
`-->` to `{/*` and `*/}`. This was very naive and caused a lot of edge
cases. It also supported nesting Markdown comments in JSX, which isn't
how regular Markdown works; nesting comments inside regular HTML doesn't
work, so it doesn't make sense that it works here. So that part is a breaking change.

This fix simply takes advantage of remark's support for comments, which
takes care of all of the edge cases and doesn't reinvent the wheel.

### DIFF
--- a/packages/mdx/md-ast-to-mdx-ast.js
+++ b/packages/mdx/md-ast-to-mdx-ast.js
@@ -2,7 +2,14 @@ const visit = require('unist-util-visit')
 
 module.exports = options => tree => {
   visit(tree, 'html', node => {
-    node.type = node.mdxType || 'jsx'
+    if (
+      node.value.startsWith('<!--') &&
+      node.value.endsWith('-->')
+    ) {
+      node.type = 'comment'
+    } else {
+      node.type = node.mdxType || 'jsx'
+    }
   })
 
   return tree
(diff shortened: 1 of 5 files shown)


packages/mdx/md-ast-to-mdx-ast.js
```diff
@@ -2,7 +2,14 @@ const visit = require('unist-util-visit')
 
 module.exports = options => tree => {
   visit(tree, 'html', node => {
-    node.type = node.mdxType || 'jsx'
+    if (
+      node.value.startsWith('<!--') &&
+      node.value.endsWith('-->')
```
silvenon: Remark contains all comments as separate `html` nodes.

packages/mdx/test/fixtures/blog-post.md
```diff
@@ -20,7 +20,6 @@ I'm an awesome paragraph.
   <Bar>hi</Bar>
     {hello}
     {/* another commment */}
-    <!-- one more comment -->
```
silvenon: I removed this line because it causes an error when Babel tries to parse it.

packages/mdx/test/index.test.js
```diff
@@ -86,20 +86,39 @@ COPY start.sh /home/start.sh
...
+
 \`\`\`md
-<!-- a code block Markdown comment -->
+<!-- a code block string -->
 \`\`\`
+
 <div>
   {/* a nested JSX comment */}
-  <!-- a nested Markdown comment -->
+  <!-- div content -->
```
silvenon: This would technically cause a Babel error, but I included it in the test to ensure that it doesn't turn into a comment.

Issue #145, opened 2018-05-17
Problem using remark-math with MDX


KrofDrakula:
I've been trying to plug in the KaTeX renderer into the chain of plugins for MDX, but so far, I'm stuck with with some of the formulae incorrectly rendering.

My webpack `rules` part is as follows:

```
{
  test: /\.mdx$/,
  use: [
    ...babelLoader,
    {
      loader: '@mdx-js/loader',
      options: {
        mdPlugins: [math],
        hastPlugins: [katex]
      }
    }
  ]
}
```

...where `math` and `katex` are the packages `remark-math` and `rehype-katex`, respectively. I'm using `preact` and `preact-compat` with the following aliases:

```
alias: {
  react: 'preact-compat/dist/preact-compat',
  'react-dom': 'preact-compat/dist/preact-compat',
  'react-addons-create-fragment': 'preact-compat/dist/preact-compat'
}
```

To illustrate the problem, I'm trying to render the following page:

```
import 'katex/dist/katex.css';
import { h } from 'preact';

## Math test

$$
\sum{1}
$$

$$
\sum{i}
$$
```

The first equation renders fine, since `1` is a literal and is just converted to a string in the JSX. However, the second equation uses `{i}` in the rendered JSX and tries to evaluate the `i`, which fails because `i` is `undefined`.

Is there a way to have JSX be ignored or forced to escape in the math portion of the MDX?

linonetwo:
I have
```
The `style` prop expects a mapping from style properties to values, not a string. For example, style={{marginRight: spacing + 'em'}} when using JSX.
    in span (created by MDXTag)
    in MDXTag (created by Context.Consumer)
    in Unknown (at hw1.md:8)
    in span (created by MDXTag)
    in MDXTag (created by Context.Consumer)
    in Unknown (at hw1.md:8)
    in span (created by MDXTag)
    in MDXTag (created by Context.Consumer)
    in Unknown (at hw1.md:8)
```

ChristopherBiscardi:
So this is because the remark plugins write out html into the AST which then tries to get used as JSX at runtime because they're all html nodes iirc.

[...]

PR #261, opened 2018-09-17
Property adjustments for "style", "aria-*" and "data-*"

silvenon:
Some plugins like [rehype-katex](https://github.com/Rokt33r/remark-math) add properties like `style` and `aria-hidden`, which need to be adjusted for JSX. This commit introduces the following changes:

  - `style` strings now turn into camelized objects
  - `aria-*` and `data-*` properties now have keys in param-case

After these changes I can run the example in #145, but I have only tried with React, not Preact. However, the issue that this solves isn't the from the OP, but rather https://github.com/mdx-js/mdx/issues/145#issuecomment-421941974.

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -1,3 +1,6 @@
+const toStyleObject = require('to-style').object
+const { paramCase } = require('change-case')
+
 function toJSX(node, parentNode = {}, options = {}) {
   const {
     // default options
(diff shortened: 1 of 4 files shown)


linonetwo:
Seems to have errors in SSR?

silvenon:
@linonetwo could you be more specific?

silvenon:
I set the testing environment for `mdx` package to `node` and tests still pass. If you were referring to now deployment, that failure is unrelated to this PR.

PR #264, opened 2018-09-18
Use ZEIT now's alias feature

johno:
https://zeit.co/blog/every-push-now


johno:
It's an upstream issue with `x0` + `styled-system` due to conflicting versions. Will get it fixed this evening when I have a nice slice of time.

PR #275, opened 2018-09-26
Add GitHub repo link to docs

pshrmn:
Adds a link to the repo from the documentation website.

I wanted to add it to the sidebar, but that broke `x0`'s pagination, so the about page was the next best solution.


silvenon:
Thanks!

PR #276, opened 2018-09-26
Make @mdx-js/mdx a dep of @mdx-js/loader

pshrmn:
`@mdx-js/loader` requires `@mdx-js/mdx`, but because it is a dev dependency, `npm install @mdx-js/loader` does not install it.


silvenon:
Looks good! Thanks!

PR #283, opened 2018-10-02
Don't use spread operator

### LINKED ISSUE #281: error on simply import runtime in create-react-app v1
`import MDX from "@mdx-js/runtime";`

then I get, 0.15.4 and tried 0.15.0

```
./node_modules/@mdx-js/mdx/mdx-hast-to-jsx.js
Module parse failed: Unexpected token (21:6)
You may need an appropriate loader to handle this file type.
|     const paramCaseRe = /^(aria[A-Z])|(data[A-Z])/
|     node.properties = Object.entries(node.properties).reduce((properties, [key, value]) => ({
|       ...properties,
|       [paramCaseRe.test(key) ? paramCase(key) : key]: value,
|     }), {})
```

silvenon:
Importing @mdx-js/runtime in create-react-app v1 fails because of spread operators.

I didn't dive deep enough to understand why, though.

Fixes #281.

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -17,10 +17,11 @@ function toJSX(node, parentNode = {}, options = {}) {
     // ariaProperty => aria-property
     // dataProperty => data-property
     const paramCaseRe = /^(aria[A-Z])|(data[A-Z])/
-    node.properties = Object.entries(node.properties).reduce((properties, [key, value]) => ({
-      ...properties,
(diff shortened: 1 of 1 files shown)


johno:
I just realized we should figure out what node version is being run, this shouldn't be required 🤔

silvenon:
I was running the example in #281 using latest Node, so I was really surprised to see the same error. I’m not that familiar with CRA, all I know is that importing the runtime using v2 works. 🧐

I think it’s best to get this fix merged and released, then afterwards I can dive into what exactly happened.

PR #284, opened 2018-10-02
Don't include metaString in code props

morgs32:
React throws a warning:
`Warning: React does not recognize the metaString prop on a DOM element.`

pasted_image_10_2_18__1_22_pm

### DIFF
--- a/packages/mdx/mdx-ast-to-mdx-hast.js
+++ b/packages/mdx/mdx-ast-to-mdx-hast.js
@@ -30,11 +30,11 @@ module.exports = function mdxAstToMdxHast() {
           props.className = ['language-' + lang]
         }
 
-        props.metaString = node.lang && node.lang.replace(langRegex, '').trim()
+        const metaString = node.lang && node.lang.replace(langRegex, '').trim()
 
         const meta =
-          props.metaString &&
-          props.metaString.split(' ').reduce((acc, cur) => {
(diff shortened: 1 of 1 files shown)


silvenon:
The infostring is added on purpose, but maybe @ChristopherBiscardi can shed some light on whether the `metaString` prop itself is necessary, because the following:

~~~
```js repl=true id=123
const a = 5
```
~~~

passes two props to the code block:

```js
{
  metaString: "repl=true",
  repl: true,
  id: 123,
}
```

In my opinion `metaString` isn't useful, but maybe it has its reasons. In any case, if you're using infostrings it's expected that you're somehow making use of those props and it's on you to not pass it to intrinsic elements. However, even not using an infostring still seems to pass `metaString: ""`, so that should probably be removed.

ChristopherBiscardi:
It's there in case someone needs the raw output after the language was stripped off. For example, the current implementation for props doesn't support arrays and you could implement support in userland by parsing the metaString prop differently.

The empty-string metaString being passed through is definitely a bug and can be removed (but that's not what this PR does). Something like this would work:

```js
const metaString = node.lang && node.lang.replace(langRegex, '').trim()
props.metaString = metaString !== "" && metaString
```

morgs32:
@silvenon happy to. Expect updates and test by EOD.

PR #285, opened 2018-10-05
Add documentation and tests for nested MDX

### LINKED ISSUE #241: Transclusion of _inline_ markdown?
Hi. [Markdown file transclusion](https://mdxjs.com/syntax#markdown-file-transclusion) (MFT) is super cool, but I cannot find a way to do the equivalent when all of the markdown is defined in the same file. Something like:

```jsx
// page.mdx
import FancyLayout from '../components/FancyLayout

# Title

[1] Parsed _markdown_ content

<FancyLayout>
  [2] More parsed _markdown_ content, in a fancy layout
</FancyLayout>
```

I would like `[2]` to be parsed as markdown, but not sure how, beyond defining it in a separate file and using MFT. I'd like to keep it all in one file, though, for the sake of non-technical authors. Do I need to somehow make FancyLayout an MDX-savvy component?

If it's relevant, I'm using [gatsby-mdx](https://github.com/ChristopherBiscardi/gatsby-mdx) on a brand-new [gatsby-starter-default](https://github.com/gatsbyjs/gatsby-starter-default).

Thanks!

silvenon:
Closes #241.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -59,6 +59,38 @@ import Contributing from './docs/contributing.md'
 <Contributing />
 ```
 
+#### Nesting MDX
+
+In certain cases you might want to nest MDX directly in a JSX component, without having to create a separate file:
+
+```jsx
+import Card from './components/Card'
+
+# Hello, world!
+
+<Card>
+
+## This is also MDX!
+
+This will be _nested_ inside the `<Card>` component.
+
+</Card>
+```
+
+The key is to separate the nested MDX with an extra newline, otherwise it will be rendered as plain text:
+
+```jsx
+import Card from './components/Card'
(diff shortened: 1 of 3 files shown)


johno:
I'm not entirely sure that we want to document this since it isn't intended behavior and isn't currently part of the spec (it really just "accidentally" works due to the jsx parsing). We've been going back and forth a lot as to whether we want to truly support this type of syntax or not.

For the time being I still lean towards leaving it out until since interleaving JSX and MD results in some ambiguity we'd have to sort out and define.

silvenon:
I agree, it's an accident that people shouldn't rely on. Ok, I'll leave #241 open then.

PR #289, opened 2018-10-07
Update Lerna and implement Yarn workspaces

### LINKED ISSUE #268: Change jest timeout for node 8 tests
The node v8 tests intermittently fail for the loader due to a timeout. We should increase the test time limit to see if that resolves the issue.

silvenon:
I updated the docs and Travis to prefer Yarn because we ship `yarn.lock` and we have workspaces. But everything works with npm too.

I also increased the timeout for the loader test. I'll proceed by adding inline comments.

Closes #268.


.travis.yml
```diff
@@ -5,5 +5,6 @@ node_js:
 git:
   depth: 5
 cache:
+  yarn: true
```
silvenon: I'm not sure if this is necessary, i.e. if it happens by default.

lerna.json
```diff
@@ -1,8 +1,8 @@
 {
-  "lerna": "2.5.1",
```
silvenon: Lerna v3 doesn't have this field.

package.json
```diff
@@ -1,19 +1,21 @@
...
   "private": true,
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "lerna bootstrap && lerna link --force-local",
     "docs": "x0 docs",
     "docs:build": "x0 build docs",
     "docs:deploy": "npm run docs:build && now dist --public && now alias",
     "test": "lerna run test",
-    "format": "prettier --no-semi --single-quote --write '{examples,packages}/**/*.js'",
-    "prepare": "lerna bootstrap && lerna link --force-local"
```
silvenon: I removed the `prepare` script because `lerna bootstrap` in Lerna v3 triggers that script, so it causes an endless loop. It would be ideal if we could keep this, because then `lerna bootstrap` would happen automatically and contributors and Travis CI wouldn't have to know about it, so I'm open to ideas.

packages/mdxast/package.json
```diff
@@ -22,7 +22,7 @@
     "unist-util-visit": "^1.3.0"
   },
   "devDependencies": {
-    "jest": "^22.4.3",
+    "jest": "^23.6.0",
```
silvenon: Tests were failing for mdxast, updating Jest fixed it.

silvenon:
I'm having a hard time figuring out why tests are failing. I'll investigate tomorrow.

.travis.yml
```diff
@@ -4,6 +4,11 @@ node_js:
   - 10
 git:
   depth: 5
+before_script:
+  # https://github.com/lerna/lerna/issues/1457#issuecomment-399678173
+  - yarn lerna exec yarn install
+  - yarn lerna link --force-local
```
silvenon: Maybe it's a Lerna + Linux issue, but `lerna bootstrap` seems to work only partially on Travis, while `lerna exec yarn install` works entirely. https://github.com/lerna/lerna/issues/1457#issuecomment-399678173

I'm adding this workaround for now, but I'll also report this issue to Lerna to see if they can figure it out.

packages/loader/test/index.test.js
```diff
@@ -51,4 +51,4 @@ const testFixture = (fixture, options = {}) => {
 
 test('it loads markdown and returns a component', async () => {
   await testFixture('fixture.md')
-})
+}, 10000)
```
silvenon: Previously Node v8 tests were failing, but now Node v10 were failing too, or instead. This seems to fix it.

silvenon:
This has been bumpy, but it's ready for reviewing!

PR #297, opened 2018-11-03
[mdx] Add detab@^2.0.0 in package.json

### LINKED ISSUE #296: Yarn pnp
Versions:
- `yarn@1.12.1`
- `@mdx-js/mdx@0.15.6`

When I try to import the module 
```
Error: Package "@mdx-js/mdx@0.15.6" (via ".../Yarn/v3/npm-@mdx-js-mdx-0.15.6-b6644c3d674b9e94a1ad933ef1a76944917e2963/node_modules/@mdx-js/mdx/mdx-ast-to-mdx-hast.js")
is trying to require the package "detab" (via "detab") without it being listed
in its dependencies (change-case, mdast-util-to-hast, remark-parse,
remark-squeeze-paragraphs, to-style, unified, unist-util-visit, @mdx-js/mdx)
```

Ayc0:
Proposal to fix #296

### DIFF
--- a/packages/mdx/package.json
+++ b/packages/mdx/package.json
@@ -23,6 +23,7 @@
   ],
   "dependencies": {
     "change-case": "^3.0.2",
+    "detab": "^2.0.0",
     "mdast-util-to-hast": "^3.0.0",
     "remark-parse": "^5.0.0",
     "remark-squeeze-paragraphs": "^3.0.1",


silvenon:
Thanks! How did this even work before? I ran `yarn why detab` in my project:

```
❯ yarn why detab
yarn why v1.12.1
[1/4] 🤔  Why do we have the module "detab"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "detab@2.0.1"
info Reasons this module exists
   - "@mdx-js#mdx#mdast-util-to-hast" depends on it
   - Hoisted from "@mdx-js#mdx#mdast-util-to-hast#detab"
info Disk size without dependencies: "16KB"
info Disk size with unique dependencies: "36KB"
info Disk size with transitive dependencies: "36KB"
info Number of shared dependencies: 1
✨  Done in 1.00s.
```

Does this mean that Yarn automatically installed it for me because `mdast-util-to-hast.js` requires it? I've never seen this before…

Ayc0:
It was working because it's was a dependency of another package so it was injected in the node_modules

I discovered this using yarn --pnp and as it removed node_modules, if something isn't specified as a dep of a devDep, it won't create any symbolic links and thus cannot work at all

Ayc0:
Can you also deploy a new version on npm?

silvenon:
> It was working because it's was a dependency of another package so it was injected in the node_modules

Oh, wow. It's pretty weird that this happens by default, I'll try to watch out for this stuff in our tests.

> Can you also deploy a new version on npm?

Released! Thanks for discovering this. 👍

Ayc0:
I think there is the same issue on `unist-builder`

PR #298, opened 2018-11-03
Add unist-builder in package.json

Ayc0:
I haven't install the project, I'm doing the modification on github directly, you may need to relaunch yarn to update the yarn.lock

Related to #297 

It seems to be the last dependency that isn't mentioned in the package.json.


silvenon:
Once #300 is merged, I'll release a minor version. I checked with ESLint and all there are no other extraneous dependencies.

PR #300, opened 2018-11-03
Implement ESLint

silvenon:
This PR is not super fun, but it's an efficient setup to prevent us from forgetting to add dependencies and to enforce Prettier, because until this point using Prettier was optional.

- integrate Prettier as an ESLint plugin
- run Prettier across the codebase
- set up crucial import rules to avoid issues like #296
- enforce ESLint through a pre-commit hook, enforcing Prettier as well!
- run ESLint on Travis CI

Recommended text editor workflow:
- turn off Prettier plugin
- use an ESLint plugin and turn on autofix on save.


docs/_app.js
```diff
@@ -46,38 +53,35 @@ const pageNames = {
...
+          <SidebarLayout {...this.props} logo={<Logo />} routes={nav} />
         </ScopeProvider>
       </RebassMDX>
     )
   }
 }
+
+App.defaultProps = {
+  title: 'MDX'
+}
```
silvenon: ESLint doesn't support class properties yet and I didn't want to introduce babel-eslint just for this.

package.json
```diff
@@ -15,13 +14,33 @@
...
+    "singleQuote": true
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.js": [
+      "eslint"
```
silvenon: Lints only staged JavaScript files, much faster!

docs/_app.js
silvenon: I didn't touch any JavaScript code except this. Everything else is Prettier.

silvenon:
In this particular instance I wonder how you got different results, because at this early stage there shouldn't be any collisions. This is how the setup works:

- eslint-config-prettier disables all ESLint rules set up from previous entries in `extends` that collide with Prettier
- eslint-plugin-prettier uses Prettier when using `eslint --fix`

I think prettier-eslint is a different approach, it applies first Prettier, then ESLint. In that case ESLint might do something that Prettier alone would do differently. In that workflow I don't see a way to **enforce** both tools because they are two rules which can override each other, what Prettier does might be an ESLint error and vice-versa. Implementing in using eslint-config-prettier and eslint-plugin-prettier should ensure that these tools are treated as one.

I wonder how you got two different outputs, this setup is quite simple and both commands should yield the same result. Do you mind explaining more about why you believe using eslint-prettier is better and how your results differ?

Ayc0:
I think it's because I wasn't using `prettier/prettier: error` in my eslint config

Ayc0:
In general, I prefer to set everything in eslint, because when you use configs like [airbnb](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb) or [standard](https://github.com/standard/eslint-config-standard), they come with their own formatting rules.

And in general when I was using prettier in eslint with those, I had some issues, either prettier was overriding those configs or it wasn't taken into account so I used prettier-eslint

silvenon:
I tried implementing remark-cli and remark-preset-wooorm as well, but remark-cli's autofixing abilities seem pretty limited at the moment. Manually fixing 200+ warnings seems like too much work, I can imagine enforcing this through the pre-commit hook and Travis being pretty tedious for everyone. @wooorm is there a better way?

silvenon:
Btw, I decided to implement xo as a shareable config rather than using xo itself. I don't understand why xo is a separate tool, it seems like an unnecessary abstraction of ESLint (sorry Sindre).

wooorm:
@silvenon Awesome work!

> manually fixing 200+ warnings seems like too much work.

I can do that in a separate PR if y’all want to go that way!

> Enforcing these rules through the pre-commit hook and Travis might turn out to be tedious for everyone.

Why? Not to go into a whole discussion, but I’d argue that coding style in markdown is as important as coding style in javascript.

---

Aside a) it may make sense to create a remark preset that doesn’t touch upon style, so that it could be used in prettier for markdown / mdx

Aside b) I personally run formatting for code and markdown in the `npm test` script as well, that way, `npm test` just makes sure the whole project is OK, rather than just running traditional tests.

silvenon:
> Why? Not to go into a whole discussion, but I’d argue that coding style in markdown is as important as coding style in javascript.

I very much agree, I love linting. My reason was the tooling, remark's autofix is limited and there doesn't seem to be a text editor extension which autofixes on file save, like ESLint. When this is the case, a person making lots of contribution to the docs might be welcomed with a large amount of warnings at the end, instead of all of that happening while writing.

But I'm just unfamiliar with this workflow, maybe I overlooked something.

> Aside a) it may make sense to create a remark preset that doesn’t touch upon style, so that it could be used in prettier for markdown / mdx

I like that, but this is all pretty new to me, so I would appreciate some guidance. 😃 

> Aside b) I personally run formatting for code and markdown in the `npm test` script as well, that way, `npm test` just makes sure the whole project is OK, rather than just running traditional tests.

Do you run the code that only lints these files or actually fixes them? I would prefer only linting, because it would be unexpected if the test script actually writes something. I'll add that.

silvenon:
In any case, I'd like to solve Markdown in a separate PR because it's important this large changeset gets merged soon.

PR #301, opened 2018-11-03
Update Node in Travis

wooorm:
* Use latest LTS Carbon (8) version
* Use latests node (currently 11)


silvenon:
Good idea, thanks!

Issue #307, opened 2018-11-09
In-page hash navigation does not set focus


AlmeroSteyn:
I have initially logged this as an issue [on the gatsby-mdx repo here](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/185);

**Describe the bug**
When I click an anchor that refers to a location inside the page, the destination element does not receive proper focus. Although the browser scroll position is reset properly, the destination element is not read out by screen readers as it does not receive focus correctly.

**To Reproduce**
1. Clone repository https://github.com/AlmeroSteyn/gatsby-mdx-minimal-repro
1. Go to homepage
1. Click on the link right at the top named `Skip to content`
1. Note how the container below with the text `This is some content` does not get a focus outline.
1. Now go to route `/focus` which is a straight React component without `MDX`.
1. Click on link right at the top named `Skip to content`
1. Note how the container below with text `This is some content` **DOES** get a focus outline on this page.

Both of these pages uses te same `Layout` component that contains the link and in-page navigation destination. When used in an `MDX` file it doesn't work, while using it as a straight page component does work

When I was looking into why this was happening I noticed that on the `MDX` pages the entire content is remounted even when the hash portion of the URL changes. This breaks the standard browser behaviour of setting focus to the target element.

Because focus is not set correctly, screen readers will not announce to the users that any action has been performed.

During triage of the issue on the `gatsby-mdx` repo it was found that setting a `shouldComponentUpdate` flag based on the `location.hash` value on the `MDXTag` component will fix the issue in some cases. However the following cases should work and properly set focus to the element in question:

1. Clicking a hash anchor when the browser URL is not set to the hash location.
1. Clicking a hash anchor when the browser URL was set before by a previous click.
1. Refreshing a page when the browser URL already contains the hash.

**Expected behavior**
Navigating to an element with an `id` corresponding to the `hash in the href` of the anchor should set focus on the element to allow screen reader users to also use applications built with `MDX`.

johno:
Thank you for the thorough bug report @AlmeroSteyn. I'll have a deeper look into it this weekend and report back next week.

##### Initial findings

At an initial glance I think it has to do with the `layoutProps` being passed to MDXRenderer/MDXTag within MDX and `gatsby-mdx` internals. Currently they're all forwarded along which will cause a rerender when the window location changes. When the rerender happens the DOM node to match the hash is then removed causing the browser to lose focus.

[...]

PR #309, opened 2018-11-10
Stabilize UserLayout

ChristopherBiscardi:
Fixes https://github.com/mdx-js/mdx/issues/307

Pulls the layout for an MDX file into a reference that will be stable
across renders. Previously, a layout defined in MDX content as an
inline arrow function:

```js
export default ({ children, ...props }) => <div {...props}>{children}</div>
```

would result in function defined inline in render, causing the
reference to be created each time resulting in a remounting.

This PR stablizes the reference by pulling the layout outside of
render, resulting in no remount.

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -95,18 +95,29 @@ function toJSX(node, parentNode = {}, options = {}) {
 
       jsxNodes.push(childNode)
     }
-
     return (
       importNodes.map(childNode => toJSX(childNode, node)).join('\n') +
       '\n' +
       exportNodes.map(childNode => toJSX(childNode, node)).join('\n') +
       '\n' +
-      (skipExport ? '' : 'export default ({components, ...props}) => ') +
-      `<MDXTag name="wrapper" ${
-        layout ? `Layout={${layout}} layoutProps={props}` : ''
-      } components={components}>${jsxNodes
-        .map(childNode => toJSX(childNode, node))
-        .join('')}</MDXTag>`
+      (skipExport
+        ? ''
+        : `export default class MDXContent extends React.Component {
+  constructor() {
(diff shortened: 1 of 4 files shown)


ChristopherBiscardi:
The additional declaration isn't making it through the compilation process. Since I tested this approach on already-compiled code it didn't have that issue when I tested originally. 

I've also since discovered that a couple of things rely on the raw MDX output, like the runtime, so this would require a bit more change than anticipated.

johno:
This is looking great, thanks!

We're definitely reaching a point where we might want to think about how to handle the MDX "templating" since it's starting to get pretty logic-filled. I wonder if we should just duplicate some of the output logic but have it handled differently based on whether `skipExport` is true or not?

ChristopherBiscardi:
I definitely agree that this templating is getting a bit hard to work with. I messed up a quote, etc here or there a couple times. I'll finish up the full conversion and then see how it looks, then change it if it still needs changing

ChristopherBiscardi:
@silvenon I added a snapshot test with the other tests in `mdx/test/index.test.js`. This seemed like the right place since both `mdx` and `runtime` packages have explicit tests against the JSX output. I don't know how to get rid of the wrapped quotes in the inline output though. Any ideas?

silvenon:
> Can also use a smaller content. I re-used a fixture that already existed.

Yeah, maybe it's better if you just use a single `# Hello World`. It should be enough to monitor this simple output.

> By wrapped quotes, I means this effect. Where there's a template string, then the string of code starts with double quotes leading to escaped doubled quotes all over the code.

The string itself may not be pretty to look at, but try to mess up the output and see the diff when you run the test. If the diff is nice, that's all that matters. 😉

ChristopherBiscardi:
rebased. Only significant change was the additional prettier dep causing a conflict in mdx/package.json. Will squash/merge after tests pass.

PR #315, opened 2018-11-17
Improve docs

wooorm:
This PR improves the docs in several ways.

#### ~~TODO~~DONE:

* [x] Add more badges
* [x] Proper readme’s per project
* [x] Proper main readme with sponsors etc

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
       "git add"
     ],
     "*.md": [
-      "remark -qfo",
+      "remark -qf",
       "git add"
     ]
   },
--- a/package.json
+++ b/package.json
@@ -1,19 +1,13 @@
 {
   "private": true,
-  "scripts": {
-    "bootstrap": "lerna bootstrap && lerna link --force-local",
-    "docs": "x0 docs",
-    "docs:build": "x0 build docs",
-    "docs:deploy": "npm run docs:build && now dist --public && now alias",
-    "test": "remark . -qf && eslint . && lerna run test",
-    "format": "remark . -qfo && eslint . --fix",
-    "publish": "lerna publish"
-  },
-  "repository": "mdx-js/mdx",
   "license": "MIT",
-  "workspaces": [
-    "packages/*"
-  ],
+  "homepage": "https://mdxjs.com",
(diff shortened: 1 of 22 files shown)


silvenon:
You're on 🔥🔥🔥

wooorm:
Alright, I think this is good! They’re mostly straightforward changes.
But most noteworthy are a CoC and I believe the readme changes need to be carefully reviewed: https://github.com/mdx-js/mdx/blob/6537433d6a5a6ef2fa88e120924008df00d2e5a0/readme.md.

packages/create-mdx/package.json
```diff
@@ -1,14 +1,33 @@
...
+    "remark",
+    "unified"
+  ],
+  "homepage": "https://mdxjs.com",
+  "repository": "mdx-js/mdx",
+  "bugs": "https://github.com/mdx-js/mdx/issues",
+  "author": "John Otander <johnotander@gmail.com> (http://johnotander.com)",
+  "contributors": [
+    "John Otander <johnotander@gmail.com> (http://johnotander.com)",
+    "Tim Neutkens <timneutkens@me.com>",
```
timneutkens: This email address is not public / should not be public 😄

```suggestion
    "Tim Neutkens <tim@zeit.co>",
```

wooorm: Hmm, I found it somewhere, that’s why I used it! But fixed!

timneutkens:
Awesome!

package.json
```diff
@@ -69,7 +71,7 @@
       "git add"
     ],
     "*.md": [
-      "remark -qfo",
+      "remark -qf",
```
silvenon: Just curious, why did you remove the `-o`? If I understand correctly, that's similar to ESLint's `--fix`, it fixes what it can so in cases where all errors of a file are auto-fixable, it proceeds to `git add` and you don't have to do anything.

wooorm: Apparently it didn’t work with the `-o`. I’m not sure exactly what lint-staged is doing, but I believe it’s outputting it somehow anyway.

silvenon: 👍

PR #317, opened 2018-11-18
Update `remark`, other dependencies

wooorm:
I thought I’d update remark because it includes the new `meta` prop, which simplifies the code here: https://github.com/mdx-js/mdx/blob/6c446d69beaf04eabe708b90a6ac9634fd1c02af/packages/mdx/mdx-ast-to-mdx-hast.js

I also updated the other dependencies/dev-dependencies.


packages/mdx/mdx-ast-to-mdx-hast.js
```diff
@@ -23,18 +23,18 @@ function mdxAstToMdxHast() {
       code(h, node) {
         const langRegex = /^[^ \t]+(?=[ \t]|$)/
```
johno: Think we can ✂️ this as well now?

package.json
```diff
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@rebass/mdx": "^1.0.0-1",
-    "styled-components": "^3.3.3"
+    "styled-components": "^4.0.0"
```
johno: I'll need to make an upstream change to `x0` before bumping `styled-components` version so we might want to skip this update as part of this PR. It's on my todo list today to get this fixed.

wooorm: OK, perfect, ping back here when you’re done!

wooorm: Or we can delay, whatever works?

johno: If you want to keep this pinned at 3 you can merge this and I can PR in this update separately (I don't want to be a blocker, currently on the road with spotty internet access).

wooorm: Downgraded! I believe this is ready now, and could be batched up with the docs changes as a new release!

PR #320, opened 2018-11-20
super before this access

### LINKED ISSUE #319: v16: super before this access
I think I forgot a `super()` in the MDXContent class from #309. Not sure why it didn't pop up before when running tests, etc.

image

ChristopherBiscardi:
closes #319

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -103,7 +103,8 @@ function toJSX(node, parentNode = {}, options = {}) {
       `${
         skipExport ? '' : 'export default'
       } class MDXContent extends React.Component {
-  constructor() {
+  constructor(props) {
+    super(props)
     this.layout = ${layout}
   }
   render() {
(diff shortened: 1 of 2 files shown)


johno:
Thanks @ChristopherBiscardi, can you update the snapshots as well?

ChristopherBiscardi:
ah right, of course. Updated.

silvenon:
I wish we could add a test proving that the warning is no longer there, i.e. a one which would show the warning before applying the patch. But that doesn't sound trivial.

Thanks for spotting and patching this so quickly!

silvenon:
Released!

ChristopherBiscardi:
I also wish we could test for that, but I'm more confused as to why it didn't show up before that in our tests. It wasn't just a warning, it was an Uncaught ReferenceError that blocked rendering.

johno:
Yeah I'm actually quite surprised this got through the test suite. I'll do a bit of digging today to see how we can get a more thorough integration test wired up to make sure we get an ❌if core outputs something that can't render.

johno:
I would've expected at least `@mdx-js/runtime` to error, though in the core tests for `@mdx-js/mdx` we're mostly testing parseability and never actually rendering (which we should change).

johno:
#321 should do it: https://travis-ci.org/mdx-js/mdx/jobs/457690098#L606

PR #323, opened 2018-11-21
Define components in render

### LINKED ISSUE #322: 0.16.2 mdx-loader throw error
### Subject of the issue

image

image

### Your environment

*   Name and version of operating system
*   Names and version of required packages
*   Version of node, npm, yarn, or names and versions of browser

### Steps to reproduce

Tell us how to reproduce this issue.  Please provide a working and simplified example.

🎉 BONUS POINTS for creating a [minimal reproduction](https://stackoverflow.com/help/mcve) and uploading it to GitHub.  This will get you the fastest support.  🎉

### Expected behaviour

What should happen?

### Actual behaviour

What happens instead?

johno:
Fixes #322

### DIFF
--- a/packages/runtime/src/index.js
+++ b/packages/runtime/src/index.js
@@ -29,7 +29,7 @@ export default ({
   const {code} = transform(jsx)
 
   const keys = Object.keys(fullScope)
-  const values = keys.map(key => fullScope[key])
+  const values = Object.values(fullScope)
   // eslint-disable-next-line no-new-func
   const fn = new Function(
     '_fn',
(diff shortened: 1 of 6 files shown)


packages/runtime/src/index.js
```diff
@@ -29,15 +29,15 @@ export default ({
...
+  const values = Object.values(fullScope)
   // eslint-disable-next-line no-new-func
   const fn = new Function(
     '_fn',
     'React',
     ...keys,
     `${code}
 
-  return React.createElement(MDXContent);`
+  return React.createElement(MDXContent, { components });`
```
johno: We need to manually pass components as a prop, otherwise `const { components = {} } = this.props` in render will override them.

PR #326, opened 2018-11-22
Assign props at the top level of render

johno:


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -105,10 +105,10 @@ function toJSX(node, parentNode = {}, options = {}) {
       } class MDXContent extends React.Component {
   constructor(props) {
     super(props)
-    this.layout = ${layout}
+    this.layout = ${layout || 'null'}
```
ChristopherBiscardi: could also remove the entire constructor instead if layout isn't defined.

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -105,10 +105,10 @@ function toJSX(node, parentNode = {}, options = {}) {
...
   }
   render() {
-    const { components = {} } = this.props
+    const { components = {}, ...props } = this.props
```
ChristopherBiscardi: won't this need an equivalent change in the runtime to pass the props into createElement?

Previously, the runtime was just using the scoped names when passed in to defined the `components` and `props` variables. If we're standardizing the class to be `this.props` access always, then we need to pass everything into createElement in the runtime like

```js
createElement(MDXContent, {components, ...props}, children)
```

ChristopherBiscardi:
I can't push it up to this branch, but this additional change passes the tests for me.

```diff
diff --git a/packages/runtime/src/index.js b/packages/runtime/src/index.js
index 12f2bdd..0cd13de 100644
--- a/packages/runtime/src/index.js
+++ b/packages/runtime/src/index.js
@@ -37,7 +37,7 @@ export default ({
     ...keys,
     `${code}
 
-  return React.createElement(MDXContent, { components });`
+  return React.createElement(MDXContent, { components, ...props });`
   )
 
   return fn({}, React, ...values)
```

packages/mdx/mdx-hast-to-jsx.js
silvenon: I think this is semantically cleaner than omitting the constructor entirely, because it's communicating that there is no user layout. `null` is semantically different than `undefined`.

johno:
Thanks @ChristopherBiscardi, that diff was the ticket for runtime 👍

Issue #312, opened 2018-11-16
Comment bug (spec)


wooorm:
Comments coming from markdown, going to MDX**H**AST, will be represented like this: `{type: 'comment', value: '<!--something-->'}`, whereas comments in hast itself [don’t have the prefix/suffix](https://github.com/syntax-tree/hast#comment).

This is not a **bug**, in that it doesn’t matter in this project, but it’s a bug compared to mdast.

https://github.com/mdx-js/mdx/blob/be4877ec897ff92166d700ef8a7ebbb87583e857/packages/mdx/md-ast-to-mdx-ast.js#L5-L6

The code above, in `md-ast-to-mdx-ast.js`, should drop the prefix and suffix

https://github.com/mdx-js/mdx/blob/be4877ec897ff92166d700ef8a7ebbb87583e857/packages/mdx/mdx-hast-to-jsx.js#L152-L153

The code above, in `mdx-hast-to-jsx.js`, should not replace, but prepend and append the JSX prefix and suffix

PR #328, opened 2018-11-22
Fix comment prefix and suffix

wooorm:
Previously, `<!--` and `-->` were part of comments nodes.
this deferred from the HAST specification.

This PR removed the prefix when comments are found in markdown, and
thus doesn’t need to replace them when rendering to JSX, but rather
prefixes and suffixes them.

Closes GH-312.

### DIFF
--- a/packages/mdx/md-ast-to-mdx-ast.js
+++ b/packages/mdx/md-ast-to-mdx-ast.js
@@ -1,9 +1,16 @@
 const visit = require('unist-util-visit')
 
+var commentOpen = '<!--'
+var commentClose = '-->'
+
 module.exports = options => tree => {
   visit(tree, 'html', node => {
-    if (node.value.startsWith('<!--') && node.value.endsWith('-->')) {
+    if (
+      node.value.startsWith(commentOpen) &&
+      node.value.endsWith(commentClose)
+    ) {
       node.type = 'comment'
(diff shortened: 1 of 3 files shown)


packages/mdx/md-ast-to-mdx-ast.js
```diff
@@ -1,9 +1,16 @@
 const visit = require('unist-util-visit')
 
+var commentOpen = '<!--'
```
timneutkens:
```suggestion
const commentOpen = '<!--'
```

packages/mdx/md-ast-to-mdx-ast.js
```diff
@@ -1,9 +1,16 @@
...
+var commentClose = '-->'
```
timneutkens:
```suggestion
const commentClose = '-->'
```

silvenon:
Seeing how I missed the `var` usage, you could also add `no-var: error` to ESLint configuration. 😅

wooorm:
Good catch! Rather not have it in this PR, but posted it in https://github.com/mdx-js/mdx/pull/338#issuecomment-441462399.

wooorm:
@silvenon Fixed!

silvenon:
Thanks!

Issue #316, opened 2018-11-18
Using `create-mdx` with something like `x0` will create a project called `next-mdx`


wooorm:
### Subject of the issue

I believe `name` should, when `example` is given, be based off of that:

https://github.com/mdx-js/mdx/blob/352264ff376d3ee96262db636c5d4ea087bd5221/packages/create-mdx/cli.js#L4

### Steps to reproduce

`npm init mdx x0`

### Expected behaviour

`ls` should include `x0-mdx` or something.

### Actual behaviour

`ls` includes `next-mdx`.

wooorm:
In fact, `.slice(2, 3)` only returns the value at `2`. So `name` is never set.

PR #329, opened 2018-11-22
Fix `create-mdx` name by basing it on example

wooorm:
Closes GH-316.


johno:
We could execute the CLI to ensure it does what we expect, but since this is so tiny I'm not sure its really required at this point.

Issue #310, opened 2018-11-10
unused code and redefined variables, and set initial value


Himenon:
### Subject of the issue

I would like to refactor the code below. Because I used TypeScript, and concerned about the error.

* Redefined variables
	* https://github.com/mdx-js/mdx/blob/master/packages/mdx/index.js#L130 <-- want to remove
	* https://github.com/mdx-js/mdx/blob/master/packages/mdx/index.js#L134
* unused code
	* https://github.com/mdx-js/mdx/blob/master/packages/mdx/test/index.test.js#L3
	* https://github.com/mdx-js/mdx/blob/master/packages/tag/test/mdx-tag.test.js#L1
	* https://github.com/mdx-js/mdx/blob/master/packages/tag/test/mdx-provider.test.js#L1
	* https://github.com/mdx-js/mdx/blob/master/packages/runtime/test/index.test.js#L3
* Specifying initial values for optional arguments
	* https://github.com/mdx-js/mdx/blob/master/packages/tag/src/mdx-provider.js#L5

### Expected behaviour

* Redefined variables

```diff
# https://github.com/mdx-js/mdx/blob/master/packages/mdx/index.js
- compile.sync = sync

module.exports = compile
exports = compile
exports.sync = sync
```
* unused code
	* remove
* Specifying initial values for optional arguments

```diff
# https://github.com/mdx-js/mdx/blob/master/packages/tag/src/mdx-provider.js#L5
- export const withMDXComponents = Component => ({ components, ...props }) => (
+ export const withMDXComponents = Component => ({ components = undefined, ...props }) => (
```

Himenon:
Thank you for your reply.

> compile.sync is part of the API, is there a comment or something that we could add to ignore the TypeScript error?

I think that `compile.sync` and `exports.sync = sync` have the same meaning. References on the `import` or `require` side will see the one defined after `compile.sync` and `exports.sync = sync`, so the result will be the same.

It is simply an error that the previously defined one is overwritten, so if either `compile.sync` or `exports.sync = sync` remains, it will fix it.

> Also, making the argument default to undefined is odd because it already is undefined if it's using the default value. Is the another workaround?

Sorry, my environment is too strict. Even with the current situation, the result is the same, so there is no problem even if it is as it is.

ChristopherBiscardi:
@Himenon What is your environment? How can we reproduce seeing these issues?

silvenon:
Oh, I see, setting `compile.sync` does seem redundant.

[...]

PR #338, opened 2018-11-25
enhanced eslint

Himenon:
This PR is related to #310.

As I mentioned in the issue, I would like to detect unused variables and improve code quality.

Therefore, we propose to use `eslint:recommended` and `plugin:react/recommended` in this PR.
Doing this will also help to lower the cost of reviews.

However, this required a moderate mitigation or correction in order to provide stringent constraints. These changes are made in this PR.

Also, if undesirable changes are made, it will be helpful if you can point them out.

thanks

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -15,6 +15,8 @@ plugins:
   - prettier
 
 extends:
+  - eslint:recommended
+  - plugin:react/recommended
   - plugin:import/errors
   - xo
   - prettier
@@ -23,9 +25,21 @@ rules:
   eqeqeq: off
   no-eq-null: off
   no-negated-condition: off
-  no-unused-vars: off
+  no-unused-vars:
+  - error
+  - vars: all
+    args: after-used
+    argsIgnorePattern: "^_"
+    varsIgnorePattern: "^React$"
   import/no-extraneous-dependencies: error
   prettier/prettier: error
+  no-console: 0
+  react/prop-types: 0
+  react/display-name: 0
+
+settings:
+  react: 
+    version: 16.3.2
 
 overrides:
   -
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -31,7 +31,6 @@ rules:
       argsIgnorePattern: "^_"
   no-var: error
   import/no-extraneous-dependencies: error
-  capitalized-comments: off
   prettier/prettier: error
   no-console: off
   react/prop-types: off
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -36,7 +36,6 @@ rules:
   no-console: off
   react/prop-types: off
   react/display-name: off
-  react/react-in-jsx-scope: off
   react/no-children-prop: off
 
 settings:
@@ -62,3 +61,8 @@ overrides:
     rules:
       import/no-unresolved: off
       import/no-extraneous-dependencies: off
+  -
+    files:
+      - "examples/next/**/*.js"
+    rules:
+      react/react-in-jsx-scope: off
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -26,16 +26,18 @@ rules:
   no-eq-null: off
   no-negated-condition: off
   no-unused-vars:
-  - error
-  - vars: all
-    args: after-used
-    argsIgnorePattern: "^_"
-    varsIgnorePattern: "^React$"
+    - error
+    - args: after-used
+      argsIgnorePattern: "^_"
+  no-var: error
   import/no-extraneous-dependencies: error
+  capitalized-comments: off
   prettier/prettier: error
-  no-console: 0
-  react/prop-types: 0
-  react/display-name: 0
+  no-console: off
+  react/prop-types: off
+  react/display-name: off
+  react/react-in-jsx-scope: off
+  react/no-children-prop: off
 
 settings:
   react: 
--- a/docs/_ui.js
+++ b/docs/_ui.js
@@ -15,10 +15,6 @@ export const LiveEditor = props => {
     case '!':
       return <LivePreview mdx={lang === '!mdx'} code={code} />
     default:
-      return (
-        <Pre p={3} mt={4} mb={4} bg="gray">
-          {props.children}
-        </Pre>
-      )
+      return <Pre p={3} mt={4} mb={4} bg="gray" children={props.children} />
   }
 }
--- a/docs/_ui.js
+++ b/docs/_ui.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import {Pre, Box, Border} from 'rebass'
+import {Pre} from 'rebass'
 import {LiveEditor as Editor, LivePreview} from '@compositor/x0/components'
 
 export const Logo = () => <img src="https://mdx-logo.now.sh" width="70" />
@@ -15,6 +15,10 @@ export const LiveEditor = props => {
     case '!':
       return <LivePreview mdx={lang === '!mdx'} code={code} />
     default:
-      return <Pre p={3} mt={4} mb={4} bg="gray" children={props.children} />
+      return (
+        <Pre p={3} mt={4} mb={4} bg="gray">
+          {props.children}
+        </Pre>
+      )
   }
 }
--- a/examples/create-react-app/config-overrides.js
+++ b/examples/create-react-app/config-overrides.js
@@ -1,5 +1,5 @@
 const {getBabelLoader} = require('react-app-rewired')
-module.exports = (config, env) => {
+module.exports = (config, _env) => {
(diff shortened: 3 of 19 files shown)


.eslintrc.yml
```diff
@@ -23,9 +25,21 @@ rules:
...
+    varsIgnorePattern: "^React$"
   import/no-extraneous-dependencies: error
   prettier/prettier: error
+  no-console: 0
+  react/prop-types: 0
+  react/display-name: 0
+
+settings:
+  react: 
+    version: 16.3.2
```
Himenon: https://github.com/yannickcr/eslint-plugin-react#configuration.

docs/_ui.js
```diff
@@ -15,6 +15,10 @@ export const LiveEditor = props => {
     case '!':
       return <LivePreview mdx={lang === '!mdx'} code={code} />
     default:
-      return <Pre p={3} mt={4} mb={4} bg="gray" children={props.children} />
+      return (
+        <Pre p={3} mt={4} mb={4} bg="gray">
+          {props.children}
```
Himenon: If you do not want it, you can avoid `react/no-children-prop: 0` by adding it to eslint.
* [react/no-children-prop](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-children-prop.md)

examples/create-react-app/config-overrides.js
```diff
@@ -1,5 +1,5 @@
 const {getBabelLoader} = require('react-app-rewired')
-module.exports = (config, env) => {
+module.exports = (config, _env) => {
```
Himenon: Unused arguments will be underscored to avoid eslint errors.
https://eslint.org/docs/rules/no-unused-vars#argsignorepattern

examples/next/pages/_document.js
```diff
@@ -1,3 +1,4 @@
+import React from 'react'
```
Himenon: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md

packages/remark-mdx/index.js
```diff
@@ -1,10 +1,8 @@
-const unified = require('unified')
-const toMDAST = require('remark-parse')
-
 const IMPORT_REGEX = /^import/
 const EXPORT_REGEX = /^export/
 const EXPORT_DEFAULT_REGEX = /^export default/
-const BLOCKS_REGEX = '[a-z\\.]+(\\.){0,1}[a-z\\.]'
+// eslint-disable-next-line
+// const BLOCKS_REGEX = '[a-z\\.]+(\\.){0,1}[a-z\\.]'
```
Himenon: It is an unused variable, but commenting out is preferable.

.eslintrc.yml
```diff
@@ -23,9 +25,21 @@ rules:
   eqeqeq: off
   no-eq-null: off
   no-negated-condition: off
-  no-unused-vars: off
+  no-unused-vars:
+  - error
```
Himenon: I wanted to do this

.eslintrc.yml
```diff
@@ -23,9 +25,21 @@ rules:
...
+  - vars: all
+    args: after-used
+    argsIgnorePattern: "^_"
+    varsIgnorePattern: "^React$"
   import/no-extraneous-dependencies: error
   prettier/prettier: error
+  no-console: 0
+  react/prop-types: 0
+  react/display-name: 0
```
Himenon: These are relaxing rules to match existing code styles.

wooorm:
This looks lovely! Btw, could you add `no-var: error` here as well? I totally missed that in https://github.com/mdx-js/mdx/pull/328#issuecomment-441287285.

packages/remark-mdx/index.js
silvenon: I suggest removing the `eslint-disable-next-line` comment, since you're commenting out the unused variable anyway.

.eslintrc.yml
silvenon: Indent rule options:

```yml
no-unused-vars:
  - error
  ...
```

docs/_ui.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import {Pre, Box, Border} from 'rebass'
+import {Pre} from 'rebass'
```
silvenon: Does this cause any errors? Can you still build the docs without problems?

examples/next/pages/_document.js
silvenon: Use `overrides` to turn off `react/react-in-jsx-scope` for the Next.js example because Next.js docs don't import React AFAICT.

docs/_ui.js
Himenon: This parameter is an unused variable.
I think, even if it is used temporarily, the value will be provided below.

https://github.com/mdx-js/mdx/blob/022b2c6e7a0d200d6f5c8f57205e40ca1bdbbacf/docs/_app.js#L11

Himenon: This is a separate case, but if you run `yarn run docs: build` when testing travis-ci you will not have to always ask this question.

.eslintrc.yml
```diff
@@ -23,9 +25,21 @@ rules:
   eqeqeq: off
   no-eq-null: off
   no-negated-condition: off
-  no-unused-vars: off
+  no-unused-vars:
+  - error
+  - vars: all
+    args: after-used
+    argsIgnorePattern: "^_"
+    varsIgnorePattern: "^React$"
```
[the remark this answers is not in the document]
Himenon: I did not write this rule, so I will turn it off.

.eslintrc.yml
```diff
@@ -23,9 +25,23 @@ rules:
...
   no-eq-null: off
   no-negated-condition: off
-  no-unused-vars: off
+  no-unused-vars:
+    - error
+    - args: after-used
+      argsIgnorePattern: "^_"
+  no-var: error
   import/no-extraneous-dependencies: error
+  capitalized-comments: off
```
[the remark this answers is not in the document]
Himenon: I think this.

https://github.com/xojs/eslint-config-xo/blob/57196751160e5b89e4661bee064da6f88a34d2a4/index.js#L238-L247

.eslintrc.yml
```diff
@@ -23,9 +25,23 @@ rules:
...
   prettier/prettier: error
+  no-console: off
+  react/prop-types: off
+  react/display-name: off
+  react/react-in-jsx-scope: off
```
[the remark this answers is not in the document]
Himenon: I overlooked the matter of `react/react-in-jsx-scope`. Thanks for the advice!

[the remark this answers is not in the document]
Himenon: When `react/displayName` is turned on, 11 errors were found.
Since it was necessary to designate the function name because it was a lot of change amount, I turned it off this time.

For exapmle......

```diff
- export default () => (
-  <MDXProvider components={components}>
-    <Doc />
-  </MDXProvider>
- )
+ export default function awesomeName() {
+   return (
+     <MDXProvider components={components}>
+       <Doc />
+     </MDXProvider>
+   )
+ }
```

If `react/prop-types` is turned on, 8 errors were found.
Also in this case too much change amount, and because we need to install prop-types, I did not include it in this PR as it is not from the main theme.

I do not normally use prop-types, so I think that I can not throw the PR immediately.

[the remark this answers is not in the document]
Himenon: It may be better to treat these two as TODO or fix it with another PR.
Like `no-var` this time.

.eslintrc.yml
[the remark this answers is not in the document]
silvenon: We would like to preserve as many xo rules as possible, so please don't disable this rule, but rather make necessary fixes. I think the rule is autofixable, so running `yarn format` should be enough.

.eslintrc.yml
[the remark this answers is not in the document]
silvenon: That makes sense, thanks for explaining 👍

packages/remark-mdx/index.js
```diff
@@ -1,7 +1,7 @@
 const IMPORT_REGEX = /^import/
 const EXPORT_REGEX = /^export/
 const EXPORT_DEFAULT_REGEX = /^export default/
-// const BLOCKS_REGEX = '[a-z\\.]+(\\.){0,1}[a-z\\.]'
+// Const BLOCKS_REGEX = '[a-z\\.]+(\\.){0,1}[a-z\\.]'
```
Himenon: It was automatically rewritten, but is this a good change?

wooorm: I think you can just remove the whole comment. It's in git if we ever need it!

silvenon: Oops, I didn't think about this. Yeah, remove it! 🔥

PR #339, opened 2018-11-28
Fix custom loader example

jescalan:
This makes the code uglier, unfortunately, but the example provided here in the docs is incorrect (I found this out the hard way).

Since all the code here is indented by 4 spaces, it is treated by the markdown parser as a code block, and you end up seeing the exports in the file inside of a formatted code block. Eliminating the extra indentation fixes this.


johno:
Whoops 😬. Thank you!

PR #353, opened 2018-12-11
Add explore ast link

johno:


johno:
Thanks for getting this [shipped](https://github.com/fkling/astexplorer/pull/344) @silvenon!

wooorm:
@johno Could you hold off on this for a sec? I can bring a fix in for remark-jsx to support inline jsx, which is missing apparently.

wooorm:
I applied the structural fix in remark-mdx. I could move it into another branch as well, if that’s preferred.

Sorry about that, somehow missed mapping inline html to jsx.

PR #354, opened 2018-12-11
Add documentation on running tests

jonsherrard:
Added a note to the contributing docs on running the test for fellow lerna newbies like me.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -65,7 +65,7 @@ prevent you from doing unnecessary work.
 
 ## Running the tests
 
-1.  `npm/yarn install`
+1.  `yarn`
 2.  `yarn lerna exec yarn install`
 3.  `yarn bootstrap`
 4.  `yarn test`
--- a/contributing.md
+++ b/contributing.md
@@ -1,9 +1,9 @@
 # Contributing
 
-Hi! 👋
+Hi!  👋
 We’re excited that you’re interested in contributing!
 Take a moment to read the following guidelines.
-And thanks for contributing to **MDX**! 👏👌✨
+And thanks for contributing to **MDX**!  👏👌✨
 
 If you’re raising an issue, please understand that people involved with this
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -62,6 +63,13 @@ Code contributions are very welcome.  It’s often good to first create an issue
 to report a bug or suggest a new feature before creating a pull request to
 prevent you from doing unnecessary work.
 
+## Running the tests
+
+1.  `npm/yarn install`
+2.  `yarn lerna exec yarn install`
```
timneutkens: `lerna bootstrap` is supposed to take care of this 🤔

jonsherrard: This issue is referenced in the travis.yml where `lerna bootstrap` has been . replaced with script 2:  https://github.com/lerna/lerna/issues/1457

Maybe I could change the instructions to say "lerna boostrap _should_ work", but if it doesn't, try: 'yarn lerna exec yarn install'" ?

timneutkens: I wonder if the alternative is yarn workspaces, this is what we do for Next.js, and just recommend everyone uses `yarn`

contributing.md
```diff
@@ -62,6 +63,13 @@ Code contributions are very welcome.  It’s often good to first create an issue
 to report a bug or suggest a new feature before creating a pull request to
 prevent you from doing unnecessary work.
 
+## Running the tests
+
+1.  `npm/yarn install`
```
johno: We should probably prefer one dep manager and update contrib docs to reflect that. Perhaps here we should prefer `yarn`?

```suggestion
1.  `yarn`
```

contributing.md
johno: I wonder if this should be moved to a troubleshooting section at the bottom of this doc?

johno:
Thanks!

Issue #341, opened 2018-11-29
The word "default" anywhere in the meta description causes an incorrect error


jescalan:
[This is the line](https://github.com/mdx-js/mdx/blob/master/packages/mdx/mdx-hast-to-jsx.js#L52) that's causing the issue. To reproduce, just include the word "default" anywhere in the metadata. For example:

```
export const meta = {
  title: "Testing 123",
  description: "This is a default description"
}

# Content here
```

Again very happy to submit a patch, just want to get this on the maintainer(s) radar

johno:
Thanks for the bug report. We'd gladly accept a patch!

jescalan:
@johno Fantastic! I'll have a patch in shortly. I'm just going to modify the regexes to be a little more precise in picking out the correct patterns.

Another option for long term stability, would be to run an actual parse using something like babel to pull out the tokens, rather than regexes. This is the second regex parsing bug I have turned up in my initial experiments, and I'm sure there are more lurking. I suppose the downside here would be a performance hit. If you're interested, I'd be happy to start tinkering with a patch for this as well and run a perf comparison.

johno:
Thanks @jescalan!

Yeah, we might have to end up going down the formal parser path for handling import/export/jsx nodes. Though I'm _hoping_ to hold out for [micromark](https://github.com/micromark/micromark) because it'd be able to handle quite a bit more of the tokenization.

One of the big remaining parsing issues we have is not handling JSX blocks with empty lines in them, which we would be able to trivially implement with babel by continuing to "walk" the document when a JSX block opens until its parseable.

jescalan:
Yeah I was thinking it would be used only for the import/export portion of the file initially, not the markdown content. In my logging around, it seemed like there is a spot in the code where the es6 code is separated out from the markdown content. That might be fairly quick to run

PR #362, opened 2018-12-18
Improve default export detection

jescalan:
Closes https://github.com/mdx-js/mdx/issues/341

Definitely preferable to get a true parser in here as is being discussed, but this at least resolves the major bug of disallowing the word "default" anywhere within front matter in the meantime! 😁

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -49,7 +49,8 @@ function toJSX(node, parentNode = {}, options = {}) {
         }
 
         if (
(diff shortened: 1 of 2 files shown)


timneutkens:
@jescalan could you add a test for this? 🙏

jescalan:
Wow, yes I'm sorry for some reason I wasn't even thinking about it. Of course, I'll add one today

jescalan:
Ok @timneutkens - thanks for keeping me in line there. Eliminated a couple unnecessary regex checks and made the tests much more thorough around this use case.

johno:
Thanks!

PR #363, opened 2018-12-19
[WIP] Newline Bugfix

### LINKED ISSUE #340: export feature does not work unless surrounded by newlines
This is a strange issue, but finally found the source after digging for a bit. Basically, if you run an [export, like this](https://mdxjs.com/syntax#export-default), but do not include newlines around the export, it is not detected at all, and will drop the export verbatim as a string into the resulting mdx file. As far as I know, es6 modules do not have any rules around adding extra newlines, so the behavior was pretty confusing to me.

I would be more than happy to submit a patch to fix this, just want to make sure that the maintainer(s) of the library give me the green light first!

To reproduce - this works:

```
import Layout from './Layout'

export default ({ children }) => <Layout some='metadata' >{children}</Layout>

# Hello, world!
```

However, this does not:

```
import Layout from './Layout'
export default ({ children }) => <Layout some='metadata' >{children}</Layout>

# Hello, world!
```

jescalan:
I've just got a test in here to reproduce, not entirely sure what the right direction is from here, any nudges welcome!

Closes #340

### DIFF
--- a/packages/mdx/test/index.test.js
+++ b/packages/mdx/test/index.test.js
@@ -247,6 +247,17 @@ it('Should support semicolons in default export statement', async () => {
   expect(() => parse(result)).not.toThrow()
 })
 
+it('Should export regardless of surrounding newlines', async () => {
+  await expect(
+    mdx(`import layout from './layout'
+export const frontMatter = { foo: 'bar' }
+export default layout({ ...frontMatter })
+
+## Hello world
+This is some content`)
+  ).resolves.not.toContain('export default layout({ ...frontMatter })')
+})
+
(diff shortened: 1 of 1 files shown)


johno:
Looking good so far! I think we're going to have to introduce babel-parse 😬 😆. I can't think of a way to solve this without bringing in a true parser for the JSX.

johno:
Thanks for opening up this PR 💟! I've added a similar test in #399 as part of the babel-parser work so going to close this for now.

jescalan:
Oh yay babel-parser! This is very exciting, thanks for all your good work on this project 😁

PR #364, opened 2018-12-21
Improve inline parsing of JSX

johno:
This pulls in the remark inline HTML parser so that we can customize it to
meet JSX-specific syntax needs.

Adds support for:
- `<>{props.foo}</>`
- `<span style={{ color: 'tomato' }} />`

### DIFF
--- a/packages/remark-mdx/index.js
+++ b/packages/remark-mdx/index.js
@@ -1,5 +1,5 @@
 const isAlphabetical = require('is-alphabetical')
-const {tag} = require('remark-parse/lib/util/html')
+const {tag} = require('./tag')
 
 const IMPORT_REGEX = /^import/
 const EXPORT_REGEX = /^export/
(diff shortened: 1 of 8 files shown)


johno:
> Do we need to update block parsing as well?

Yeah, I plan on making that a separate PR.

> Use remark-mdx inside mdx and the other things.

Yep yep, think that will come in the next day or two.

silvenon:
Omg, this is huge! 👍👍👍

silvenon:
Is it ok to close those issues now or after the release?

wooorm:
There’s still some more work to do. @johno did some work on that. But I’m not sure if there’s a current time frame.

JamesMessinger:
Thanks, @wooorm.   Can you explain what the difference is between this package and the `@mdx-js/mdx` package?  It seems like duplicate work to me, but perhaps I'm missing the bigger picture.

johno:
MDX core will be moving to use `remark-mdx` as the "syntactic extension". The idea is that this package will allow other usages in the wild for MDX syntax (like `prettier` and `remark` formatting). End users won't really even know that this change will happen since it won't affect the public API.

Though, since it's in progress and not currently ready/wired up it isn't being used by the core MDX library yet. I'm hoping to be able to move everything over sometime this week/weekend, maybe next week.

JamesMessinger:
Thanks for the explanation, @johno. 👍  Makes a lot more sense now.  And I'm **super excited** to hear that it's only a week or two away.  ::waits patiently::

PR #366, opened 2018-12-24
Handle backslash escaping in template literals

ChristopherBiscardi:
There's a bug that exists when `<code>` children end in `\` or have "unicode escape sequence-looking"  values like `\unlikely\`.

I haven't been able to fix it yet, but here's a test that throws when these values are used in code blocks or inline.

It looks like these are the issue-causing sequences, which have been changed since ES2016, but only for *tagged* template literals. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates_and_escape_sequences


ChristopherBiscardi:
this is looking good to me. I've been unable to get it to fail so far.

PR #369, opened 2018-12-28
Add `isMDXComponent` static class prop to generated classes

### LINKED ISSUE #368: Checking whether a component is MDX
### Subject of the issue

I'm using Next to serve both `.js` and `.mdx` pages and I couldn't find any (_clean_) way to detect whether the page component in `getInitialProps` or the app `render()` method itself is a MDX component. I also created the page element on the `render` method to check if the component had a `displayName` or any special prop, but I couldn't find any. I also tried to check if that element was `instanceof MDXTag` but it returned false (as the page component doesn't extend MDXTag, it just returns it from the `render` method 😓)

One dirty way I found to check if the component is MDX was `element.toString()` but I would prefer to avoid such a hack.

I need this method to conditionally render a layout and would prefer to not have to declare a boolean export on each `.mdx` file.

The API of a `isMDXElement` method could be similar to [`React.isValidElement()`](https://reactjs.org/docs/react-api.html#isvalidelement). Maybe components produced by `@mdx-js/loader` could have a `static` class property or a named export?

I'm opening this before creating a PR because I believe the approach should be discussed before the implementation.

hugmanrique:
There's currently no way to detect whether a class was generated by `@mdx-js/mdx`, and this PR adds a `isMDXComponent()` static class property to check whether the exported component was generated by mdx.

Closes #368

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -118,8 +118,7 @@ function toJSX(node, parentNode = {}, options = {}) {
                .join('')}
            </MDXTag>
   }
-}
-MDXContent.isMDXComponent = () => true`
+}MDXContent.isMDXComponent = () => true`
     )
   }
   // Recursively walk through children
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -118,7 +118,8 @@ function toJSX(node, parentNode = {}, options = {}) {
                .join('')}
            </MDXTag>
   }
-}MDXContent.isMDXComponent = () => true`
+}
+MDXContent.isMDXComponent = () => true`
     )
   }
   // Recursively walk through children
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -103,7 +103,6 @@ function toJSX(node, parentNode = {}, options = {}) {
       `${
         skipExport ? '' : 'export default'
       } class MDXContent extends React.Component {
-  static isMDXComponent = () => true
   constructor(props) {
     super(props)
     this.layout = ${layout || 'null'}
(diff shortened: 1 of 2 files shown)


hugmanrique:
I'm sorry for all the commit spam, just wanted to try how the GitHub editor experience was 😝

helfi92:
Are there any blockers for this PR? I was wondering if we could have it merged. Many thanks.

silvenon:
Couldn't you just do something like:

```jsx
import Changelog from './changelog.mdx'

Changelog.displayName === 'MDXContent'
// or
Change.displayName.includes('MDX')
```

hugmanrique:
This sadly didn't work for uglified builds where class names get changed.

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -118,7 +118,8 @@ function toJSX(node, parentNode = {}, options = {}) {
                .join('')}
            </MDXTag>
   }
-}`
+}
+MDXContent.isMDXComponent = () => true`
```
johno: I think we can make this a property that's a boolean rather than a function:

```suggestion
MDXContent.isMDXComponent = true`
```

packages/mdx/test/index.test.js
```diff
@@ -106,6 +106,7 @@ it('Should match sample blog post snapshot', async () => {
     );
   }
 }
+MDXContent.isMDXComponent = () => true;
```
johno:
```suggestion
MDXContent.isMDXComponent = true;
```

packages/mdx/test/index.test.js
```diff
@@ -289,6 +290,11 @@ it('Should recognize components as properties', async () => {
   )
 })
 
+it('Should contain static isMDXComponent() function', async () => {
+  const result = await mdx('# Hello World')
+  expect(result).toContain('MDXContent.isMDXComponent = () => true')
```
johno:
```suggestion
  expect(result).toContain('MDXContent.isMDXComponent = true')
```

silvenon:
> This sadly didn't work for uglified builds where class names get changed.

Are you sure you didn't accidentally try `name` instead of `displayName`? AFAIK `displayName` stays as-is after uglifying because it's a property value.

And thanks for your work on this PR, I'm just curious. 😄

johno:
Thanks @hugmanrique!

PR #376, opened 2019-01-21
Retry: Automate canary releases

leimonio:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,8 @@ cache:
   yarn: true
   directories:
     - node_modules
+deploy:
+  provider: script
+  script: yarn publish-ci
+  on:
+    branch: master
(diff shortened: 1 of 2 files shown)


leimonio:
@timneutkens @wooorm it seems like we're missing some configuration here. Is this the first try for automated releases to npm https://travis-ci.org/mdx-js/mdx/jobs/482421219#L592

wooorm:
Hmm, I have no experience with this, so I don’t know. But @johno’s been working on this!

johno:
Ah, yeah I'm assuming we haven't set up an npm token yet on Travis (I know I haven't yet), I can do so this evening. Thank you for implementing this @leimonio!

leimonio:
@johno is this something that has to be enabled on the npm account? Thanks for taking care of it! We can then rerun the worklfow, and check if deploy gets completed.

PR #379, opened 2019-01-23
Pass exports to Layout

### LINKED ISSUE #378: Pass exports to layout component
It would be neat if MDX automatically passed all named exports to the default exported layout component. For example, right now I have to do the following:

```JS
import BlogPost from '../../components/BlogPost';

export const meta = {
  published: true,
  title: 'Tech Choices I Regret at Spectrum',
  // ...more meta data...
}

export default ({ children }) => <BlogPost meta={meta}>{children}</BlogPost>
```

It would be neat if I could do this instead and `BlogPost` would get `meta` passed as a prop automatically (on top of `children` of course):

```JS
import BlogPost from '../../components/BlogPost';

export const meta = {
  published: true,
  title: 'Tech Choices I Regret at Spectrum',
  // ...more meta data...
}

export default BlogPost;
```

That way, my `BlogPost` could have a `getInitialProps` method to fetch data with Next.js. Right now, I have to asynchronously fetch data in the `componentDidMount` lifecycle, which is not the best user experience.

mxstbr:
Closes #378

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -123,7 +123,7 @@ ${skipExport ? '' : 'export default'} class MDXContent extends React.Component {
              name="wrapper"
              ${
                layout
-                 ? `Layout={this.layout} layoutProps={{ ...layoutProps, ...props }}`
+                 ? `Layout={this.layout} layoutProps={Object.assign({}, layoutProps, props)}`
                  : ''
              }
              components={components}>${jsxNodes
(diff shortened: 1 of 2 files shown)


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -112,7 +121,11 @@ function toJSX(node, parentNode = {}, options = {}) {
 
     return <MDXTag
              name="wrapper"
-             ${layout ? `Layout={this.layout} layoutProps={props}` : ''}
+             ${
+               layout
+                 ? `Layout={this.layout} layoutProps={{ ...layoutProps, ...props }}`
```
johno: I think we might need to tweak this to use `Object.assign` or configure `buble` to polyfill:

```suggestion
                 ? `Layout={this.layout} layoutProps={Object.assign(layoutProps, props)}`
```

balupton:
Is this functionality documented anywhere? As the PR isn't that clear on how to make use of it.

silvenon:
@balupton I don't think so, I opened a new issue for this. #418 I'll document it next week.

PR #395, opened 2019-02-11
Don't mark remark-mdx as private

johno:


packages/remark-mdx/readme.md
```diff
@@ -4,15 +4,15 @@
...
 
 ## Installation
 
-This project is not published.
-
-## Usage
-
-This project is not usable.
+```sh
+npm i -S remark-mdx
```
silvenon: I personally prefer writing commands in full in documentation. Also, I think `-S` will get outdated soon because newer versions of npm seem to have `-P`/`--save-prod`, which is the default behavior for some time now, so maybe it's fine to leave it out? Otherwise `--save` might be more familiar name, experience users will know that they can shorten it.

Issue #345, opened 2018-12-06
Use babel-parse for import/export/jsx blocks


johno:
This issue is intended to discuss bringing in [@babel/parse](https://github.com/babel/babel/tree/master/packages/babel-parser) to handle ES/JSX parsing.

I've discussed this a few times with @timneutkens and @ChristopherBiscardi in passing but it's something we haven't been prioritizing. Over the last week I've been thinking about it more heavily after @jescalan reported a bunch of new parse issues.

We've reached a point where most open issues with the [bug label](https://github.com/mdx-js/mdx/issues?q=is%3Aissue+is%3Aopen+label%3Abug) are related to parsing exports and JSX. So we ultimately need to be able to parse import/export/jsx nodes to handle the following scenarios:

- Parse out the `default` export since that maps to the layout
- Walk a JSX node to its completion (right now newlines aren't properly handled)
- Handle all use cases where import/export/JSX blocks aren't separated by empty lines
- Allow (eventually) for the interleaving of [MDX in JSX blocks](https://github.com/mdx-js/specification/issues/11)

There might be a small performance hit, but since it's build time it's not something I'm personally worried about. The big bummer will be the heavier bundle, but that really only comes into play for browser runtimes (which isn't really our targeted env anyway).

- Anyone have any thoughts, insights, or FUD about this?
- Is there anyone that wants to take this on?

silvenon:
I really like this idea because that way we can finally gain the much needed stability. I brought it up before, but a concern for performance was raised, so I'm glad to see interest in it again.

Other than the scenarios in your list, I wonder if we could also solve _inline_ JSX with this. (#222)

My FUD is that we have to parse the content with two different parsers that are mutually incompatible, I'm not sure know how we can combine them in a stable way. E.g. if we're trying to solve the issue with newlines, how do we know how many lines to parse with Babel until we start parsing with remark again?

johno:
> My FUD is that we have to parse the content with two different parsers that are mutually incompatible

Similarly to how we're extending the block parser currently, we'll be able to continue doing the same thing.

> if we're trying to solve the issue with newlines, how do we know how many lines to parse with Babel until we start parsing with remark again

My first naive approach might be to continue walking the document until the babel parser doesn't error. When that occurs we know we either have a completed JSX block or we've reached the end of the document and there's a syntax error.

---

[...]

PR #399, opened 2019-02-12
Improve import and export parsing with babel

johno:
This introduces babel for better handling import and
export parsing. Anything that is an import/export block
will be parsed by babel and split up ensuring that
edge cases will no longer occur based on how the
JS is grouped. It also makes parsing the export default
nodes more robust.

- #345
- #340

### DIFF
--- a/packages/remark-mdx/extract-imports-and-exports.js
+++ b/packages/remark-mdx/extract-imports-and-exports.js
@@ -1,0 +1,51 @@
+const {transformSync} = require('@babel/core')
+const generate = require('@babel/generator').default
+const declare = require('@babel/helper-plugin-utils').declare
+
+const stripTrailingSemi = str => str.replace(/;$/, '')
+
+class BabelPluginExtractImportsAndExports {
+  constructor(code) {
+    const nodes = []
+    this.state = {nodes}
+
+    this.plugin = declare(api => {
+      api.assertVersion(7)
+
+      return {
+        visitor: {
+          ExportDefaultDeclaration(path) {
+            const declaration = path.node.declaration
+            const value = generate(declaration, {}, code).code
+
+            nodes.push({type: 'export', value, default: true})
+          },
+          ExportNamedDeclaration(path) {
+            const value = generate(path.node, {}, code).code
+
+            nodes.push({type: 'export', value: stripTrailingSemi(value)})
+          },
+          ImportDeclaration(path) {
+            const value = generate(path.node, {}, code).code
+
+            nodes.push({type: 'import', value: stripTrailingSemi(value)})
+          }
+        }
+      }
+    })
+  }
+}
+
+module.exports = value => {
+  const instance = new BabelPluginExtractImportsAndExports(value)
+
+  transformSync(value, {
+    plugins: [
+      '@babel/plugin-syntax-jsx',
+      '@babel/plugin-proposal-object-rest-spread',
+      instance.plugin
+    ]
+  })
+
+  return instance.state.nodes
+}
--- a/packages/remark-mdx/extract-imports-and-exports.js
+++ b/packages/remark-mdx/extract-imports-and-exports.js
@@ -1,11 +1,8 @@
 const {transformSync} = require('@babel/core')
-const generate = require('@babel/generator').default
 const declare = require('@babel/helper-plugin-utils').declare
 
-const stripTrailingSemi = str => str.replace(/;$/, '')
-
 class BabelPluginExtractImportsAndExports {
-  constructor(code) {
+  constructor() {
     const nodes = []
     this.state = {nodes}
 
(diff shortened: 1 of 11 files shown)


packages/remark-mdx/index.js
```diff
@@ -102,19 +101,12 @@ function tokenizeEsSyntax(eat, value) {
...
-    return eat(subvalue)({
-      type: 'export',
-      default: isExportDefault(subvalue),
-      value: subvalue
-    })
-  }
+  if (isExport(subvalue) || isImport(subvalue)) {
+    const [first, ...rest] = extractImportsAndExports(subvalue)
+
+    eat(subvalue)(first)
```
wooorm: This way loses position information, which could be a problem for tooling in the remark/mdx ecosystems (not for MDX -> JSX). Now `first` gets the positional info for the whole eaten value, and `rest` gets the position at the end of `value` as it’s start and end positions.

You could use `eat.now()`  first, to get the position we’re at, and then I’d believe babel in `extractImportsAndExports` has access to the positions of the import/export nodes, so you could patch the positions there.
Then finally, you could `eat(subvalue)` to mark the text as parsed, but that not use the returned function, as the nodes already have position info attached?

johno: Ah, nice, yeah I figured I wasn't doing something right there. Will make that fix, thx!

wooorm: It could potentially be a bit hard to figure out the positions, e.g., if the block starts (for remark) at 5:3 (L5C3), you’ll have to do some maths to adjust the Babel nodes 😬

P.S. Babel uses 0-indexed columns, unist uses 1-indexed!

packages/remark-mdx/test/fixtures/import-export.js
```diff
@@ -0,0 +1,21 @@
...
+  {
+    description: 'Separates import from the default export',
+    mdx: [
+      "import Foo from './foo'",
+      'export default props => <article {...props} />'
+    ].join('\n')
+  },
+  {
+    description: 'Handles other types of exports',
+    mdx: 'export const metadata = { some: "stuff" }'
```
ChristopherBiscardi: Do we need to add a case here for exports split across multiple lines?

```
description: 'Handles other types of exports',
mdx: [
  'export const metadata = {'
  '  some: "stuff"'
  '}'
].join('\n')
```

packages/remark-mdx/extract-imports-and-exports.js
```diff
@@ -0,0 +1,51 @@
...
+  transformSync(value, {
+    plugins: [
+      '@babel/plugin-syntax-jsx',
+      '@babel/plugin-proposal-object-rest-spread',
+      instance.plugin
+    ]
+  })
+
+  return instance.state.nodes
+}
```
ChristopherBiscardi: Oh does this mean we can import this stuff from mdx now instead of keeping it in gatsby-mdx?

packages/remark-mdx/test/fixtures/import-export.js
johno: I can add in a test for that 👍 . This still won't handle empty lines yet but I think that will be in my second pass because it will require a bit more complexity to be able to group separate nodes together for a single, valid import/export.

packages/remark-mdx/extract-imports-and-exports.js
johno: Yep yep.

packages/remark-mdx/extract-imports-and-exports.js
```diff
@@ -0,0 +1,62 @@
...
+            const {start} = path.node
+            nodes.push({type: 'import', start})
+          }
+        }
+      }
+    })
+  }
+}
+
+const partitionString = (str, indices) => {
```
wooorm: This function is a bit complex (maybe because of the `.reduce`?) But from glossing over the file I’m not really sure what it does exactly. Maybe a different name, a comment, or a `.map`?

packages/remark-mdx/test/test.js
```diff
@@ -0,0 +1,39 @@
+const unified = require('unified')
+const remarkParse = require('remark-parse')
+const remarkMDX = require('..')
+const mdxAstToMdxHast = require('../../mdx/mdx-ast-to-mdx-hast')
```
wooorm: L3 uses MDX, L4 uses Mdx. Should this be more consistent? no strong feeling but I like L4.

packages/remark-mdx/test/import-export.test.js
```diff
@@ -0,0 +1,10 @@
+const extract = require('../extract-imports-and-exports')
+const fixtures = require('./fixtures/import-export')
+
+fixtures.forEach(fixture => {
+  it(fixture.description, () => {
+    const result = extract(fixture.mdx)
+
+    expect(result).toMatchSnapshot(fixture.result)
+  })
+})
```
wooorm: Should we have a test here to make sure positional info for the nodes is correct?

packages/remark-mdx/extract-imports-and-exports.js
wooorm: This isn’t blocking btw, but just thought I’d mention it

johno:
Good calls @wooorm. Addressed in latest commits.

PR #401, opened 2019-02-15
Replace MDXTag with custom pragma

ChristopherBiscardi:
* Enables simpler support for alternative UI tooling like Vue (need to write a new pragma function).
* Removes the awkward behavior when trying to replace an element via MDXProvider (for more information, [read how this currently behaves](https://www.christopherbiscardi.com/post/codeblocks-mdx-and-mdx-utils) with the MDXTag in between `pre` and `code`.
* Enable the replacement of base JSX tags as well as their markdown counterparts. In the following case, the `MDXProvider` component for an `h2` would replace the markdown *and* the jsx element, whereas the current `MDXTag` implementation only lets us control rendering of the markdown second level heading.
  ```markdown
  ## some h2

  <h2>some other h2</h2>
  ```

### Interesting tidbits and screenshots

React Devtools will show the original tag and what was rendered instead by createElement, leading to better debugging.
image


johno:
@ChristopherBiscardi do you mind temporarily including the `my-site-2` workspace code? Would be cool to take this for a spin.

ChristopherBiscardi:
@johno done.

ChristopherBiscardi:
Gave names to the components that happen in the createElement call. `MDXCreateElement` and `MDXCreateElementInner` so they show up in React DevTools. Layout is swapped over to new approach as well.

image

#### TODO:
- [ ] make sure wrapper is replaceable

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -159,9 +156,9 @@ MDXContent.isMDXComponent = true`
       props = JSON.stringify(node.properties)
     }
 
-    return `<MDXTag name="${node.tagName}" components={components}${
-      parentNode.tagName ? ` parentName="${parentNode.tagName}"` : ''
-    }${props ? ` props={${props}}` : ''}>${children}</MDXTag>`
+    return `<${node.tagName} ${
+      parentNode.tagName ? `parentName="${parentNode.tagName}"` : ''
+    }${props ? ` props={${props}}` : ''}>${children}</${node.tagName}>`
```
ChristopherBiscardi: might want to change this props= to a spread.

ChristopherBiscardi:
Looking good to me.

alexandernanberg:
@ChristopherBiscardi Will this effect how `MDXProvider` works? E.g. right now I'm mapping the `wrapper` component to `React.Fragment` to avoid having a wrapping `div`. Will that still work? I know it's undocumented but it's a really nice feature imo!

johno:
The wrapper will still behave as it currently does since the mapping will behave the same. It will even render out more cleanly in React devtools because it won't be shown as "MDXTag".

alexandernanberg:
Ah gotcha, thanks!

PR #405, opened 2019-02-15
Grouped imports shouldn't have empty newline separation

johno:
Before, each import is turned into its own node, even when grouped like so:

```js
import Foo1 from "./foo"
import Foo2 from "./foo"
import Foo3 from "./foo"
```

So, when combining back to a string we ended up with:

```js
import Foo1 from "./foo"

import Foo2 from "./foo"

import Foo3 from "./foo"
```

We now combine adjacent nodes of the same type. Though, if the node is a default export it is treated specially.

### DIFF
--- a/packages/remark-mdx/extract-imports-and-exports.js
+++ b/packages/remark-mdx/extract-imports-and-exports.js
@@ -49,8 +49,42 @@ module.exports = value => {
   const nodeStarts = sortedNodes.map(n => n.start)
   const values = partitionString(value, nodeStarts)
 
-  return sortedNodes.map(({start: _, ...node}, i) => {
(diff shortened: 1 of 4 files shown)


wooorm:
This looks a bit unwieldy! I thought I’d try to wrap my head around what’s going on to see what I’d come up. It’s pseudo-code, so it probably won’t work, but it looks a bit cleaner!

```js
var merged = []
var length = allNodes.length
var index = -1
var prevType
var currType
var node
var head

while (++index < length) {
  node = allNodes[index]
  currType = node.default ? 'default' : node.type

  if (currType !== prevType) {
    merged.push(node)
    head = node
  } else {
    head.value += node.value
  }

  prevType = currType
}
```

PR #408, opened 2019-02-22
Remove destructuring props to prevent shadowing 'components' variable

### LINKED ISSUE #206: MDXProvider doesn't seem to work in a custom layout
This might be by design, but the MDXProvider doesn't appear to work when included in a custom layout, e.g.:

```jsx
import React from 'react'
import { MDXProvider } from '@mdx-js/tag'
import components from './components'

export default ({ children }) =>
  <MDXProvider components={components}>
    {children}
  </MDXProvider>
```

filoxo:
Closes #206 

Removes destructuring of `components` from props which was being shadowed after transpiling.

### DIFF
--- a/packages/tag/src/mdx-provider.js
+++ b/packages/tag/src/mdx-provider.js
@@ -2,16 +2,16 @@ import React from 'react'
 
 const {Provider, Consumer} = React.createContext({})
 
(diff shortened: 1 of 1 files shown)


johno:
Thanks!

PR #409, opened 2019-02-25
Don't destructure props in transpiled JSX

johno:
Related #408

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -117,16 +117,14 @@ ${skipExport ? '' : 'export default'} class MDXContent extends React.Component {
     this.layout = ${layout || 'null'}
   }
   render() {
-    const { components, ...props } = this.props
-
     return <MDXTag
              name="wrapper"
              ${
                layout
-                 ? `Layout={this.layout} layoutProps={Object.assign({}, layoutProps, props)}`
+                 ? `Layout={this.layout} layoutProps={Object.assign({}, layoutProps, this.props)}`
                  : ''
              }
-             components={components}>${jsxNodes
+             components={this.props.components}>${jsxNodes
                .map(childNode => toJSX(childNode, node))
                .join('')}
            </MDXTag>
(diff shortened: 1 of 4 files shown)


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -117,16 +117,14 @@ ${skipExport ? '' : 'export default'} class MDXContent extends React.Component {
...
   }
   render() {
-    const { components, ...props } = this.props
-
     return <MDXTag
              name="wrapper"
              ${
                layout
-                 ? `Layout={this.layout} layoutProps={Object.assign({}, layoutProps, props)}`
+                 ? `Layout={this.layout} layoutProps={Object.assign({}, layoutProps, this.props)}`
```
silvenon: I'm concerned about the way [`layoutProps` are being applied in `MDXTag`](https://github.com/mdx-js/mdx/blob/3913c39282a66d988498c3f12ed2cec58992eb58/packages/tag/src/mdx-tag.js#L30), which allows for overriding its `components` prop in two ways in this case:

1. `MDXContent`'s `this.props.components` (which is what's happening here)
2. `layoutProps.components` if the MDX file has `export const components = ...` somewhere

Even if `this.props.components` will always be the same as `components` in `MDXTag`, it doesn't feel safe and doesn't seem like something we should count on. Also, I think `layoutProps.components` will always be overridden by `this.props.components`, so if people happen to be using that export, that problem will be really hard to debug.

I know this is not within the scope of this PR, but I think this is a good time to develop a strategy for more predictable behavior.

johno: Hmmm, yeah that's a good point and is something that can be unpredictable. AFAICT, though, with the [new pragma implementation](https://github.com/mdx-js/mdx/blob/v1/packages/mdx/create-element.js) and [context](https://github.com/mdx-js/mdx/blob/v1/packages/tag/src/mdx-context.js) it _should_ be less of a footgun.

I am realizing, though, that would could figure out a way to use some magic components key like `__MDX_CONTEXT_COMPONENTS__` once passed in at the top level. It will also become easier to handle with `useContext`.

johno:
Going to close as a result of [this discussion](https://github.com/mdx-js/mdx/pull/409#discussion_r261267241). We can hold off on fixing this until v1 where we can more safely pass a key that won't conflict in user land.

Issue #410, opened 2019-02-25
Allow for merging of components passed in context


johno:
There are use cases with MDX where it would be nice to be able to nest MDXProviders so that a subset of components can be overridden within an MDX document. This is currently achievable, but requires direct usage of `withMDXComponents` which isn't a great experience for the end user.

One example of this need is with `gatsby-mdx` and Gatsby Themes where multiple themes might want to merge in a [subset of components](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/292) to be rendered.

CSS in JS libraries like styled-components offer the ability to [pass a function](https://www.styled-components.com/docs/advanced#function-themes) instead of an object which then receives any context from above (thx @jxnblk for pointing me to this). This is an approach we could take and comes with the advantage of not being a breaking change.

### Pseudocode example

Here all components will be rendered from the root MDXProvider except for any `ul` elements. Those will be overridden by the `faqComponents` function which merges in its own implementation.

```js
const faqComponents = components => ({ ...components, ul: FaqList })

return (
  <MDXProvider components={baseComponents}>
    <MDXProvider components={faqComponents}>
      {children}
   </MDXProvider>
  </MDXProvider>
)
```

johno:
After some thinking and discussion with @jxnblk and @ChristopherBiscardi we believe it makes the most sense to automatically merge component context. The functional form can be used for advanced use cases that might want to opt out of merging or customize how the merge occurs.

johno:
This has landed in the v1 alpha.

PR #411, opened 2019-02-25
Allow for merging of components passed to context with a function

johno:
If a function is passed to the theme provider it is invoked
with the out context's components. This allows users to opt
in to merging those components.

Related #410


johno:
This only handles half of the intended functionality for v1 since it doesn't merge outer context objects yet, but I'll go ahead and merge this for now.

PR #413, opened 2019-02-26
Fix newline issue in paragraphs

### LINKED ISSUE #412: newlines should always render as a space
### Make it possible to render a space for newlines between markdown

Example input:

```markdown
[this is a link](https://google.com/a-very-long-link-that-takes-an-entire-line-by-itself)
[this is another link](https://google.com/which-is-also-a-very-long-link-that-takes-an-entire-line-by-itself)
```

This will be rendered as:

```html
<a href="https://google.com/a-very-long-link-that-takes-an-entire-line-by-itself">this is a link</a><a href="https://google.com/which-is-also-a-very-long-link-that-takes-an-entire-line-by-itself">this is another link</a>
```

The solution is to add a space to the end of the first line or place them on the same line with a space between them, but prettier formats it the way it is above if you have "proseWrap" set to "always" which I do so I can reasonably edit the markdown.

So I _think_ this is an issue with how mdx converts those. I'm not sure how to solve this problem... I would like to make it possible to configure MDX to add a newline (or maybe this is a bug that can be fixed?)

### Your environment

N/A

### Steps to reproduce

See above. Copy/paste that into [the demo](https://mdxjs.com/)

### Expected behaviour

I would like the newline to be interpreted as a space when two links are side-by-side (like it is for other things).

### Actual behaviour

No space is inserted.

### More examples:

```markdown
_italics_
**bold**
```

That gets converted to:

```html
<em>italics</em><strong>bold</strong>
```

But this:

```markdown
italics
bold
```

Gets converted to:

```html
italics bold
```

Which I think is correct. I'd basically just like that space to be applied in all cases.

johno:
In paragraph nodes newlines were being removed which causes
issues when there are two subsequent nodes like links, spans,
emphasis, bold, etc.

Now, if the parent node is a paragraph tag, we preserve the
newline since it's significant and results in a space that
separates two inline nodes on render.

---

Closes #412

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -168,9 +168,13 @@ MDXContent.isMDXComponent = true`
   if (node.type === 'text') {
     // Don't wrap newlines unless specifically instructed to by the flag,
     // to avoid issues like React warnings caused by text nodes in tables.
-    if (node.value === '\n' && !preserveNewlines) {
+    const shouldPreserveNewlines =
+      preserveNewlines || parentNode.tagName === 'p'
+
+    if (node.value === '\n' && !shouldPreserveNewlines) {
       return node.value
     }
+
     return '{`' + node.value.replace(/`/g, '\\`').replace(/\$/g, '\\$') + '`}'
   }
 
--- a/packages/mdx/test/index.test.js
+++ b/packages/mdx/test/index.test.js
@@ -204,6 +204,24 @@ Some text <!-- an inline comment -->
   expect(result).toContain('<!-- a template literal -->')
 })
 
+it('Should turn a newline into a space with adjacent anchors', async () => {
+  const result = await renderWithReact(`
+  [foo](/foo)
+  [bar](/bar)
+  `)
+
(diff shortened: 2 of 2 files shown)


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -168,9 +168,13 @@ MDXContent.isMDXComponent = true`
   if (node.type === 'text') {
     // Don't wrap newlines unless specifically instructed to by the flag,
     // to avoid issues like React warnings caused by text nodes in tables.
-    if (node.value === '\n' && !preserveNewlines) {
+    const shouldPreserveNewlines =
+      preserveNewlines || parentNode.tagName === 'p'
```
wooorm: Hmm, there are other nodes that accept [“phrasing” content](https://github.com/syntax-tree/mdast), but aren’t. Headings and list-items as well, so would it be maybe be needed to add their HTML equivalents here as well?

wooorm: If so, there should be tests for that as well!

johno: Good idea, done ✅

kentcdodds:
Thank you so much!

PR #427, opened 2019-03-01
Convert MDXContent to a function component

### LINKED ISSUE #389: Convert Compiled MDXContent Class back to Function
in v0.16 we change the compiled MDX to be a class to solve some rendering issues and stabilize the reference to layouts defined inline.

In order to support hooks, etc in MDX content, we should convert it back to a function while maintaining the stable layout reference.

alexandernanberg:
Closes #389

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -71,11 +71,10 @@ function toJSX(node, parentNode = {}, options = {}) {
       `const layoutProps = {
   ${exportNames.join(',\n')}
 };
+const Layout = ${layout || 'null'}
 ${
   skipExport ? '' : 'export default'
 } function MDXContent({ components, ...props }) {
-  const Layout = ${layout || 'null'}
-
   return (
     <div
       name="wrapper"
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -71,7 +71,7 @@ function toJSX(node, parentNode = {}, options = {}) {
       `const layoutProps = {
   ${exportNames.join(',\n')}
 };
-const Layout = ${layout || 'null'}
+${layout ? `const MDXLayout = ${layout}` : ''}
 ${
   skipExport ? '' : 'export default'
 } function MDXContent({ components, ...props }) {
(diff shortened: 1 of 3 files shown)


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -111,23 +111,20 @@ function toJSX(node, parentNode = {}, options = {}) {
...
-             components={components}>
-             ${layout ? `<Layout {...layoutProps} {...props}>` : ''}
-             ${jsxNodes.map(childNode => toJSX(childNode, node)).join('')}
-             ${layout ? `</Layout>` : ''}
-           </div>
-  }
+${
+  skipExport ? '' : 'export default'
+} function MDXContent({ components, ...props }) {
+  const Layout = ${layout || 'null'}
```
johno: I think we're going to need to place the layout outside of the function because it will be unstable if the default export is an inlined function:

```md
# Hello, world!

The following function will result in an unstable component:

export default props => <main {...props} />
```

I wonder if we need to do something like:

```js
const MDXLayout = ${layout || 'null'}

// ...

<div
  name="wrapper"
  components={components}>
  <MDXLayout {...layoutProps} {...props}>
     {jsxNodes}
  </MDXLayout>
</div>
```

silvenon: Definitely! No reason to initialize it on each render, it should be defined outside.

alexandernanberg: Done! Can change the name to `MDXLayout` if you'd like as well, might be a bit clearer.

packages/mdx/create-element.js
```diff
@@ -0,0 +1,59 @@
+const React = require('react')
+const {withMDXComponents} = require('@mdx-js/tag')
+
+const TYPE_PROP_NAME = '__MDX_TYPE_PLEASE_DO_NOT_USE__'
```
silvenon: In case someone sees this and is concerned that this is some kind of a bug, we could make it clearer that this is for internal use only, something like `__INTERNAL_MDX_TYPE__ ` or `__INTERNAL_MDX_TYPE_DO_NOT_USE_DIRECTLY__`.

alexandernanberg:
@silvenon I think @johno rebased this branch which caused some commits from the v1 branch to be included in this PR. I guess you can check my individual commits instead of the files changed until we fix it

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -112,26 +112,19 @@ function toJSX(node, parentNode = {}, options = {}) {
...
-               layout
-                 ? `Layout={this.layout} layoutProps={Object.assign({}, layoutProps, props)}`
-                 : ''
-             }
-             components={components}>${jsxNodes
-               .map(childNode => toJSX(childNode, node))
-               .join('')}
-           </MDXTag>
-  }
+const Layout = ${layout || 'null'}
```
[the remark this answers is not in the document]
alexandernanberg: Done!

alexandernanberg:
Rebased to get rid of the other commits

johno:
Thank you!

PR #429, opened 2019-03-02
Enable remark plugins that insert JSX and HTML

ChristopherBiscardi:
Changed `html`-to-`jsx` node relabelling to happen before remark
plugins are applied instead of after remark plugins are applied. This,
combined with the parsing of `raw` nodes in `HAST`, allows remark
plugins to insert `jsx` nodes AND `html` nodes.

Previously, all `html` nodes were indiscriminately relabelled as
`jsx`, which should only be true for the user-layer when manually
writing MDX and not when `operating on the MDX ASTs with plugins.

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -1,6 +1,8 @@
 const unified = require('unified')
 const toMDAST = require('remark-parse')
 const squeeze = require('remark-squeeze-paragraphs')
+const visit = require('unist-util-visit')
+const raw = require('hast-util-raw')
 const toMDXAST = require('./md-ast-to-mdx-ast')
 const mdxAstToMdxHast = require('./mdx-ast-to-mdx-hast')
 const mdxHastToJsx = require('./mdx-hast-to-jsx')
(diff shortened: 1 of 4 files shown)


GerardKetuma:
@ChristopherBiscardi @johno This PR breaks gatsby-remark-responsive-iframe when used with gatsby-mdx. I narrowed it down to this by upgrading the tag of this package (mdx-js/mdx) from 0.17.5 where it was working until 0.20.0-alpha where this PR is merged and it breaks the iframe in my mdx. The iframe renders as is and it does not become responsive anymore. But the iframe still renders but with the width and height specified in the mdx file.

ChristopherBiscardi:
@GerardKetuma `gatsby-remark-responsive-iframe` used a faulty node type to do its AST manipulation, [we fixed it upstream](https://github.com/gatsbyjs/gatsby/commit/bb4690545e777b24caa7c91caab8e068df85320d) so you'll have to upgrade that plugin.

GerardKetuma:
@ChristopherBiscardi Thanks for the update. I'll do the update of the plugin. Also thanks for all your hard work on all things MDX related.

PR #432, opened 2019-03-02
Add example for syntax highlighting and live editor

### LINKED ISSUE #253: Add demo for code syntax highlighting
In #221 there was a great suggestion to add a demo for how to implement syntax highlighting, most probably it will be using [@mapbox/rehype-prism](https://github.com/mapbox/rehype-prism). I will add this demo, but first I need to solve the problem outlined in https://github.com/mdx-js/mdx/issues/221#issuecomment-419183553.

johno:
Fixes #253


endiliey:
Hi,

We're looking to switch to MDX from remarkable for [facebook/Docusaurus](https://github.com/facebook/Docusaurus) v2.

Can we have an example for those who wanted to use mdx + syntax highlighting via webpack-loader ?
This example seems to be on how to use it on Gatsby.

Something like this:

```js
const rehypePrism = require('@mapbox/rehype-prism');

module.exports = {
    test: /(\.mdx?)$/,
    use: [
        {
            loader: '@mdx-js/loader',
            options: {
                hastPlugins: [rehypePrism]
            }
        }
    ]
};
```

johno:
Yeah! I've been meaning to document this better so I've created an issue #446

PR #440, opened 2019-03-06
Add useMDXComponents hook

alexandernanberg:
Some notable changes:
- Adds a `useMDXComponents` hook
- Replaces the usage of `withMDXComponents` with  `useMDXComponents`. That will allow us to remove the `MDXCreateElementInner` component to get a "flatter" tree.
- The argument passed to the `useMDXComponents` hook can either be an object or function. So `MDXTag` and `MDXCreateElement` will work the same way as `MDXProvider`. (we might not want this?)

### DIFF
--- a/packages/mdx/create-element.js
+++ b/packages/mdx/create-element.js
@@ -1,5 +1,5 @@
 const React = require('react')
-const {withMDXComponents} = require('@mdx-js/tag')
+const {useMDXComponents} = require('@mdx-js/tag')
 
 const TYPE_PROP_NAME = '__MDX_TYPE_PLEASE_DO_NOT_USE__'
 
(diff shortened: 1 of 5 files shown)


ChristopherBiscardi:
dope. Looks like a good idea to me. `MDXTag` isn't used in the 1.0 branch any more at all, so we should probably just remove that file instead of updating mdx-tag to use context.

ChristopherBiscardi:
also it looked like the snapshot tests needed to be updated

alexandernanberg:
Oh wasn't aware of that! Should I revert the changes of the `MDXTag` file or delete it in this PR? Hmm one of the snapshot errors seem to occur in the v1 branch as well. Also not sure why it added an extra line in the other snapshot 🤔

johno:
There is an existing failing test in v1 (from me 🙀 😬).

johno:
I'll go ahead and merge this as is and make an issue to remove MDXTag. Thank you!

PR #441, opened 2019-03-06
Automatically merge outer context for components in the provider

### LINKED ISSUE #410: Allow for merging of components passed in context

johno:
Closes #410

### DIFF
--- a/packages/tag/test/mdx-provider.test.js
+++ b/packages/tag/test/mdx-provider.test.js
@@ -42,7 +42,7 @@ it('Should merge components when there is nested context', () => {
 
   const result = renderToString(
     <MDXProvider components={components}>
-      <MDXProvider components={components => ({...components, h2: CustomH2})}>
+      <MDXProvider components={{h2: CustomH2}}>
         <MDXTag name="h1" />
         <MDXTag name="h2" />
       </MDXProvider>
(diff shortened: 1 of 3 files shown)


packages/tag/test/mdx-provider.test.js
```diff
@@ -55,3 +55,22 @@ it('Should merge components when there is nested context', () => {
   // MDXTag picks up overridden component context
   expect(result).toMatch(/style="color:papayawhip"/)
 })
+
+it('Should merge components when there is nested context as a function', () => {
```
ChristopherBiscardi: Test name combined with what's under test makes it feel like the outer h1 component should still be used, but I don't think that's true?

Issue #238, opened 2018-08-24
Vue Support


brennj:
Hey guys, I'm in love with this library. Its super cool and easy to use. I use Vue in my day job, and I would really love to be able to write some component docs with this, because as far as I know - there is nothing in the Vue world for this, even though we have JSX.

I think its because of the differences that Vue needs, in comparison to being able to define a functional React component that makes it slightly non-trivial? I'm rusty with my parsers, so not sure how another layer of abstraction ontop of parsing with deciding to use a JSX outputter is.

i.e.
```
export default {
  render() {
    return <div>Hello World</div>;
  }
};
```

in comparison to:
```
const MyComponent = () => <div>Hello World</div>
```

I miss React. :) Would love to understand whether it was possible or not.

Thanks guys!

silvenon:
I have zero experience with Vue, but MDX has a `compilers` option which is an array of processors that operate on JSX. Maybe you could add a custom one there which modifies the final JSX output into something Vue-friendly?

What would really help is if you could take a look at JSX that MDX outputs and outline which changes in that JSX would be necessary for it to work in Vue.

brennj:
I'm having a look into it @silvenon! As far as I understand theres a few tasks:

1. Make compatible `MDXTag` and `MDXProvider` Vue components (I've worked on this today over at  https://github.com/brennj/mdx/tree/vue-plugin ! I've written some tests to match the React side and looks good so far, maybe)
2. Like you say, pass a custom compiler that will change the following code:

```
export default ({components, ...props}) => <MDXTag name="wrapper"  components={components}><MDXTag name="p" components={components}>{`Hello World!`}</MDXTag></MDXTag>
```

into Vue friendly code. I'm still trying to figure this bit out as theres some specific React parts in the loader and stuff but I think it should be fairly fine. I might see how far I can get and offer up a PR if I'm happy with it :)

johno:
I have a PR open (https://github.com/mdx-js/mdx/pull/455) that gets us a lot of the way there thanks to the fine work of @brennj.

I'm currently stuck on the way Vue handles element creation (I've never really used it before 🙃). I've created the custom pragma function for Vue to receive the type and props which is happening as expected. But, for whatever reason I can't seem to get it to render.

If anyone wants to tinker around in that branch there's a directory (`examples/vue`) you can get running and then edit `packages/vue/create-element.js`. Would love any insights as to what I'm doing wrong.

ChristopherBiscardi:
[...]

PR #455, opened 2019-03-08
Implement basic Vue support

johno:


Aaron-Pool:
Super excited to see this happening @johno Thank you so much! Again, if you need any help with anything, let me know!

johno:
Thanks @Aaron-Pool! If you know much about how Vue's `createElement` works I could use some help/insight here: https://github.com/mdx-js/mdx/issues/238#issuecomment-470816311

Everything is working up to here: https://github.com/mdx-js/mdx/blob/b0ede790c87e9759bba154a8c5c1ab7c02c41032/packages/vue/src/create-element.js

Aaron-Pool:
@johno I've had to sludge through some vue source code to work out bugs before, so I'd be glad to take a look! I'll get back to you if I find anything!

johno:
Thanks! Here's where I'm currently at:

It seems like the compiled MDX code:

```js
const layoutProps = {};

export default function MDXContent({ components, ...props }) {
  return (
    <div
      name="wrapper"
      components={components}>
      <h1 >{`Hello, world!`}</h1>
      <p >{`Here's more stuff`}</p>
      <div style={{ padding: '21px', backgroundColor: 'tomato' }} />
    </div>
  )
}
MDXContent.isMDXComponent = true
```

Is somehow incompatible with Vue? Though create element _is_ getting called for the JSX elements:

```js
import Vue from 'vue'

export default (type, props, children) => {
  console.log({ type, props, children })

  return Vue.component('mdx-component', {
    render: function(createElement) {
      return createElement(type, props, children)
    }
  })
}
```

But it's like I have to invoke the component twice:

```js
<template>
  <div id="app">
    <Test />
  </div>
</template>

<script>
import Test from './test.mdx'

console.log(Test)

export default {
  name: 'app',
  components: {
    Test: Test({})
  }
}
</script>
```

And even then it's not rendering anything to the DOM, but an empty MDXContent _does_ show up in Vue devtools.

johno:
Also, I've just pushed the latest code I'm working with to the `vue-support` branch just now. Thanks for looking into it 💟

Aaron-Pool:
Great! I'll look into it as soon as I can! Hopefully when I get a free minute at work, but if not, I hope to dig into it some tonight.

Aaron-Pool:
Alright, I looked things over, let's touch on a few things:

First off, you should go ahead and swap out your `babel-plugin-transform-vue-jsx` plugin to use the new [Vue Jsx Babel Preset](https://github.com/vuejs/jsx) which is set to replace it. It's in beta right now, but it's almost done and it's what you're "supposed to use" when you're using babel 7. In fact, it's what the official cli uses, as of a few releases ago. There are no breaking changes, everything should "just work". It's a good idea because, first and foremost, it will make sure the project stays up to date with the Vue jsx ecosystem. Secondly, it ensures that functional components work the way you're expecting.

The compiled code you gave me here

```js
const layoutProps = {};

export default function MDXContent({ components, ...props }) {
  return (
    <div
      name="wrapper"
      components={components}>
      ...etc
    </div>
  )
}
MDXContent.isMDXComponent = true
```
_may_ work with the old babel-plugin, but I'm not entirely sure it will. I think the first babel-plugin may have required you return an object with a render function, for example:

```js
const layoutProps = {};

export default function MDXContent({ components, ...props }) {
  return {render() { return (
    <div
      name="wrapper"
      components={components}>
      .... etc
    </div>
  });
}
MDXContent.isMDXComponent = true
```

Also, on that topic, why are you giving a `component` prop to a `div` here? I know MDX does some stuff to replace primitive elements with components of the same name, but I don't see any of that being imported in the output above, so how is that handled?

Lastly, this could be me being an idiot, and not understanding your approach, but I think maybe your background with react is throwing you off in the way you're declaring the components here:

```js
export default {
  name: 'app',
  components: {
    Test: Test({})
  }
}
```

The `components` property of a vue instance doesn't declare component instances, rather it defines the component definitions to be used when the template is being parsed, so I think what you want is just:

```js
export default {
  name: 'app',
  components: { Test }
}
```

Which tells vue, "hey, if you see the tag with the same name as this property, treat it like a vue component with this definition as the value".

johno:
Thank you for the thorough review and explanation!

> First off, you should go ahead and swap out your babel-plugin-transform-vue-jsx plugin to use the new Vue Jsx Babel Preset which is set to replace the old babel plugin.

👍 

> may work with the old babel-plugin, but I'm not entirely sure it will. I think the first babel-plugin may have required you return an object with a render function

Ahhhh, I think this is what I have wrong. I'll need to change MDX core to have the ability to return that code variation.

> Also, on that topic, why are you giving a component prop to a div here? I know MDX does some stuff to replace primitive elements with components of the same name, but I don't see any of that being imported in the output above, so how is that handled?

Yeah, that isn't currently implemented, but with MDX there's a notion of a "component provider" which `createElement` uses. This allows Markdown elements to render components and even be composed (so you can use your own fancy code block for example). When I get things rendered to the page that will be the next step. Looks like this is the API I need https://vuejs.org/v2/api/#provide-inject

> The components property of a vue instance doesn't declare component instances, rather it defines the component definitions to be used when the template is being parsed

Yeah, that's just a random hack I added when fiddling. Think that's a result of the not returning the object with a render function you mentioned above.

---

🙏 🙏 🙏 🙏 🙏 🙏 🙏 

This is _extremely_ helpful, thanks for taking the time. Think I have what I need to give it another shot later today.

Aaron-Pool:
@johno Great! Although, like I said, as long as you upgrade your jsx plugin, you shouldn't _need_ to change anything in core. The newer vue-jsx preset has support for functional jsx components out of the box 👍 

Good luck! I'll be checking back in throughout the day, if anything else comes up. I _really_ like the mdx dev experience (particularly when integrated with things like docz), and I'm glad to lend a hand to get into my vue wheelhouse 😄 

Also! Just a thought, if you wanted to potentially, peel away any unnecessary complexity (and also maybe use an approach you're more comfortable with), I would start getting a test working with just vue render functions (guide [here](https://vuejs.org/v2/guide/render-function.html)). The vue loader just ends up compiling the single file component into a render function anyways, so you're probably just making things harder for yourself getting a SFC example to work before setting up a render function example.

johno:
> Although, like I said, as long as you upgrade your jsx plugin, you shouldn't need to change anything in core. The newer vue-jsx preset has support for functional jsx components out of the box 👍

Ohhh, thanks for clarifying, I misread. That's great news because I was surprised it would need to be handled differently (being vanilla JSX and all). Now it makes sense again. 😹

johno:
I'm going to merge this for now and add another todo for code cleanup and handling component context. Thanks all!

brianzelip:
Hi @johno. Regarding @Aaron-Pool 's comment above:

> I would start getting a test working with just vue render functions (guide [here](https://vuejs.org/v2/guide/render-function.html)) before I worried about vue Single File Components.

Here's an example from my [WIP attempt at getting mdx v1 into Vue](https://github.com/brianzelip/markdown-vue/blob/75757b5618c50414f1c87b258b33bd566b2b7904/src/components/App.js):

```js
// src/components/App.js

import TheBio from '../content/bio.mdx';
import TheFooter from './TheFooter.js';

export default {
  name: 'App',
  components: { TheBio, TheFooter },
  render() {
    return (
      <main>
        <h1>Brian Zelip</h1>
        <TheBio />
        <TheFooter />
      </main>
    );
  }
};

```

The mdx might not be handled right above, but the component format is correct for a Vue render function.

BTW - mdx still not rendering. The parcel cli warning given is:

> With `useBuiltIns` option, required direct setting of `corejs` option

PR #470, opened 2019-03-15
Change wrapper default element to React.Fragment

### LINKED ISSUE #460: Make the wrapper component a React.Fragment as default
### Subject of the issue

I think it would make sense to change the default wrapper component to a `React.Fragment` instead of a `div`.  I almost always have to do this manually and see no reason why you would need a `div` as the default.

alexandernanberg:
Closes #460 

Not 100% sure that I made the change in the right place. I guess you could make the change in `mdx-hast-to-jsx.js` as well?

### DIFF
--- a/packages/mdx/create-element.js
+++ b/packages/mdx/create-element.js
@@ -5,7 +5,7 @@ const TYPE_PROP_NAME = 'mdxType'
 
 const DEFAULTS = {
   inlineCode: 'code',
-  wrapper: 'div'
+  wrapper: React.Fragment
 }
 
 const MDXCreateElement = ({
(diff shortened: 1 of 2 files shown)


alexandernanberg:
Yeah, done some digging and not sure how to test this in a good way... The `renderWithReact` util doesn't seem to change the wrapper element 🤔 Got any ideas?

johno:
As discussed in #480 I think this was a v1 regression. `renderWithReact` should be sufficient to test this AFAICT (though you'll need to rebase against v1 again).

alexandernanberg:
I don't know if it'd be better to change one of the existing test instead... let me know if you want me to change that!

johno:
Published as `1.0.0-alpha.9`.

PR #473, opened 2019-03-19
Delete ._app.js.swp

karlhorky:
Hi there, thanks for this awesome project! Really awesome for slides and static sites. Just a quick PR to remove a swap file, because it appears that it has been committed in error.


timneutkens:
Thanks!

PR #475, opened 2019-03-19
 Add wrapExport option for wrapping exported function

pshrmn:
Add a `memo` compiler option to wrap the generated component in a `React.memo` call. This could help prevent unnecessary renders if any of the components define objects/arrays/functions inline.

```js
mdx(content, {
  memo: true
})
```

```jsx
function MDXContent({ components, ...props }) {
  // ...
}
export default React.memo(MDXContent);
```

I had to re-organize that code strings a little bit in order to do this. Hopefully that is okay because doing it with template expressions was a bit messy.

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -7,7 +7,8 @@ function toJSX(node, parentNode = {}, options = {}) {
   const {
     // Default options
     skipExport = false,
-    preserveNewlines = false
+    preserveNewlines = false,
+    wrapExport
   } = options
   let children = ''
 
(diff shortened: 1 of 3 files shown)


packages/mdx/test/index.test.js
```diff
@@ -379,7 +387,7 @@ export default function MDXContent({ components, ...props }) {
       name=\\"wrapper\\"
       components={components}>
       <MDXLayout {...layoutProps} {...props}>
-      
+
```
pshrmn: This was an issue with VSCode. I disabled "Trim Trailing Whitespace", but it insisted on deleting this line every time I saved the file... The test passes, so hopefully this is fine.

pshrmn:
Argh, the failing test has to do with the whitespace change. I'll have to see if I can't finagle VSCode or make the changes in a different editor if that fails.

packages/mdx/test/index.test.js
```diff
@@ -296,6 +296,14 @@ it('Should render elements without wrapping blank new lines', async () => {
   expect(result).not.toContain('{`\n`}')
 })
 
+it('Should wrap export in React.memo when memo option is true', async () => {
+  const result = await mdx(`# Test`, {
+    memo: true
```
ChristopherBiscardi: why add an option instead of always doing this?

pshrmn: That could work too. I wasn't sure if there was a reason that it isn't being done already, so I figured that an option would be the safe route.

johno: The one thing that I'm seeing here, though, is `React.memo` will make the default output coupled to React again. I wonder if we should somehow wrap with `React.memo` in the loader?

pshrmn: Oh yeah, I had misinterpreted how the `vue-loader` works and was thinking that it uses a different HAST to JSX function. A few possible solutions that come to mind:

1. Use a `wrapExport` option instead of `memo`.
```js
const result = await mdx(content, {
  wrapExport: "React.memo"
})
```

2. Have `mdx` return separate file parts (either the `MDXContent` function and everything else or something more granular). This would allow more freedom in how the loaders output file contents, and would make the `skipExport` option moot, but also has a larger potential to break things.

```js
const { fn, rest } = await mdx(...);
const code = `/* @jsx mdx */
import React from 'react'
import mdx from '@mdx-js/mdx/create-element'
${rest}

export default ${fn}
`
```

Unless there is compelling reason for `mdx` to return the output broken into parts, I would go with 1 since it is a lot simpler to implement.

packages/mdx/test/index.test.js
johno: Yeah we really need to eventually nuke this flakey test and come up with better assertions rather than a snapshot test.

packages/mdx/test/index.test.js
johno: I like the `wrapExport` idea, especially since it doesn't require breaking the output into multiple parts (I think this could be tricky and unintuitive for users).

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -7,7 +7,8 @@ function toJSX(node, parentNode = {}, options = {}) {
   const {
     // Default options
     skipExport = false,
-    preserveNewlines = false
+    preserveNewlines = false,
+    wrapExport
```
pshrmn: The one thing that worries me about the option is that there are now two options (`skipExport` and `wrapExport`) that determine what gets returned. Should these be completely separate or is there a situation where you might wrap the component but not export it?
```js
// should this be possible? and what would the wrapped variable be named?
const ??? = React.memo(MDXContent)
```
This could be written to support that, but I'm not sure whether or not it should be supported.

johno: Hypothetically these options should be mutually exclusive, so I'm not _super_ worried about the overlap. `skipExport` is used pretty rarely and only for advanced use cases.

johno:
Thanks!

johno:
Published as `1.0.0-alpha.11`.

PR #477, opened 2019-03-20
Remove trailing space from prop-less JSX tags

pshrmn:
Purely aesthetic. The extra spaces in the tests were kind of bugging me.


johno:
Thanks!

PR #485, opened 2019-03-27
Pass filename to babel transform if present

jescalan:
I have been occasionally getting errors from babel about how it won't transform without a filename. This PR makes it such that if there is a filename present, it is passed to babel. In practice this has resolved my errors. I added some modifications to the tests to account for this change, but there is not a full integration test in place to replicate the scenario. I started down the road of adding one but it ended up being quite involved so I figured I'd start here.


johno:
Thanks!

PR #487, opened 2019-03-28
Rename @mdx-js/tag to @mdx-js/react

johno:


packages/react/readme.md
```diff
@@ -1,37 +1,42 @@
...
 ```
 
 ## Usage
 
 ```jsx
+/* @jsx mdx */
 import React from 'react'
 import {renderToString} from 'react-dom/server'
-import {MDXTag} from '@mdx/tag'
+import {MDXProvider, mdx} from '@mdx/react'
```
alexandernanberg: Should you be able to import `mdx` here? Doesn't look like it implemented if that's the case

johno:
Ah, whoops, that’s definitely a docs bug, nice catch!

PR #488, opened 2019-03-28
Add remove exports plugin

johno:


ChristopherBiscardi:
This plugin should remove the `export` keyword but not the `const` to mirror how it would work outside of the playground right?

johno:
Right now this purely strips them. I'll be adding to the functionality when I begin building out the playground.

PR #493, opened 2019-03-29
Fix missing "create-element.js" file

alexandernanberg:


johno:
Thank you!

johno:
Published as `1.0.0-alpha.16`.

PR #494, opened 2019-03-30
Move to babel v7

alexandernanberg:
Upgrades to babel v7 and jest v24

Also added the `transform-runtime` plugin which _should_ mean smaller bundle size for people using mdx


alexandernanberg:
Hmm not sure how to fix the doc site, tried adding babel-core bridge but that didn’t work either

packages/react/.babelrc
```diff
@@ -1,7 +1,4 @@
 {
-  "presets": [
-    "env",
-    "stage-0",
-    "react"
-  ]
+  "presets": ["@babel/preset-env", "@babel/preset-react"],
+  "plugins": ["@babel/plugin-transform-runtime"]
```
timneutkens: Why is this needed? I actually think MDX doesn't need `preset-env` even 🤔 It doesn't do anything special right 🤔

alexandernanberg: Do you mean why any babel transpilation is needed? I assumed it was because both JSX and object spread is used in the [context.js](https://github.com/mdx-js/mdx/blob/v1/packages/react/src/context.js) file 🤔

timneutkens: So it'd only be needed for https://github.com/mdx-js/mdx/blob/v1/packages/react/src/context.js#L19

`preset-env` doesn't compile JSX, preset-react does 👍

alexandernanberg: Yeah exactly, wasn't sure if you referred to both presets or just the env 😅 

I would guess it's needed for https://github.com/mdx-js/mdx/blob/0cca7119cc1b8980722939db2b85a521ecc56532/packages/react/src/context.js#L10 as well right?

If we really want to we could replace the object spread with `Object.assign` and JSX with `React.createElement` and remove the babel transpilation altogether.

timneutkens: Oh yeah you're right

alexandernanberg: Yeah, I'm fine with either so I'll leave it up to you to decide and I'll update the PR 😄 AFAIK the runtime has no value if we aren't using the preset-env and add `useBuiltIns` to preset-react.

 It would be fairly simple to fix that for IE users by simply having them add a polyfill.

What about the import/export statements, do they need to be transpiled?

timneutkens: I think we should use https://www.npmjs.com/package/microbundle to output multiple bundles 👍

alexandernanberg: Ah yeah, I've been wanting to have a look at microbundle! Did some initial testing with it at it looks promising. Just wondering how we'd solve the `create-element.js` file since that one isn't included in the `index.js` 🤔

pshrmn: Is there any particular reason that the `mdx` import has to come from the `create-element` module instead of having the `@mdx-js/react` package export it from the main module? Browsing through the source, I don't see any reason that it couldn't be imported from the main module.

johno:
> Hmm not sure how to fix the doc site, tried adding babel-core bridge but that didn’t work either

I'll be working on porting the docs site this weekend to something a bit more powerful/flexible so it shouldn't be an issue. Since this is in the v1 branch we can go ahead and merge with a broken docs site for the time being.

packages/react/.babelrc
johno: We should be able to export `create-element` from `index.js` as well 👍

alexandernanberg: Would we export it as mdx like it is done here then? https://github.com/mdx-js/mdx/blob/v1/packages/vue/src/index.js

pshrmn: @alexandernanberg I actually went ahead and made the change in #495

alexandernanberg:
I've added microbundle now! Not sure if we want an umd build or not and what file extensions to use for the es build 🤔

johno:
Thanks!

PR #495, opened 2019-03-30
Export mdx from @mdx-js/react's index module

pshrmn:
Currently, `mdx` is exported/imported from its own module in the `@mdx-js/react` package. This PR moves the export to the package's index module.

```js
// before
import mdx from "@mdx-js/react/create-element";
// after
import { mdx } from "@mdx-js/react";
```
I saw that the `@mdx-js/vue` package exports its `mdx` function from the package's index module, so this PR would make the `@mdx-js/react` package consistent with that. Of course if I have overlooked some reason for keeping the export separate, this PR can be ignored :smile:.


johno:
Thank you!

johno:
Published as `1.0.0-alpha.17`.

PR #497, opened 2019-04-02
fix code in guide syntax highlighting guide

SimonSiefke:
Without `{...props}` all the tags are invisible, because they have no content


johno:
Thank you!

PR #502, opened 2019-04-05
Dark mode

jxnblk:
😎 This adds a dark mode color scheme to the docs site. I'm not crazy about the toggle button/icon design, so more than happy to change that. The actually colors used can be adjusted as well. Unlike styled-system.com this is only a binary dark/light color setup

Screen Shot 2019-04-05 at 12 37 34 PM


johno:
Woooooot!

PR #504, opened 2019-04-05
Implement search

johno:


docs/src/components/layout.js
```diff
@@ -173,6 +176,62 @@ export default props => {
     window.localStorage.setItem('dark', dark)
   }, [dark])
 
+  // Copied from Gatsby (MIT License)
+  // https://github.com/gatsbyjs/gatsby/blob/30070ec9a9e678ab8f23adda952510938c09bfcb/www/src/components/search-form.js
+  useEffect(() => {
```
johno: We should prolly break this all out into a `use-search` hook that a new Search component uses.

PR #505, opened 2019-04-05
Adjust search dropdown styles for mobile and add skip link

jxnblk:
- Pulls search out into its own module
- Adds search icon
- Attempts to style search dropdown better for mobile (could be better)
- Bonus: adds skip link

### DIFF
--- a/docs/src/components/search.js
+++ b/docs/src/components/search.js
@@ -1,0 +1,146 @@
+import React, { useEffect, useRef } from 'react'
+import { navigate } from 'gatsby'
+import { css } from '@styled-system/css'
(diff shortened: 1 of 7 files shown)


docs/src/components/search.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+    next[key] = value + ' !important'
+  }
+  return next
+}
+
+const styles = theme => important(css({
+  display: 'flex',
+  alignItems: 'center',
+  // <3 CSS
```
johno: 💟

johno:
Wonderful, thank you!

Issue #454, opened 2019-03-07
Multiple MDXs in a Single File


ChristopherBiscardi:
# multiple MDXs in a single file

The wider ecosystem seems to have a need for multiple MDX components sourced from a single file. Consider the following examples.

### mdx-deck

[mdx-deck](https://github.com/jxnblk/mdx-deck) uses multiple MDX components from a single file to render a slide deck.

    # This is the title of my deck
    ---
    # About Me
    ---
    ```jsx
    <CodeSnippet />
    ```
    ---
    import Demo from './components/Demo'
    
    <Demo />
    ---
    # The end

![image](https://user-images.githubusercontent.com/551247/53987627-6862df80-40d6-11e9-819a-57d78f3f19e9.png)

### small content

There is a common use case in websites to have multiple blobs of smaller content. I've used some mock content from the Gatsbyjs.org site here to illustrate the point, but pretty much all product sites do this somewhere including 

<details><summary>Next</summary>

![image](https://user-images.githubusercontent.com/551247/53987158-0f467c00-40d5-11e9-9670-0c169d604389.png)

</details> 
<details><summary>React</summary> 

![image](https://user-images.githubusercontent.com/551247/53986821-0dc88400-40d4-11e9-896e-d091200f7dc2.png)

</details>
<details><summary>Terraform</summary> 

![image](https://user-images.githubusercontent.com/551247/53986799-fee1d180-40d3-11e9-97a9-8129dd4a63ed.png)

</details>

```markdown
# Modern web tech without the headache

Enjoy the power of the latest web technologies – React.js , Webpack , modern JavaScript and CSS and more — all set up and waiting for you to start building.
---
# Bring your own data

Gatsby’s rich data plugin ecosystem lets you build sites with the data you want — from one or many sources: Pull data from headless CMSs, SaaS services, APIs, databases, your file system, and more directly into your pages using GraphQL .
---
# Scale to the entire internet

Gatsby.js is Internet Scale. Forget complicated deploys with databases and servers and their expensive, time-consuming setup costs, maintenance, and scaling fears. Gatsby.js builds your site as “static” files which can be deployed easily on dozens of services.
```

![image](https://user-images.githubusercontent.com/551247/53987653-7d3f7300-40d6-11e9-90b1-45b6de0a8d56.png)

### "Storybook", Documentation sites, and other interesting use cases

[...]

PR #507, opened 2019-04-10
Implement first pass of MDXs

johno:
This implements the basic idea of MDXs. Though
we still need to figure out how to best handle
the layout mechanism.

cc/ @christopherbiscardi, @jxnblk, @timneutkens

---

Related #454

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -30,6 +30,7 @@ function createMdxAstCompiler(options) {
   const fn = unified()
     .use(toMDAST, options)
     .use(remarkMdx, options)
+    .use(mdxHastToJsx, options) // Set JSX compiler early so it can be overridden
     .use(squeeze, options)
     .use(toMDXAST, options)
 
@@ -80,8 +81,6 @@ function applyHastPluginsAndCompilers(compiler, options) {
       compiler.use(plugin, options)
     }
   })
-
-  compiler.use(mdxHastToJsx, options)
 
   for (const compilerPlugin of compilers) {
     compiler.use(compilerPlugin, options)
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -30,7 +30,6 @@ function createMdxAstCompiler(options) {
   const fn = unified()
     .use(toMDAST, options)
     .use(remarkMdx, options)
-    .use(mdxHastToJsx, options) // Set JSX compiler early so it can be overridden
     .use(squeeze, options)
     .use(toMDXAST, options)
 
@@ -81,6 +80,8 @@ function applyHastPluginsAndCompilers(compiler, options) {
       compiler.use(plugin, options)
     }
   })
+
+  compiler.use(mdxHastToJsx, options)
 
   for (const compilerPlugin of compilers) {
     compiler.use(compilerPlugin, options)
--- a/packages/remark-mdxs/index.js
+++ b/packages/remark-mdxs/index.js
@@ -1,0 +1,76 @@
+const visit = require('unist-util-visit')
+const remove = require('unist-util-remove')
+const {toJSX} = require('@mdx-js/mdx/mdx-hast-to-jsx')
+
+module.exports = function({delimiter = 'hr'}) {
+  this.Compiler = tree => {
+    const splits = []
+    const documents = []
+
+    const importNodes = tree.children.filter(n => n.type === 'import')
+    const exportNodes = tree.children.filter(n => n.type === 'export')
+
+    const layout = exportNodes.find(node => node.default)
+
+    // We don't care about imports and exports when handling
+    // multiple MDX documents
+    let mdxsTree = remove(remove(tree, 'export'), 'import')
+    const {children} = mdxsTree
+
+    visit(mdxsTree, node => {
+      if (node.tagName === delimiter) {
+        splits.push(children.indexOf(node))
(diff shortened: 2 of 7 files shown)


packages/remark-mdxs/index.js
```diff
@@ -0,0 +1,76 @@
...
+      '',
+      exportNodes
+        .filter(n => !n.default)
+        .map(n => n.value.trim())
+        .join('\n'),
+      '',
+      `const MDXSLayout = ${
+        layout
+          ? layout.value
+              .replace(/^export\s+default\s+/, '')
```
johno: We should prolly make this a shared util.

packages/remark-mdxs/index.js
```diff
@@ -0,0 +1,76 @@
...
+function MDXSContent${i}({ components, ...props }) {
+  return (
+    ${jsx.trim()}
+  )
+}
+    `.trim()
+      )
+
+    const defaultExport = `
+const MDXSWrapper = props => [
```
johno: I guess we could actually make this the `MDXLayout`?

packages/remark-mdxs/readme.md
```diff
@@ -0,0 +1,55 @@
...
+> :warning: This project is currently in alpha
+
+[MDXs][] syntax support for [remark][].
+
+## Installation
+
+```sh
+npm install --save remark-mdxs
+```
+
```
jxnblk: From a usage standpoint, is the thinking that you would use something like a separate webpack loader rule that only parses `.mdxs` files while handling `.mdx` files like any other?

packages/mdx/index.js
```diff
@@ -30,6 +30,7 @@ function createMdxAstCompiler(options) {
   const fn = unified()
     .use(toMDAST, options)
     .use(remarkMdx, options)
+    .use(mdxHastToJsx, options) // Set JSX compiler early so it can be overridden
```
johno: Looks like this is causing some issues in other tests. Wonder if we should instead use (and document) the [`compilers` API](https://github.com/mdx-js/mdx/blob/master/packages/mdx/index.js#L62)?

packages/remark-mdxs/index.js
```diff
@@ -0,0 +1,76 @@
+const visit = require('unist-util-visit')
+const remove = require('unist-util-remove')
+const {toJSX} = require('@mdx-js/mdx/mdx-hast-to-jsx')
+
+module.exports = function({delimiter = 'hr'}) {
```
jxnblk: Just curious, I'm using `thematicBreak` in `@mdx-deck/mdx-plugin` but with the way the delimiter is used below will either `hr` or `thematicBreak` work?

packages/remark-mdxs/readme.md
johno: Yeah that's the idea 👍

packages/remark-mdxs/index.js
johno: As is it won't work because we're operating on MDXHAST (since it's acting as a custom compiler for MDX). But it might make sense to make MDXS more standalone and not use `mdx` core and `remark-mdx` directly instead. That would allow us to operate on the MDXAST.

packages/remark-mdxs/readme.md
ChristopherBiscardi: That seems like it will turn out to be kind of unfortunate. It means we can't delineate between MDX content and MDXS content when fetching from remote sources that use `text/markdown` content types (mdx still doesn't have it's own entry [in the mime-db](https://github.com/jshttp/mime-db/pull/136) and even if it did, we're going to want compat with markdown types from headless CMSs anyway).

Since every MDXs document is technically a valid MDX document, I'd much rather see a const export used to specify intent to be MDXs content.

johno: Hmm, that's a good point 🤔. I'd mostly been thinking about local fs. How do you think intent for MDXs should be expressed?

Something like?:

```jsx
export const MDXS = true
```

ChristopherBiscardi: yeah this is the approach I landed on as well. I think that the "first MDX file" in an MDXs document can be used as the "global exports" (compared to exports defined in subsequent "files"), so you'd end up with something like this where the first MDX doesn't have any content, but is just imports/exports.

```
// global export
export const MDXS = true
---
// file-local export
export const slideLayout = SomeComponent
# some slide
---
# another slide
```

polarathene:
I'm not too familiar with the PR, but are you proposing to use `---` for indicating the split between separate MDX docs? Isn't that meant to be treated as an <hr>? 

Does a user have to swap any markdown usage there directly for <hr> markup to use this feature?

johno:
Going to close this for now, will likely reopen after MDX v2 ships.

packages/remark-mdxs/index.js
karlhorky: Regarding these architectural things, if `remark-mdxs` (or a different implementation of MDXs) were to be implemented today, do either of you (or @ChristianMurphy or @wooorm) have any recommendations for trying to make a new MDXs PR for MDX v2?

@EricCote or @ProchaLu (or both) may take a shot at implementing something like this in the next weeks, see this:

- https://github.com/0phoff/MDXP/issues/12#issuecomment-1359354136

PR #522, opened 2019-04-12
Use a Set for shortcode names

### LINKED ISSUE #518: Shortcodes breaks if same component if used more than once
### Subject of the issue

Shortcodes breaks if same component if used more than once. We probably want to add a check if the `makeShortcode` is already added or not.

### Your environment

`@mdx-js/mdx`@1.0.2
`@mdx-js/react`@1.0.2

### Steps to reproduce

```md

<Foo bg='red'>
  <Bar>hi</Bar>
  {hello}
  {/* another commment */}
</Foo>

<Foo bg='red'>
  <Bar>hi</Bar>
</Foo>

```

### Expected behaviour

Should work without errors

### Actual behaviour

```
SyntaxError: unknown: Identifier 'Foo' has already been declared (17:6)

      15 | const Foo = makeShortcode("Foo");
      16 | const Bar = makeShortcode("Bar");
    > 17 | const Foo = makeShortcode("Foo");
         |       ^
      18 | const Bar = makeShortcode("Bar");
      19 | const layoutProps = {
      20 |   foo
```

chasemccoy:
Fixes #518

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -174,7 +174,7 @@ MDXContent.isMDXComponent = true`
   return <div {...props}/>
 };
 ` +
-      jsxNames
+      new Set(jsxNames)
         .filter(name => !importExportNames.includes(name))
         .map(name => {
           return `const ${name} = makeShortcode("${name}");`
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Dope! LGTM

Just waiting for tests to pass

PR #529, opened 2019-04-16
Fix Firefox bug in remark-mdx

Jarred-Sumner:
### Firebox bug
Firefox doesn't support the `dotAll` flag in RegExp, which is used for the JSX tag . This means using `remark-mdx` in the browser breaks on Firefox right now. 

This should fix that, and I added an additional test for checking a JSX block with props. Truthfully, I haven't tried this on Firefox yet

### `skipImportsAndExports` to `remark-mdx`
 `skipImportsAndExports` in `remark-mdx` lets developers skip trying to parse `import` and `export`.

This means we can run `remark-mdx` in the browser without needing to import babel, reducing the bundle size. Theoretically, this could also be used in `node` if you don't want to support `import` or `export`.

Additionally, I made `@babel/core` and `@babel/standalone` an `optionalDependency` (and `devDependency`), so it doesn't install them by default unless you already have it/want it. If `remark-mdx` tries to parse imports/exports when babel is not available, it writes a message to `console.error` to inform the developer that they need to add `@babel/core` or `@babel/standalone` as a dependency (or pass `skipImportsAndExports` to `remark-mdx`). 

That way, if a developer does run into this issue, it at least tells them what they should do to fix it. 

This is probably a breaking change. 

### Babel build step for `remark-mdx`

This adds babel as a build step for `remark-mdx`. 
- There's a `build` and a `start` script in the `package.json`, matching the style used in other packages of this repo. 
- There's a `src` and `dist` folder, the `dist` is published on npm

In hindsight...this isn't strictly necessary? Initially, I tried to fix the Firefox bug via a babel transform, however the babel transform only supported regex literals. Instead, I manually did what the transform would've done, which is replace the `.` character with `\0-\uFFFF]/`

### Tests

I added tests for most of these changes. 

I also learned about the [testEnvironment](https://jestjs.io/docs/en/configuration#testenvironment-string) flag in Jest, so I have copied `extract-imports-and-exports` test to run once in the browser environment and once in the node environment, allowing us to make sure the babel standalone stuff is configured correctly.


Jarred-Sumner:
I manually verified that this fixes the Firefox bug

Jarred-Sumner:
Closing in favor of https://github.com/mdx-js/mdx/pull/532

PR #531, opened 2019-04-16
Only pass children to React.Fragment

### LINKED ISSUE #530: Upgrading to v1 seems to cause an issue when exporting arbitrary props
### Subject of the issue

Upgrading to v1 throws warnings on MDX documents that export arbitrary props, possibly because of the change from wrapping the document in a `div` to using a React Fragment?

### Steps to reproduce

I have an MDX document like this:

```mdx
export const author = "Foo"

# Heading
```

When I transform it I seem to get this warning:

```
Warning: Invalid prop `author` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props.
    in Fragment
    in MDXCreateElement
    in MDXContent
```

As far as I can tell, this is because all the props exported from the document are automatically passed to the “wrapper” component, which used to be a `<div>` and is now a `<React.Fragment>`.

The switch is fine and I think it makes sense, I’m just unclear if what is now happening is expected?

It seems like the default wrapper should be something like `({ children }) => <React.Fragment>{children}</React.Fragment>`?

johno:
Since the default wrapper is a React Fragment we
should ignore all props passed aside from children
so it doesn't cause React to warn.

---

Fixes #530


johno:
Published as 1.0.6

PR #532, opened 2019-04-16
Add dotAll polyfill to fix Firefox bug

Jarred-Sumner:
Firefox doesn't support the [dotAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) flag (`s`) in Regexp, which is used for the JSX tag . This means using `remark-mdx` in the browser breaks on Firefox right now when parsing MDX.

image

I manually verified this fixes the bug in Firefox

### DIFF
--- a/packages/remark-mdx/tag.js
+++ b/packages/remark-mdx/tag.js
@@ -3,11 +3,15 @@
 //
 // MIT License https://github.com/remarkjs/remark/blob/master/license
 
+// https://github.com/DmitrySoshnikov/babel-plugin-transform-modern-regexp#dotall-s-flag
+// Firefox and other browsers don't support the dotAll ("s") flag, but it can be polyfilled via this:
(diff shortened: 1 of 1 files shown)


wooorm:
Why not use `[\s\S]`? I believe we used to do that before the fancy dot-all flag 😅

Jarred-Sumner:
> Why not use [\s\S]?

I don't have a strong opinion. Prettier seems to get rid of the `\`'s though when I save `'[\s\S]'`, since it doesn't know it's a regex.

wooorm:
Ah yeah, in a string you'd need to double escape them, so it would be cleaner in this case to just replace the dot inside the regex with these. Sorry for the confusion, I mean that *that's* what we used to do!

johno:
Going to go ahead and merge then, thanks all!

PR #534, opened 2019-04-16
Ensure class property is passed as className

johno:
The hast-util-raw package returns a class prop which
needs to be renamed to className.


ChristopherBiscardi:
Does it make sense to do this processing here with the other aria stuff? https://github.com/mdx-js/mdx/blob/aad59f60092dcdb5c562e4a55b821dff5abd0131/packages/mdx/mdx-hast-to-jsx.js#L74-L91

johno:
I think eventually we'll want it to be colocated, but ideally that type of processing will happen as a plugin _before_ JSX compilation. So we'll leave as a cleanup item for the future ❤️

PR #547, opened 2019-04-22
apparently sometimes node.properties can be undefined

ChristopherBiscardi:


ChristopherBiscardi:
@wooorm thanks for the heads up. Might be a gatsby-remark-* plugin doing something weird.

PR #576, opened 2019-05-13
Stop forwarding options to remark/rehype plugins

### LINKED ISSUE #575: Forwarding options to remarkPlugins/rehypePlugins can break them
### Subject of the issue

Currently `@mdx-js/mdx` forwards its options to every plugin in `remarkPlugins` and `rehypePlugins`, which can easily lead to some of them breaking. Is there a specific reason why we're doing this? If it's a behavior left over from something else, I can remove it.

### Your environment

*   Name and version of operating system: macOS 10.14.4
*   Names and version of required packages: 1.0.18
*   Version of node, npm, yarn, or names and versions of browser: yarn 1.16.0

### Steps to reproduce

```js
import mdx from '@mdx-js/mdx'
import detectFrontmatter from 'remark-frontmatter'

const content = `
---
title: Hello World!
---
`

mdx(content, {
  remarkPlugins: [
    detectFrontmatter,
  ],
})
```

### Expected behaviour

remark-frontmatter shouldn't be getting unexpected options.

### Actual behaviour

Options from the `mdx()` call get forwarded to remark-frontmatter, which thinks we're passing the definition for detecting frontmatter, but the shape of the object doesn't satisfy its criteria so it crashes.

### Temporary workaround

Passing `undefined` as an option to all plugins, which results in default, i.e. expected behavior.

```js
mdx(content, {
  remarkPlugins: [
    [detectFrontmatter, undefined],
  ],
})
```

silvenon:
This no longer serves its initial purpose, and is definitely unexpected behavior that can break plugins which expect certain properties or, worse, if a MDX option happen to match an option of that plugin.

This change shouldn't introduce problems because it's hard to imagine why someone would rely on this undocumented and buggy behavior of this advanced feature instead of filing an issue in this very actively maintained project. Also, if someone was getting around this bug by always forcing custom options, that workaround will continue to work after this change.

Btw, I'm using `toMatchInlineSnapshot` instead of something like `toBeUndefined` just so it's slightly easier to maintain if defaults change to something else, like an empty object instead of `undefined`.

Fixes #575.

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -38,7 +38,7 @@ function createMdxAstCompiler(options) {
     if (Array.isArray(plugin) && plugin.length > 1) {
       fn.use(plugin[0], plugin[1])
     } else {
-      fn.use(plugin, options)
+      fn.use(plugin)
     }
   })
 
@@ -78,7 +78,7 @@ function applyHastPluginsAndCompilers(compiler, options) {
     if (Array.isArray(plugin) && plugin.length > 1) {
       compiler.use(plugin[0], plugin[1])
     } else {
-      compiler.use(plugin, options)
+      compiler.use(plugin)
     }
   })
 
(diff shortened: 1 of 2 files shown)


johno:
Thanks @silvenon!

PR #594, opened 2019-06-04
chore: yarn upgrade

ChristianMurphy:
updating dependencies with versions that are in range, makes `yarn audit` happy.


ChristianMurphy:
This also allows `yarn` to install `iltorb` dependency on Node 12.

jxnblk:
Looks like this might've broken emotion on the docs site – I'm looking into it https://mdxjs.com/

ChristianMurphy:
Good catch @jxnblk.
Looks like one of the nested updates is messing with the loader or the theme.
I can revert, if need be.

jxnblk:
I'm not entirely sure why it broke, but the Now deploy in #596 looks like it works?? I've also opened #597 to update some of the dependencies used in the docs site

PR #527, opened 2019-04-14
Issue 449: Migrate travis to circleci

leimonio:
Migrating CI/CD configuration from TravisCI to CircleCI

Some further steps needed to enable CircleCI: (cc @johno )
- [x] [Enable CircleCI](https://circleci.com/docs/2.0/getting-started/#setting-up-your-build-on-circleci) build for mdx repository 
- [ ] Add `NPM_AUTH_TOKEN` as [environment variable on CircleCI](https://circleci.com/gh/leimonio/mdx/edit#env-vars)

Closes #449

Issue #449, opened 2019-03-07
Move to CircleCI


johno:

johno:
Would you be interested in taking this on still @leimonio?

leimonio:
Yeap @johno thanks for mentioning. 👌

johno:
Hm, going to reopen for now. Looks like the ci release isn't working as expected: https://circleci.com/gh/mdx-js/mdx/6

leimonio:
@johno can you try once more? https://github.com/mdx-js/mdx/pull/538

johno:
Looks like it's almost there now (tests are being run on PRs again thank you!). Looks like the npmrc addition is resulting in an uncommitted change so the publish won't work.

I think we could prolly remove the `.npmrc` and gitignore it since it isn't being used anyway. I'll give that a shot shortly.

johno:
Seems to have worked. Thanks again @leimonio! 💟

leimonio:
Great stuff!! Let me know if I can help anything else, happy to further contribute! :)

PR #595, opened 2019-06-04
ci: setup travis ci to automatically run test suite for each PR & commit

ChristianMurphy:
This will automatically run `yarn test` for each incoming commit and PR, and report the status back to GitHub.

https://travis-ci.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: node_js
+node_js:
+  - 8
+  - 10
+  - 12
+cache: yarn


.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+node_js:
+  - 8
+  - 10
+  - 12
```
ChristianMurphy: `iltorb` may fail during install phase

ChristianMurphy: :notebook: this will be resolved by #594

ChristianMurphy:
It appears Circle CI may also be setup, but is not reporting status.
ref #449 #527 #538 

What was the issue encountered that caused the migration away from Travis?
Any objections to bringing it back?

wooorm:
I don't remember the issue, but there was something going on 🤷‍♂️ If it's green now, I'm personally up for bringing it back!

johno:
Hmmm, the switch to CircleCI was intentional but I have no idea why it [stopped reporting 🤔 ](https://circleci.com/gh/mdx-js/mdx).

ChristianMurphy:
digging into the circle CI config it looks like https://github.com/mdx-js/mdx/blob/deff36bebfedb3a9de0a0575ee9a1b55b9b8aa18/.circleci/config.yml#L14-L17 is limiting it to only running on `master` branch.

ChristianMurphy:
> the switch to CircleCI was intentional

Okay, I'll repeat my question from before:

What was the issue encountered that caused the migration away from Travis?
Any objections to bringing it back?

johno:
> What was the issue encountered that caused the migration away from Travis?

I was mostly worried about the Travis acquisition and downtime. Funnily enough, we subsequently saw lots of CircleCI downtime after moving.

> Any objections to bringing it back?

Nope. To me CI is CI so whatever works!

PR #596, opened 2019-06-04
Update docs dependencies

jxnblk:
This is intended to fix the styles on mdxjs.com by updating the dependencies. I *think* there is a mismatched emotion version causing issues. Opening a draft PR to test with the Now deploy


ChristianMurphy:
Thanks @jxnblk! :bow:

PR #600, opened 2019-06-06
chore: update docs links to unify more gatsby mdx docs on .org

gillkyle:
As per discussion in https://github.com/gatsbyjs/gatsby/issues/14258, this is a starting point for unifying more of the `gatsby-mdx` docs on Gatsbyjs.org by linking directly to existing docs there rather than the repos.


johno:
Thanks!

PR #604, opened 2019-06-17
Improve block parsing for alphanumeric component names

### LINKED ISSUE #593: Headers tags wraps with paragraph
### Subject of the issue

h1 and H1 tags wraps with unnecessary paragraph tag 

### Steps to reproduce

you can reproduce this bug in [playground](https://mdxjs.com/playground)

Снимок экрана 2019-06-04 в 10 35 07

### Expected behaviour

h1 and H1 tags renders without wrapping p

### Actual behaviour

every h1 and H1 tag wraps with paragraph tag

johno:
Component names and HTML can be alphanumeric. This
updates the block parsing regex to reflect that.

Closes #593

### DIFF
--- a/packages/remark-mdx/block.js
+++ b/packages/remark-mdx/block.js
@@ -26,7 +26,7 @@ const elementCloseExpression = /^$/
 const otherElementOpenExpression = new RegExp(openCloseTag.source + '\\s*$')
 
 function blockHtml(eat, value, silent) {
-  const blocks = '[a-z\\.]+(\\.){0,1}[a-z\\.]'
+  const blocks = '[a-z0-9\\.]+(\\.){0,1}[a-z0-9\\.]'
   const elementOpenExpression = new RegExp(
     '^</?(' + blocks + ')(?=(\\s|/?>|$))',
     'i'
--- a/packages/remark-mdx/block.js
+++ b/packages/remark-mdx/block.js
@@ -24,13 +24,15 @@ const cdataOpenExpression = /^<!\[CDATA\[/
 const cdataCloseExpression = /\]\]>/
 const elementCloseExpression = /^$/
 const otherElementOpenExpression = new RegExp(openCloseTag.source + '\\s*$')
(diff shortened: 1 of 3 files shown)


chasemccoy:
@johno For what it's worth, I was seeing the issue in #593 without alphanumeric tags. 

My code is essentially like this:

```
<Paragraph bg='red.500' color='white'>Foo...</Paragraph>
```

And that was rendering as two `p` tags, one inside the other.

packages/remark-mdx/block.js
```diff
@@ -26,7 +26,7 @@ const elementCloseExpression = /^$/
 const otherElementOpenExpression = new RegExp(openCloseTag.source + '\\s*$')
 
 function blockHtml(eat, value, silent) {
-  const blocks = '[a-z\\.]+(\\.){0,1}[a-z\\.]'
+  const blocks = '[a-z0-9\\.]+(\\.){0,1}[a-z0-9\\.]'
```
ChristianMurphy: Would this regex allow for pure numeric blocks like:
```html
<00/>
<0.0/>
```

?

Is that intended?

wooorm:
Does this solve the issue? 🤔 `h1` is supported in remark as HTML, so maybe it’s something else?

Could a failing/fixed test help?

Note there’s two elements:

https://github.com/mdx-js/mdx/blob/4599f979cefe06d602f5150bf6dc4d746cb3b1b4/packages/remark-mdx/block.js#L50-L51

packages/remark-mdx/block.js
johno: Good catch, I've gone ahead and made the regex more thorough and added more test cases.

johno:
> Does this solve the issue? 🤔 h1 is supported in remark as HTML, so maybe it’s something else?

MDX is overriding the remark HTML handling so that we can handle JSX syntax which is more flexible. It now _should_ solve the issue with my latest changes this morning.

> Could a failing/fixed test help?

Added a bunch more test cases 💟

PR #609, opened 2019-06-18
Fix(gatsby-theme-mdx): move react & react-dom to peerDependencies (#608)

### LINKED ISSUE #608: gatsby-theme-mdx has react & react-dom in package.json dependencies
### Subject of the issue

`gatsby-theme-mdx`'s `package.json` has `react` & `react-dom` in `dependencies`. It makes all kind of unwanted behavior in monorepo setups. For example:

If you have `react` & `react-dom` installed in the parent directory (`./`) and `gatsby-theme-mdx` installed in a child directory (`./site`) directory, it creates a nasty bug. If you use hooks in your application, this error pops up:

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

And it is specifically caused by the point number 1 described in the error message. If you remove the parent directory's `node_modules` (i.e. `./node_modules`) everything works fine.

### Expected behaviour

`react` & `react-dom` should not be in the `dependencies`. They should be in the `peerDependencies`.

### Actual behaviour

`react` & `react-dom` are in the `dependencies`. So, it is installed automatically and causes error related to version mismatch.

MunifTanjim:
Fixes #608

### DIFF
--- a/packages/gatsby-theme-mdx/package.json
+++ b/packages/gatsby-theme-mdx/package.json
@@ -32,8 +32,6 @@
     "is-absolute-url": "^3.0.0",
     "lodash.flatten": "^4.4.0",
     "prism-react-renderer": "^0.1.6",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
     "react-helmet": "^5.2.0",
     "react-live": "^2.0.1",
     "remark-emoji": "^2.0.2",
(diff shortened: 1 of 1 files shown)


johno:
Thanks!

PR #614, opened 2019-06-23
Updated gatsby.md to replace @next tag with @latest

jhalvorson:
This is a tiny change to update the `@next` tag in the Gatsby docs to `@latest`. If you install the current `@next` which is `1.0.0-rc.4` and if `gatsby-remark-prismjs` is also installed then the project fails to build.


johno:
Thanks!

PR #624, opened 2019-07-10
Remove large (+20kb) 'change-case' dependency

### LINKED ISSUE #623: The 'change-case' package is bloating bundle size
The [`change-case`](https://www.npmjs.com/package/change-case) dependency is `>20kb` which is pushing us over our bundle size budget. Can it be replaced/inlined?

wilsonpage:
Fixes #623

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -2,11 +2,13 @@ const {transformSync} = require('@babel/core')
 const declare = require('@babel/helper-plugin-utils').declare
 const {types: t} = require('@babel/core')
 const toStyleObject = require('to-style').object
-const {paramCase} = require('change-case')
 const uniq = require('lodash.uniq')
 const {toTemplateLiteral} = require('./util')
 
 const STARTS_WITH_CAPITAL_LETTER = /^[A-Z]/
+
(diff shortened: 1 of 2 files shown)


ChristianMurphy:
@wilsonpage would focusing the npm package to https://github.com/blakeembrey/param-case reduce the bundle size, and keep the edge case handling?

johno:
@ChristianMurphy at this point we _shouldn't_ need any edge case handling since we're only [transforming two known cases](https://github.com/mdx-js/mdx/blob/3686c8dee70addecfb9730726f42e6ce20a5ce35/packages/mdx/mdx-hast-to-jsx.js#L88-L98).

timneutkens:
Updated the pr description to auto-close the issue.

wilsonpage:
@ChristianMurphy `param-case` still depends on the large dependency `no-case`.

wilsonpage:
Updated the PR to cope with the `data123` case.

wilsonpage:
@ChristianMurphy I'm afraid I don't have enough context on this project to write a test. Would take me a while to grok, which I can't afford right now :( Would be great if one of you lot could add one!

johno:
I'll add a few tests in a follow up PR, thanks for the contribution @wilsonpage!

PR #642, opened 2019-07-15
Trim code

karlhorky:
Thanks for this project and example! Really helpful for creating documents with interactive parts.

Looks like the following image without this change, there is an extra empty line at the bottom of the code fence when used like this:

<pre>
```html
&lt;div id="1">&lt;/div>
```
</pre>

Screen Shot 2019-07-15 at 16 27 34

## When inspected:

Screen Shot 2019-07-15 at 16 21 26


johno:
Thanks!

karlhorky:
No worries, thanks for the merge!

Issue #685, opened 2019-07-25
MDXProvider doesn't do anything in a new project made with npm init mdx


queerviolet:
### Subject of the issue

In a new project just created with `mdx init`, changing the `components` prop for `MDXProvider` doesn't do anything.

After some digging, it looks like `MDXProvider` is providing a different context than is being consumed by `MDXCreateElement`

### Your environment

*   **OS**:  macOS Mojave
*   **Packages**: Whatever's `npm init mdx` does, which is this:

```
    "@mdx-js/loader": "^1.1.0",
    "@mdx-js/mdx": "^1.1.0",
    "@next/mdx": "^9.0.1",
    "next": "^9.0.1",
    "react": "^16.8.6",
    "react-dom": "^16.8.6",
    "remark-emoji": "^2.0.2",
    "remark-images": "^1.0.0",
    "unified-ui": "^0.0.3"
```

*   **Env**:
  - node v12.3.1
  - npm 6.9.0
  - yarn 1.17.3

### Steps to reproduce

Minimal repro repo: https://github.com/queerviolet/mdx-context-repro

1. `npm init mdx`
2. Edit the generated `pages/_app.js` file like so:

```jsx
import React from 'react'
import {MDXProvider} from '@mdx-js/tag'
import {Provider} from 'unified-ui'

export default ({Component, pageProps}) => (
  <MDXProvider components={{ p() { return 'Hello?' } }}>
    <Provider>
      <Component {...pageProps} />
    </Provider>
  </MDXProvider>
)
```

### Expected behaviour

Well gosh, I'd expect to see a page full of 'Hello?'s.

### Actual behaviour

There's no change whatsoever in the rendered output.

### Why?

It looks like a bundling problem.

Issue one is that there's another `MDXProvider` inserting itself in the middle:

![Screen Shot 2019-07-25 at 4 04 06 PM](https://user-images.githubusercontent.com/7763406/61905060-25bf6980-aef6-11e9-8735-ce913d0e82ec.png)

Ripping out the `unified-ui` `Provider` gets rid of the inner one, which I do not control.

![Screen Shot 2019-07-25 at 4 11 32 PM](https://user-images.githubusercontent.com/7763406/61905396-fceba400-aef6-11e9-9139-1036ac7aacee.png)

[...]

PR #687, opened 2019-07-27
import mdxProvider from @mdx-js/react

shedali:
fixes parcel example not permitting overriding html elements

fixed as per comment at  https://github.com/mdx-js/mdx/issues/685#issuecomment-515203337


johno:
Thank you!

PR #694, opened 2019-07-30
fix(deps): react should be a peer dependency

zkochan:
React is used in @mdx-js/runtime, so it should be declared as a
peer dependency.


johno:
Thank you!

Issue #698, opened 2019-07-31
Exports should be transformed to variables in the playground/runtime


johno:
I've [previously written code to do this](https://github.com/gatsbyjs/gatsby/blob/57390e8da2d89c6abd0ad235c7389f39d072a36f/packages/gatsby-plugin-mdx/utils/babel-plugin-remove-export-keywords/index.js) so we should port that code into a package in this monorepo and name it `babel-plugin-remove-export-keywords`.

PR #699, opened 2019-07-31
feat: Add `babel-plugin-remove-export-keywords`

johnletey:
@johno Here you go!

Closes GH-698

### DIFF
--- a/packages/babel-plugin-remove-export-keywords/index.js
+++ b/packages/babel-plugin-remove-export-keywords/index.js
@@ -1,0 +1,14 @@
+module.exports = function removeExportKeywords() {
+    return {
+      visitor: {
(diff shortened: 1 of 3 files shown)


johno:
Thanks a ton! 😻 😻😻

Do you mind maybe bringing over the test and setting up the `package.json` to mirror the others in this project? 🙏

johnletey:
Sorry that took so long @johno ... but I have added the test (which doesn't pass because new-lines are added when it formats, but I didn't want to change anything before running it by you) and also have formatted the `package.json` like the rest of the packages

johno:
Amazing! Feel free to change the fixture so that it passes 👍

johnletey:
Will do!

PR #715, opened 2019-08-06
Stringify another array property

karlhorky:
Building on @devongovett's #158, this adds the `sandbox` property to the list of properties that should allow spaces.

Without this fix, using `gatsby-remark-embedded-codesandbox` along with `gatsby-plugin-mdx` results in an `iframe` with a `sandbox` attribute of `allow-modals,allow-forms,allow-popups,allow-scripts,allow-same-origin`, even though `gatsby-remark-embedded-codesandbox` is correctly returning space-separated strings.

If this is truly the correct solution, then probably we should include other HTML properties that accept spaces.

If there's a better / more correct way of doing this, I'm all ears! The remark plugin `gatsby-remark-embedded-codesandbox` does appear to be doing everything right however, returning a `node.value` with a `sandbox` attribute separated by spaces. It seems like MDX itself is changing the spaces to commas...

Fixes https://github.com/elboman/gatsby-remark-embedded-codesandbox/issues/7

cc @timneutkens @elboman

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -232,13 +232,19 @@ export default ${fnPostMdxTypeProp}`
 
   if (node.type === 'element') {
     let props = ''
+    
+    const shouldBeStrings = ['className', 'sandbox']
 
-    if (node.properties && Array.isArray(node.properties.className)) {
-      node.properties.className = node.properties.className.join(' ')
-    }
+    if (node.properties) {
+      shouldBeStrings.forEach(prop => {
+        if (Array.isArray(node.properties[prop])) {
+          node.properties[prop] = node.properties[prop].join(' ')
+        }
+      })
 
-    if (node.properties && Object.keys(node.properties).length > 0) {
-      props = JSON.stringify(node.properties)
+      if (Object.keys(node.properties).length > 0) {
+        props = JSON.stringify(node.properties)
+      }
     }
 
     return `<${node.tagName}${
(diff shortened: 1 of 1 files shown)


wooorm:
@johno Is there a reason https://github.com/syntax-tree/hast-to-hyperscript isn’t used? It does exactly all this for both React and Vue!

timneutkens:
I remember it does too much conversion. We wanted the available properties for users to override to be MDAST elements not HAST elements

wooorm:
Ahh, right, this is MDXHAST, not hast 🤔

We could move the [attribute handling](https://github.com/syntax-tree/hast-to-hyperscript/blob/d2a1b3d5336beca8ee6b68e8cefe46efeac71e93/index.js#L148) in `hast-to-hyperscript` to its own file, but I think you’ll then pull in more  weight that needed.

Alternatively, I’d suggest to depend on `property-information` and inline the code in https://github.com/syntax-tree/hast-to-hyperscript/blob/d2a1b3d5336beca8ee6b68e8cefe46efeac71e93/index.js#L148-L198 (and remove all the unneeded non-react things)

johno:
Yeah, I think we'd need a handful of changes so it'd work the way we want for MDX. That said, at the very least, we should break out a lot of this shared code into a package/file these libraries could share. Though we should probably make that it's own issue.

I'm cool merging this as is since it fixes something that's indeed broken while we investigate how we can better implement the MDXHAST to JSX conversion.

wooorm:
Yup, agreed!

johno:
Cools, will do. Thanks for the PR @karlhorky!

karlhorky:
No worries, thanks for the merge / release!

PR #725, opened 2019-08-09
Loader renderer option

axdg:
This fixes #711 

- adds the `prepend` to the Webpack loader's `options`.
- it removes the inline pragma `/* @jsx mdx */` - this is not needed since [mdx itself adds this during compilation](https://github.com/mdx-js/mdx/blob/master/packages/mdx/index.js#L112).
- add to README to reflect changes.

All changes are completely backwards compatible... this would just allow for other frameworks to be used with the loader.

### DIFF
--- a/packages/loader/index.js
+++ b/packages/loader/index.js
@@ -1,11 +1,19 @@
 const {getOptions} = require('loader-utils')
 const mdx = require('@mdx-js/mdx')
 
-module.exports = async function(content) {
+const DEFAULT_PREPEND = `
+  import React from 'react'
+  import { mdx } from '@mdx-js/react'
+`
+  .split('\n')
+  .map(l => l.trim())
+
+const loader = async function(content) {
   const callback = this.async()
   const options = Object.assign({}, getOptions(this), {
     filepath: this.resourcePath
   })
+
   let result
 
   try {
@@ -14,11 +22,10 @@ module.exports = async function(content) {
     return callback(err)
   }
 
-  const code = `/* @jsx mdx */
-  import React from 'react'
-  import { mdx } from '@mdx-js/react'
-  ${result}
-  `
+  const {prepend = DEFAULT_PREPEND} = options
 
+  const code = `${prepend}\n${result}`
   return callback(null, code)
 }
+
+module.exports = loader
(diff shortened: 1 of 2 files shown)


packages/loader/index.js
```diff
@@ -1,11 +1,19 @@
 const {getOptions} = require('loader-utils')
 const mdx = require('@mdx-js/mdx')
 
-module.exports = async function(content) {
+const DEFAULT_PREPEND = `
+  import React from 'react'
+  import { mdx } from '@mdx-js/react'
+`
+  .split('\n')
```
johno: In order to avoid this logic maybe we can dedent the template string?

```js
const DEFAULT_PREPEND = `
import React from 'react'
import { mdx } from '@mdx-js/react'
`
```

packages/loader/readme.md
```diff
@@ -32,6 +32,53 @@ module: {
...
 ```
 
+The `prepend` option specifies a string that will be prepended to the generated source allowing for the use of any `createElement` implementation. By default, that string is:
+
+```js
+import React from 'react'
+import { mdx } from '@mdx-js/react'
+```
+
+One could use any other implementation. The example below wraps a generic JSX compatible function named `h`.
```
johno:
```suggestion
Using the `prepend` option, one can swap out React for another implementation.  The example below wraps a generic JSX compatible function named `h`.
```

axdg: Yep, sounds good.

axdg:
@johno, I've made the suggested changes.

I agree that `prepend` is not a great name for the options... `renderer` is better, but go with whatever you think works.

axdg:
P.S. I don't like `pragma`, since this is a little ambiguous (as is the way that the word is used to refer to custom `createElement` wrappers in the documentation / blog). 

From wikipedia:

>In computer programming, a directive or pragma (from "pragmatic") is a language construct that specifies how a compiler (or other translator) should process its input.

The pragma in this case would be the directive `/* @jsx mdx */`, not the actual import / implementation.

johno:
Yeah, I _think_ I'm liking `renderer` the most. Let's go with that so it's slightly more specific/descriptive. 🙏

axdg:
Updated that last commit and the docs (feel free to squash anyway) - `renderer` seems pretty intuitive to me.

johno:
Thanks!

wooorm:
@johno ah sorry I missed this ping! Render is good!

PR #734, opened 2019-08-16
Use babel standalone in runtime, change exports to variable declarations

johno:
This allows us to add in additional babel plugins including a default which transforms exports into variable declarations for usage in playgrounds and other browser environments.


johno:
This is merged into the v2 branch.

PR #745, opened 2019-08-22
Break out attribute coercion to its own package

johno:
Translate HTML attributes to React-friendly attributes. This doesn't take care of all cases yet, nor is it well tests. More thorough tests will happen as part of the integration tests when it's wired into `@mdx-js/mdx`.


packages/babel-plugin-html-attributes-to-jsx/readme.md
```diff
@@ -0,0 +1,62 @@
...
+  </div>
+)
+```
+
+### Output
+
+```js
+const Foo = () => (
+  <div srcSet="foo">
+    <Button mdxType="Button" />
```
ChristopherBiscardi: is `mdxType` going to be part of this plugin or should it be taken out of the docs?

Issue #585, opened 2019-05-25
MDXCreateElement breaks forwardRef()


atomiks:
### Subject of the issue

I'm not sure if this is `gatsby-mdx` or this repo, so forgive me here. I upgraded from v0 to v1 and am experiencing this problem. v0 seemed to work okay.

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

### Your environment

*   macOS 10.14.4
*   
```
    "@mdx-js/mdx": "^1.0.19",
    "@mdx-js/react": "^1.0.16",
    "gatsby-mdx": "^0.6.3",
```

*   Node 10, npm 6.8

### Steps to reproduce

I don't actually know if this reproduces (haven't tested sorry 😥 ), but this is the general idea so hopefully it gives pointers:

Framework.js:

```jsx
import {useRef, cloneElement} from 'react';
import styled from 'styled-components';

export function Parent({ children }) {
  const ref = useRef();

  return cloneElement(children, {
    ref(node) {
      ref.current = node;
    }
  });
}

export const Child = styled.button``;
```

index.mdx:

```jsx
import {Parent, Child} from '../components/Framework';

<Parent>
  <Child />
</Parent>
```

I think the `<Child />` is no longer wrapped in `forwardRef`, causing the error message

### Expected behaviour

MDXCreateElement should keep `forwardRef` return value so the element can be given a ref

### Actual behaviour

It doesn't?

atomiks:
I got some time to verify using [this](https://gatsby-mdx.netlify.com/getting-started) starter kit. The same error occurs. 

Here it is simplified as much as possible, without `styled-components`:

```jsx
import React, { useRef, cloneElement, forwardRef } from 'react';

export function Parent({ children }) {
  const ref = useRef();
  return cloneElement(children, { ref });
}

export const Child = forwardRef((props, ref) => (
  <button ref={ref} />
));
```

```jsx
import { Parent, Child } from '../components/test';

<Parent>
  <Child />
</Parent>
```

This is a blocker for me upgrading to v1 for my documentation website, which I really want to do, but can't since the core component uses `forwardRef` everywhere and accesses its ref like that.

peduarte:
Can validate 👍

atomiks:
I think I found a fix with node_modules debugging.

Wrap `MDXCreateElement` itself in `forwardRef` like this:

```js
const MDXCreateElement = forwardRef((props, ref) => {
[...]

PR #751, opened 2019-08-27
Fix MDXCreateElement breaking forwardRef() 

Taym95:
…ref (#585)
I used the [proposed solution](https://github.com/mdx-js/mdx/issues/585#issuecomment-521906656) to fix it.


johno:
Thanks @Taym95!

PR #752, opened 2019-08-27
[MDXCreateElement] forward ref

Ayc0:
Enable setting `ref={...}` in MDX


johno:
Thanks for opening up a PR!!! I'm going to close this since I just merged #751 that solves the same issue.

PR #758, opened 2019-08-30
Fix and improve suggested types

rosskevin:
The previous type suggestions were broken and did not work, further their was no restriction on elements provided.  Used https://mdxjs.com/getting-started#table-of-components to restrict `components` and exported types for reuse/composition.

This could be the basis for in-repo types, but I suggest not accepting types in-repo unless it is accompanied by tests to exercise usage (and I don't have time to add, sorry).


johno:
Thank you!

PR #763, opened 2019-09-03
Don't parse ES syntax in open blocks

johno:
When a block is open, such as a list or blockquote,
we shouldn't attempt to parse import/export syntax.

Additionally, if a paragraph begins with import/export
and doesn't parse, we now make it a raw paragraph
node.

Related gatsbyjs/gatsby#17315

### DIFF
--- a/packages/mdx/test/fixtures/imports.mdx
+++ b/packages/mdx/test/fixtures/imports.mdx
@@ -1,0 +1,4 @@
+Hello, world!
+
+- import is a word
+- export is a word in lists, too!
--- a/packages/mdx/test/fixtures/imports.mdx
+++ b/packages/mdx/test/fixtures/imports.mdx
@@ -1,4 +1,6 @@
 Hello, world!
 
+import is a word
+
 - import is a word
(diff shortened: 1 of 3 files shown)


JounQin:
This fix breaks a fixture of [eslint-mdx](https://github.com/rx-ts/eslint-mdx).

The error detail should be checked whether it is a syntax error.

Example:

```js
import A from 'x'
import A from 'y'
```

It should throw error instead of parsed as paragraph.

You can check https://travis-ci.com/rx-ts/eslint-mdx/builds/125786664#L355 for detail.

JounQin:
This fix breaks a fixture of [eslint-mdx](https://github.com/rx-ts/eslint-mdx).

The error detail should be checked whether it is a syntax error.

Example:

```js
import A from 'x'
import A from 'y'
```

It should throw error instead of parsed as paragraph.

You can check https://travis-ci.com/rx-ts/eslint-mdx/builds/125786664#L355 for detail.

johno:
Hmm, yeah I guess we should revert the error catching piece 🤔

PR #775, opened 2019-09-09
Adds @babel/core to the peer dependencies

arcanis:
`@babel/core` is used here:

https://github.com/mdx-js/mdx/blob/master/packages/babel-plugin-apply-mdx-type-props/index.js#L1

That said, I think the dependency could be removed if you were using something like that instead:

```js
const {types: t} = api;
```

Cf here for an example:

https://github.com/arcanis/babel-plugin-lazy-import/blob/master/index.js#L15


johno:
Thanks! Yeah it looks like referencing types from `api` is the best solution, but for now I'll merge this since it will also work. Thanks for the PR!

PR #778, opened 2019-09-10
[WIP] Use babel for JSX generation

### LINKED ISSUE #742: Remove `layoutProps` from transpiled output
Since these values are already available in scope since they're exports, we can avoid instantiating the object and passing to MDXLayout.

johno:
---

Closes #742


millette:
Meanwhile, I submitted #790

johno:
Due to performance concerns we're going to avoid using Babel directly for codegen. Related to #1152.

PR #803, opened 2019-10-06
Preserve metastring language from codeblocks

pomber:
If there is a class arg in the metastring:
````
```js class=foo
```
````
`className="language-js"` is being replaced with `className="foo"`.

This PR adds the class to the list instead of replacing it: `className="language-js foo"`


johno:
Thank you!

PR #811, opened 2019-10-11
feat: add typescript typings for @mdx-js/mdx, @mdx-js/react, @mdx-js/runtime, and @mdx-js/vue

ChristianMurphy:
This adds typings for `@mdx-js/mdx`, `@mdx-js/react`, `@mdx-js/runtime`, and `@mdx-js/vue`.

A few notes on the typings
* the typings target TypeScript 3.4+, because they depend on the Unified typings, which require at least 3.4
* Even through TypeScript is not officially supported, adding typings may break the setup for TypeScript adopters who hand rolled typings, out of an abundance of caution Unified has been releasing typings as major releases to communicate this. This may, or may not be an approach mdx wants to take
* These typings don't necessarily need to live in this code base, the could also be published to definitely typed instead, if preferred.

/cc @Rokt33r


johno:
Thanks for this! I think it makes sense for the types to live in the monorepo 👍. Since we're probably about a month out from a v2 release maybe that's when we can merge in the types to make sure we avoid possible breakage?

You mind maybe pointing this PR to the v2 branch and then we can merge it @ChristianMurphy (when ready)?

ChristianMurphy:
> You mind maybe pointing this PR to the v2 branch and then we can merge it @ChristianMurphy (when ready)?

Sure, re-targeted to v2 branch in the github ui.
Looks like the automatic rebase ran into some conflicts.
Will manually cherrypick, resolvse conflicts, and force push later today.

ChristianMurphy:
re-targeted and re-based off `v2` branch.

docs/advanced/typescript.mdx
```diff
@@ -1,67 +1,37 @@
...
-    | 'inlineCode'
-    | 'hr'
-    | 'a'
-    | 'img'
+## Typings for Components and Utilities
 
-  export type Components = {
-    [key in ComponentType]?: React.ComponentType<{ children: React.ReactNode }>
-  }
+* \\\\\\\\\\`@mdx-js/mdx`
```
ChristianMurphy: The backslashes were added by prettier :man_shrugging:

packages/runtime/types/index.d.ts
```diff
@@ -0,0 +1,29 @@
+// TypeScript Version: 3.5
```
ChristianMurphy: This is a minor release higher than the others `Omit` was an addition in TypeScript 3.5

johno:
Going to go ahead and merge into v2. We'l prolly want to verify we don't need any changes after all v2 updates are made, but there shouldn't actually be any breaking changes for users so it probably won't be an issue.

Thanks!

PR #812, opened 2019-10-12
ci: update setup node action, add cross platform testing

ChristianMurphy:
resolves
```
Input 'version' has been deprecated with message: The version property will not be supported after October 1, 2019. Use node-version instead
```

warning

---

adds cross platform CI to github actions workflow.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+*.js   text eol=lf
+*.json text eol=lf
+*.jsx  text eol=lf
+*.md   text eol=lf
+*.mdx  text eol=lf
+*.ts   text eol=lf
+*.tsx  text eol=lf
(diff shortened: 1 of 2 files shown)


ChristianMurphy:
> What's the gitattributes for?

The lint task checks that line endings are unix style, line feeds.
Windows clones files with carriage return line feeds by default.
Meaning the tests always will fail when testing on a Windows machine.
(for example https://github.com/mdx-js/mdx/pull/812/checks?check_run_id=257393858)

The gitattributes file tells Windows to clone files that have source code with line feed newlines.
Allowing the lint checks to pass.

wooorm:
Ah was wondering what failed, prettier!

ChristianMurphy:
Will this also need to be applied to the v2 branch?

PR #824, opened 2019-10-22
Fix remark-mdx-remove-exports usage result in readme

Luiginator:
The usage example of the readme in 'remark-mdx-remove-exports' shows the wrong result. It shows the result of the 'remark-mdx-remove-imports' package. This PR adjusts the example result.


johno:
Thank you!

Issue #801, opened 2019-10-05
MDXRenderer fails to apply custom delete (del) component


markmichon:
### Custom delete component fails to render

First ran into this over in [system-ui/theme-ui#401](https://github.com/system-ui/theme-ui/issues/401) where applying a custom delete component failed to render when passed into MDXRenderer's component prop. MDX rendered a `del`, but not the custom one.

I've tried tracking down exactly where the issue happens, but only managed to get a few failing tests going in mdx-js and mdx-js/react.

### Steps to reproduce / Expected / Etc

[Here's a commit with the two failing tests](https://github.com/markmichon/mdx/commit/90fc32b48cbc27de65ff4bb30d6cf9134e325fe3), or see the input/output below.

For mdx core:

```js
const components = {
  delete: ({children}) =>
    React.createElement('del', {style: {color: 'crimson'}}, children)
}
it('renders delete component from context', async () => {
  const result = await renderWithReact(`Hello, ~~world~~ MDX!`, {components})

  expect(result).toContain(
    '<p>Hello, <del style="color:crimson">world</del> MDX!</p>'
  )
})
``` 

```
Expected substring: "<p>Hello, <del style=\"color:crimson\">world</del> MDX!</p>
"
    Received string:    "<p>Hello, <del>world</del> MDX!</p>"
```

Similar deal with @mdx-js/react:

```js
const CustomDelete = props => <del style={{color: 'crimson'}} {...props} />

it('Should render delete as del', () => {
  const result = renderToString(
    <MDXProvider components={{delete: CustomDelete}}>
      <Fixture />
    </MDXProvider>
  )

  // CustomDelete is rendered
  expect(result).toMatch(/style="color:crimson"/)
})
```

```
Expected pattern: /style="color:crimson"/
    Received string:  "<h1></h1><h2></h2><p>h3, h4</p>"
```

I suspect it's further up the chain somewhere, so maybe someone with better knowledge of the ecosystem can hunt it down or point me in the right direction. Thanks!

johno:
Thanks for reporting! This definitely looks like a bug. Do you mind opening up a PR with your failing tests @markmichon? That way we could use that as a starting point to get a fix in.

Luiginator:
@markmichon @johno I found some issues in the test cases and created a PR that proofs that the del works also with custom components: https://github.com/mdx-js/mdx/pull/825

markmichon:
[...]

PR #825, opened 2019-10-22
Mdx renderer del working tests

Luiginator:
[This](https://github.com/mdx-js/mdx/issues/801) issue reported that the del is not working correctly. There were some issues in the tests created with the issue. I use this PR to proof that the del substitution is working. I don't think these tests cover any new test case so I don't think we have to merge this PR.


johno:
It seems worth it to merge for now! At some point it might be wise to test all MD elements individually to ensure they're rendered as expected, merging this will remind me to do that.

Thanks for the PR!

JounQin:
`lerna` breaks on Circle CI again after this PR merged, no idea that is happening.

PR #833, opened 2019-10-29
Update ast.mdx

skvg:


johno:
Thanks for the PR! The Button component is exposed as a shortcode so it doesn't need an explicit import. Additionally, the node type in the [mdast](https://github.com/syntax-tree/mdast) is of type `html` rather than `HTML`.

As such, I'm going to go ahead and close this for now, but would love any new docs changes if you encounter anything else! Thanks! 🏅

PR #866, opened 2019-12-15
fix: Cannot read property 'type' of undefined

zbeyens:
Just got this error when using `eslint-mdx`

### DIFF
--- a/packages/remark-mdx/extract-imports-and-exports.js
+++ b/packages/remark-mdx/extract-imports-and-exports.js
@@ -66,7 +66,10 @@ module.exports = (value, vfile) => {
 
   // Group adjacent nodes of the same type so that they can be combined
   // into a single node later, this also ensures that order is preserved
-  let currType = allNodes[0].type
+  let currType
+  if (allNodes.length) {
+    currType = allNodes[0].type
+  }
   const groupedNodes = allNodes.reduce(
     (acc, curr) => {
       // Default export nodes shouldn't be grouped with other exports


JounQin:
Thx @zbeyens , do you have an reproduction about this so that I can add a fixture into `eslint-mdx`?

zbeyens:
@JounQin here is the repo https://github.com/zbeyens/slate-plugins-next/tree/cd3cdd7898fca7387138d684e0ebae8d0ccab81a

Where the error occurs in this file: https://github.com/zbeyens/slate-plugins-next/blob/cd3cdd7898/src/stories/index.stories.mdx

zbeyens:
This error also occurs when running `yarn storybook` as `@storybook/addon-docs` is using `@mdx-js/loader`

JounQin:
@zbeyens Sorry but I tried to run your reproduction with `git checkout cd3cdd7898 && yarn && yarn lint`, there is no error reported, can you provide more details?

JounQin:
@zbeyens I got this error on running `yarn build-storybook`, so maybe it's a bug here, but nothing related to `eslint-mdx`?

zbeyens:
Sorry, currently `yarn lint` doesn't lint `.mdx` files. I got the ESLint error on VSCode.

I think the reason is the wrong syntax used in my `index.stories.mdx` files (I was exporting modules).

I just tried on a correct example and it is working.

JounQin:
@zbeyens Then I have no idea if it is just working as expected, @mdx-js/maintainers?

And another suggestion for your `lint` script: `eslint src --cache --ext md,mdx,js,jsx,ts,tsx`

wooorm:
Reading this conversation it seems there’s a problem with Storybook, not with MDX, so I’ll close this here.

If i misjudged the problem, please respond here and I’ll reopen.

JounQin:
@wooorm It's not only related to storybook, but I don't if the error thrown is expected.

wooorm:
Then I don’t know what the problem is. Reading this thread again, it seems invalid syntax was used. 

@zbeyens, [what is the issue here](https://mdxjs.com/contributing#submitting-an-issue)?

johno:
Ah, so I think it's the export syntax [being used here](https://raw.githubusercontent.com/zbeyens/slate-plugins-next/cd3cdd7898fca7387138d684e0ebae8d0ccab81a/src/stories/index.stories.mdx). I don't believe it's currently being handled.

Can we get a test case added to this PR @zbeyens? This does seem to be the proper fix since `export *` syntax is essentially a noop for MDX.

---

Interestingly `export default {}` syntax is invalid because MDX expects it to be a layout 🤔

JounQin:
It seems there is no error in favor of #911, and the reproduction is also fixed.

https://e58dff50-b397-4d0f-aede-8d4af4c0ea36.ws-ap01.gitpod.io/#/workspace/slate-plugins-next

PR #918, opened 2020-01-21
Update babel-config.mdx

LeeBurton:
corrected the plugin names in the presets example


JounQin:
They are fine. See https://babeljs.io/docs/en/options#name-normalization

LeeBurton:
Wouldn’t work for me 😳

JounQin:
Then you may open an issue on babel's side.

PR #314, opened 2018-11-17
Markdown style

wooorm:
This PR adds code to enforce markdown and prose code style, and fixes docs. Upon merging, it’ll close GH-311.

## Overview

* Add configuration for markdown (excluding the test files)
* Add `@mdx-js/remark-mdx` to parse imports, exports, and JSX to MDAST,
  and stringify them back to markdown. It’s private, and not tested yet,
  but it makes sure remark can run on mdx files too (note: we need to talk
  about the license file there)
* Add `remark` to the `"format"` script, fixing code where possible
* Add `remark` to the `"test"` script, but not outputting in this case
* Add to lint-staged

## Docs

* Use reference links and definitions everywhere
* Use `*` for lists and strong; add more spaces to lists
* Fix broken links, references, and remove superfluous definitions
* Fix AST spec (add comment type, default field)
* Add missing code languages
* Wrap lines around 80 chars
* Use smart quotes (`“”` and `‘’`)

PR #927, opened 2020-01-27
chore: enable prettier for .md/.mdx files - close #844

### LINKED ISSUE #844: [discussion] use `prettier` for `.md/.mdx` also
### Subject of the feature

As title, I'm not quite sure why we are not using `prettier` for `mdx`.

### Problem

With [remark-preset-prettier](https://github.com/JounQin/remark-preset-prettier) we can disable all rules conflict with `prettier`.

### Expected behavior

Enable `prettier` everywhere.

JounQin:
@mdx-js/maintainers Please help to review whether should we do this.


docs/404.mdx
```diff
@@ -1,3 +1,3 @@
 # 404
 
-Page not found.  [Return Home](/)
```
wooorm: This PR seems to break all paragraphs?

JounQin: It seems weird to me, I'll try to figure out.

JounQin:
```
retext-sentence-spacing:double-space - Expected `2` spaces between sentences, not `1`
```
🤣 I should upgrade `remark-preset-prettier` first.

wooorm: Huh, why are you turning spaces and quotes off? They’re there for a reason.
If prettier is removing all spaces, then that’s a problem with prettier?

wooorm: If you want 1 space between sentences, you can change the setting instead of turning it off

JounQin: > If prettier is removing all spaces, then that’s a problem with prettier?

It seems comprehensive effect of `eslint`, `prettier` and `remark-lint` together.

wooorm: > If we're using prettier for all files, they should be handlered by prettier only.

There’s both ESLint and Prettier, right? So why ditch remark?

JounQin: @wooorm It seems I can not override/disable `retext-sentence-spacing` at all by following:

```json
[
  "retext-sentence-spacing",
  {
    "preferred": 1
  }
]
```

Or

```json
[
  "retext-sentence-spacing",
  false
]
```

Any advice again? 🤣

JounQin: > > If we're using prettier for all files, they should be handlered by prettier only.
> 
> There’s both ESLint and Prettier, right? So why ditch remark?

https://github.com/mdx-js/eslint-mdx/blob/master/packages/eslint-plugin-mdx/src/rules/remark.ts#L38-L87

`remark` is under the hook of ESLint rule `mdx/remark`.

wooorm: Where are you doing this?

wooorm: > remark is under the hook of ESLint rule mdx/remark.
 
Right, I know. What I mean is maybe best explained like this: this is a project that uses ESLint. Now, you are adding prettier. Any rule that prettier does differently, you completely remove from ESLint. This is what you are doing by turning of remark-lint rules. You are removing ESLint rules.

I think it’s better to have both ESLint and Prettier. Which can be done by changing the ESLint rules (a.k.a., changing the remark lint rules)

JounQin: > Where are you doing this?

`remarkConfig` field in `package.json`: https://github.com/mdx-js/mdx/blob/chore/prettier/package.json#L128 and then `eslint-plugin-mdx` will read it, so I should check `eslint-plugin-mdx`.

JounQin: @wooorm `eslint-plugin-prettier` is `prettier` for ESLint. `remark` and `prettier` rules could conflict, so `prettier` recommends disable all rules which may conflict with `prettier` like https://github.com/prettier/eslint-config-prettier.

wooorm: Could you provide more information, so I can better help you? What *exactly* is it you are doing.

I don’t understand why you mention `eslint-plugin-mdx`, we’re talking about markdown/mdx here (prose), not jsx/mdx.

JounQin: `mdx` files are now formatted by ESLint with `eslint-plugin-prettier` and `eslint-plugin-mdx` via `retext/remark` related plugins under the hook.

wooorm: How interesting. Well, this is getting really complicated. I suggest removing *everything* that is now in `remarkConfig`, and replacing it with the plugins you want to use instead.

Why use lots of plugins, then turn most of them off, and add more? Wouldn’t it be clearer to just add the things you want?

JounQin: > Why use lots of plugins, then turn most of them off, and add more? Wouldn’t it be clearer to just add the things you want?

Yep, that would be great. `remark-preset-wooorm` has been enabled for a long time, so I don't know are you willing to remove it. Or let's simply use a `prettier` compatible config like https://github.com/1stG/configs/blob/master/packages/remark-config/index.js

wooorm: Feel free to add the config you want.

The PR in its current state is broken though, and adding `remark-preset-wooorm` [caught lots of problems](https://github.com/mdx-js/mdx/pull/314/files) that https://github.com/1stG/configs/blob/master/packages/remark-config/index.js won’t catch.

JounQin: I'll try to use `.remarkrc.js` instead to resolve the conflict problem first.

JounQin: @wooorm OK, I found out the key problem is `remark-retext`. I can not disable `retext-sentence-spacing` directly, but override whole config of `remark-retext` and it works.

https://github.com/mdx-js/mdx/blob/chore/prettier/.remarkrc.js#L10-L18

.eslintignore
```diff
@@ -6,6 +6,7 @@ public
 .next
 .*cache
 *.min.js
+!/.*.js
```
ChristianMurphy: is this intended to ignore many files?
Or just `.remarkrc.js`?

JounQin: .*.js files are ignored by ESLint by default, `!` makes they are not ignored.

docs/advanced/transform-content.mdx
```diff
@@ -19,7 +19,7 @@ Say we have the following file, `example.mdx`:
 import Component from 'Component';
 
 export const meta = {
-    prop: 'value'
+prop: 'value'
```
wooorm: What’s happening here? 🤔

JounQin: It should be `mdx` but the original type is `md`.

JounQin: @wooorm fixed in https://github.com/mdx-js/mdx/pull/927/commits/a5a4fe4bb5973dd24d790a74604c7849e0924f0b

JounQin:
@johno Could you help to uninstall Circle CI? It is really noised after removing its configuration.

docs/advanced/ast.mdx
```diff
@@ -1,11 +1,16 @@
-import { Button } from '@rebass/emotion'
+import {Button} from '@rebass/emotion'
```
JounQin: Actually I prefer the previous one personally which seems more readable.
Shall/Should we change `bracketSpacing` config of `prettier`?

@ChristianMurphy @wooorm @johno

JounQin: And about `trailingComma`, `all` is much better for diffing between commits, or `es5` for compatibility.

wooorm: I personally prefer the “current” style for both (as in: no spaces in braces and no trailing commas)

JounQin: That's OK, we can discuss it in the future of not, let's just keep it as is in this PR.😀

silvenon:
I'm very late to this. This is generally awesome, I'm just sad about curly braces losing surrounding space:

```diff
-import { Button } from '@rebass/emotion'
+import {Button} from '@rebass/emotion'
```

There's no way around that? (No pun intended.) Prettier adds the space in JS, so I'm confused why it doesn't here.

ChristianMurphy:
It is configurable with https://prettier.io/docs/en/options.html#bracket-spacing

silvenon:
Oh, I didn't realize that we have it explicitly configured to `false`, thanks!

PR #937, opened 2020-02-08
Add `sup` to `ComponentType` to support footnotes

eyelidlessness:
I also alphabetized the components to make it easier to read and spot what's available or missing.


johno:
Thank you!

PR #953, opened 2020-02-20
Handle JSX member expressions

johno:
JSX member expressions in MDX weren't being properly
handled. This means that they weren't actually rendered
because the MDX type wasn't passed for the pragma.

So, if your MDX contained something like the following:

```mdx
# Hello, world!

<Foo.Bar />
```

The rendered output was incorrect.

Now, the MDX type is properly added, the shortcode is constructed
if needed, and is pulled out from the components object in
context.

### DIFF
--- a/packages/babel-plugin-apply-mdx-type-props/index.js
+++ b/packages/babel-plugin-apply-mdx-type-props/index.js
@@ -2,6 +2,12 @@ const {types: t} = require('@babel/core')
 const {declare} = require('@babel/helper-plugin-utils')
 const {startsWithCapitalLetter} = require('@mdx-js/util')
 
+const applyMdxType = (node, jsxName) => {
+  node.attributes.push(
+    t.jSXAttribute(t.jSXIdentifier(`mdxType`), t.stringLiteral(jsxName))
+  )
+}
+
 class BabelPluginApplyMdxTypeProp {
   constructor() {
     const names = []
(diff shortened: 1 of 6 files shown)


packages/react/src/create-element.js
```diff
@@ -20,9 +20,21 @@ const MDXCreateElement = forwardRef((props, ref) => {
 
   const components = useMDXComponents(propComponents)
   const type = mdxType
+
+  const getMemberExpressionElement = () => {
+    if (!type.includes('.')) {
+      return
+    }
+
+    const [objName, propName] = type.split('.')
```
jxnblk: Will this *not* handle `<Foo.Bar.Baz />` or am I misunderstanding this?

johno: It won't. This is mostly a temporary monkey patch that will be better addressed in the v2 branch.

ChristopherBiscardi: if we wanted this to be more fully supported we could copy/pasta lodash's `get` code, or something similar: https://lodash.com/docs/4.17.15#get

packages/babel-plugin-apply-mdx-type-props/index.js
```diff
@@ -2,28 +2,41 @@ const {types: t} = require('@babel/core')
...
       api.assertVersion(7)
+      const {types: t} = api
 
       return {
         visitor: {
           JSXOpeningElement(path) {
+            if (t.isJSXMemberExpression(path.node.name)) {
+              const objName = path.node.name.object.name
+              const propName = path.node.name.property.name
+              const fullJsxName = `${objName}.${propName}`
```
ChristopherBiscardi: this would also need to be checked if we swapped to something more full featured in the `getMemberExpression` call. Perhaps we should add a warning if there are too many nested objects.

wooorm:
@johno I suggest ignoring this PR, unless it’s good and ready to go, so we can focus our efforts on MDX@2, which I believe properly addresses it (at least on the MDX parsing side)

JounQin:
@johno I suppose that this PR can be closed because we are not developing on v1 anymore?

PR #959, opened 2020-02-24
Clarify steps to update TS imports

KarnellSchultz:
When using this guide I found that the instructions to edit the tsconfig.json file were a bit vague. Also it assumes the reader knows what 'include' is in the tsconfig file.

Hopefully my edit clarifies those steps a bit.


JounQin:
I don't think `include` is required

KarnellSchultz:
You're correct, now that I'm coming back and testing it again, I found the imports working even without the `import` being included in the tsconfig.

PR #960, opened 2020-02-28
process content before and after mdx compile 

foxundermoon:
add the `beforeCompile` and `afterCompile` option for injecting code when building.

```js
const injection = require('./lib/next/mdxInjection');
const withMDX = require('@foxmn/next-mdx')({
  extension: /\.mdx?$/,	  extension: /\.mdx?$/,
  options: {	  
      rehypePlugins: [rehypePrism],
      beforeCompile: injection.beforeCompile,
      afterCompile: injection.afterCompile,
});
```
## exampe  
- add common header and footer code
https://github.com/WeFoxTech/website/pull/8

```js
const fs = require('fs');
const path = require('path');

async function readMdxLib(name) {
  const fpath = path.join(__dirname, '..', 'mdx', name);
  const content = await fs.promises.readFile(fpath, { encoding: 'utf8' });
  return content;
}

const beforeCompile = async (path, content) => {
  if (/\/website\/pages\//.test(path)) {
    const header = await readMdxLib('pagesHeader.js');
    return `${header}\n${content}`;
  }
  return content;
};
const afterCompile = async (path, content) => {
  if (/\/website\/pages\//.test(path)) {
    const footer = await readMdxLib('pagesFooter.js');
    content = content.replace(/export default function MDXContent/, 'export function MDXContent');
    return `${content}\n${footer}`;
  }
  return content;
};

module.exports = {
  beforeCompile,
  afterCompile,
};
```

### DIFF
--- a/packages/loader/index.js
+++ b/packages/loader/index.js
@@ -12,7 +12,12 @@ const loader = async function(content) {
     filepath: this.resourcePath
   })
 
-  const {renderer = DEFAULT_RENDERER, header, footer} = options
+  const {
+    renderer = DEFAULT_RENDERER,
+    header,
+    footer,
+    defaultExport = true
+  } = options
 
   let headerCode = ''
   if (header) {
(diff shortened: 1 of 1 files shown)


wooorm:
Hi! 👋 Am I correct in assuming that this is still WIP? Your example use shows `beforeCompile` and `afterCompile` as if they were accepted by MDX, but the code in your PR does not define those?

PR #967, opened 2020-03-07
docs: add temporary link of reference

VarunDevPro:
Added a temporary link to refer table of components on the getting-started page. Since I am always forgetting where the list exists


johno:
Thanks!

PR #991, opened 2020-03-23
fix: not evaluate code block starting with \$

### LINKED ISSUE #902: Unable to render \${...} in code block
### Subject of the issue

*Feel free to close if this is not a bug (or bug of some other library).*

I'm trying to render the following

    ```
    python command subdir=\${model.nb_layers}
    ```

but it crashes because MDX tries to evaluate the contents of the `${...}`. I suspect some double escaping occurs and the problem (if indeed a problem) lies in this [line](https://github.com/mdx-js/mdx/blob/master/packages/util/index.js#L32). Or it could be a remark issue. Not entirely sure which.

I also posted on Spectrum [here](https://spectrum.chat/mdx/general/render-in-mdx-code-block~c04330c7-d0e0-4826-a59b-0c185a18036a) because I wasn't sure whether this was a bug.

### Your environment

*   **OS**: macOS 10.15.1
*   **Packages**: `@mdx-js/mdx@1.15.1`
*   **Env**: Node 8.16.2, Yarn 1.16.0

### Steps to reproduce

Repo here - https://github.com/yangshun/hydra/blame/master/website/docs/configure_hydra/workdir.md#L67-L69

### Expected behaviour

Renders fine. It renders ok for GFM.

### Actual behaviour

MDX tries to evaluate the contents of the `${...}`. I suspect some double escaping occurs.

lex111:
_Probably_ fixes #902

For example, this code block

<pre><code>```
python command subdir=\${model.nb_layers}
```
</code></pre>

Produces:

Before:

```
<pre><code parentName="pre" {...{}}>{`python command subdir=\\${model.nb_layers}
      `}</code></pre>
```

-> Uncaught ReferenceError: model is not defined

After:

```
<pre><code parentName="pre" {...{}}>{`python command subdir=\\\$\{model.nb_layers}
      `}</code></pre>
```

-> `python command subdir=\${model.nb_layers}` (same as input)

* * *

Another example:

```
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.' . "$here\$sut"
```

Produces:

Before:

```
<pre><code parentName="pre" {...{}}>{`$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\\.Tests\\.', '.' . "$here\$sut"
      `}</code></pre>
```

-> `$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.' . "$here$sut` (missed slash -> `$here$sut`, should be `$here\$sut`)

After:

```
<pre><code parentName="pre" {...{}}>{`$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\\.Tests\\.', '.' . "$here\\$sut"
      `}</code></pre>
```

-> `$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.' . "$here\$sut"` (same as input)

### DIFF
--- a/packages/util/index.js
+++ b/packages/util/index.js
@@ -32,6 +32,8 @@ const toTemplateLiteral = text => {
     // and ignore "\$" since it's already escaped and is common
     // with prettier https://github.com/mdx-js/mdx/issues/606
     .replace(/`/g, '\\`') // Escape "`"" since
+    .replace(/(\\\$)/g, '\\$1') // Escape \$ so render it as it is
+    .replace(/(\\\$)(\{)/g, '\\$1\\$2') // Escape \${} so render it as it is
     .replace(/\$\{/g, '\\${') // Escape ${} in text so that it doesn't eval
 
   return '{`' + escaped + '`}'
--- a/packages/util/test/test.js
+++ b/packages/util/test/test.js
@@ -44,4 +44,22 @@ describe('toTemplateLiteral', () => {
     // eslint-disable-next-line no-template-curly-in-string
     expect(result).toEqual('{`All the \\${`}')
   })
+
+  it("escapes string with slash in front of '${' so that it is not evaluated", () => {
+    const originalString = String.raw`Hello \${world}`
+    const result = toTemplateLiteral(originalString)
+
+    expect(result).toEqual('{`Hello \\\\\\$\\{world}`}')
+    // eslint-disable-next-line no-eval
+    expect(originalString).toEqual(eval(result))
+  })
+
+  it("escapes string with slash in front of '$' so that it is not evaluated", () => {
+    const originalString = String.raw`My vars $foo\$bar`
+    const result = toTemplateLiteral(originalString)
+
+    expect(result).toEqual('{`My vars $foo\\\\$bar`}')
+    // eslint-disable-next-line no-eval
+    expect(originalString).toEqual(eval(result))
+  })
 })

### CODE REVIEW

**Comment by yangshun** [comment]
cc @omry Finally!

**Comment by wooorm** [comment]
I think this looks great, but I *believe* (as people could use template interpolation) this will be technically breaking and therefore will have to wait for 2.0

**Comment by lex111** [comment]
@wooorm but the other tests are passing, so nothing's broken? Especially [this test](https://github.com/mdx-js/mdx/blob/master/packages/mdx/test/index.test.js#L324) that uses template interpolation, isn't that it? What other case of template interpolation could there be?

Then should I close this PR?

**Comment by wooorm** [comment]
That the tests didn’t fail, and still don’t, may not be the proof that this is non-breaking!

I believe the escape function is only used in text nodes, so in the MD of MDX, whereas the test case you linked is the JSX in MDX.

I’m thinking of this (weird) behavior:

```markdown
<div>

If you’re fine with a slash, this works: \${1 + 1}

</div>

Or <span>\${Math.PI}</span>.
```

Which with current MDX compiles to:

```jsx
    <div>
      <p>{`If you’re fine with a slash, this works: \\${1 + 1}`}</p>
    </div>
    <p>{`Or `}<span>{`\\${Math.PI}`}</span>{`.`}</p>
```

It’s a bit weird though, so I’m curious to hear what others think.

lex111:
> That the tests didn’t fail, and still don’t, may not be the proof that this is non-breaking!

Take it easy. I just meant that the tests should cover the use case that you mean.

In short, close PR if you want, I’m not interested in doing it anymore.

wooorm:
I’m sorry, I didn’t mean to insult you. I don’t want to close the PR.

omry:
@lex111 (@yangshun opened #902 after I ran into the issue).
Thanks for fixing it!

I think @wooorm is just concerned that this may break compatibility and that it's best to land it for the next major version and not as a minor version fix.
It's a legitimate concern that should be evaluated, there is no need to get offended.

@wooorm, in my opinion that the example that you gave seems very unnatural and it seems very unlikely that someone would rely on it. why would anyone in their right mind add a backslash if they don't have to?

lex111:
No problem, although it is strange to me that the tests in this case do not guarantee the correct.
After all, we can apply new replacements for only fenced code blocks. Feel free to go further with this PR.

johno:
Considering it is an "accidental feature" and pretty unnatural to use I think we should be okay merging and releasing as a patch. This "functionality" hasn't been documented and isn't part of the spec.

Thank you very much for the PR. @lex111 and your thoughts all!

bravo-kernel:
👏

PR #1014, opened 2020-04-13
Fixes #990

### LINKED ISSUE #990: jsxFrag should be set with latest version of @babel-plugin-transform-react-jsx
### Subject of the issue

I'm getting the following error when compiling mdx files that use a React Fragment, `<></>` on a Gatsby site.

```
transform-react-jsx: pragma has been set but pragmaFrag has not been set
```

The issue is because MDX specifies `/* @jsx mdx */ as the pragma, but never specifies a pragmaFrag which is now a requirement with `babel-plugin-transform-react-jsx@7.9.x`

See the note here:
https://babeljs.io/docs/en/next/babel-plugin-transform-react-jsx.html#customizing-with-the-classic-runtime

### Your environment

*   **OS**: Mac OS mojave
*   **Packages**:  @mdx-js/mdx, @mdx-js/parcel-plugin-mdx, gatsby-plugin-mdx (Plus others I'm sure) 
*   **Env**: node 12.14, npm 6.13

### Steps to reproduce

Unfortunately the code is private so I can't provide it as an example, but let me know if you want me to recreate in a public repo.

### Expected behaviour

MDX should set a jsxFrag value when specifying a custom jsx pragma.

So for example, every time `/* @jsx mdx */` is set `/* @jsxFrag ... */` should also be set.

Alternatively, babel-plugin-transform-react-jsx should use the default React.Fragment if none is specified.

Related issues:
- babel-plugin-transform-react-jsx: https://github.com/babel/babel/issues/11321
- gatsby-plugin-mdx: https://github.com/gatsbyjs/gatsby/issues/22500

Let me know if there's a fragment pragma that makes sense and I'd be happy to do a PR. 

### Actual behaviour

babel-plugin-transform-react-jsx throws an error.

adammockor:
Replace /* @jsx mdx */ with /* @jsxFrag React.Fragment */.

This fixes problem for me (custom cra + mdx). Could it break something?

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -109,7 +109,7 @@ function sync(mdx, options) {
 
   const {contents} = compiler.processSync(fileOpts)
 
-  return `/* @jsx mdx */
+  return `/* @jsxFrag React.Fragment */
 ${contents}`
 }
 
@@ -124,7 +124,7 @@ async function compile(mdx, options = {}) {
 
   const {contents} = await compiler.process(fileOpts)
 
-  return `/* @jsx mdx */
+  return `/* @jsxFrag React.Fragment */
 ${contents}`
 }
 


wooorm:
thanks for your patience, @adammockor!

I think this will break things, yes, because `jsx` and `jsxFrag` are two separate inline babel options in the classic runtime: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#example

wooorm:
Would you be interested in continuing work on this? E.g., adding some tests to make sure it works, and also for Vue? For my earlier comment, adding both `/** @jsx mdx */` and  / @jsxFrag mdx.Fragment */, and actually adding`mdx.Fragment`, might be the way to go?

eric-burel:
I am not sure I get "automatic" and "classic" runtime stuffs yet, but it seems that the "automatic" runtime expectes you not to set `pragma` and `pragmaFrag` at all, as it triggers `SyntaxError: /code/vulcan-next-starter/README.md: pragma and pragmaFrag cannot be set when runtime is automatic.`.

karlhorky:
Yeah, if both plugins / presets support the automatic runtime, then you do not need to use the pragma comments:

- My first attempt at upgrading to the new Emotion babel preset (where I needed to remove `pragma` and `pragmaFrag`): https://github.com/vercel/next.js/issues/18461#issuecomment-718788358
- The final solution using `preset-react.importSource`: https://github.com/vercel/next.js/issues/18096#issuecomment-729868888
- The supporting PR in Emotion: https://github.com/emotion-js/emotion/pull/1970

karlhorky:
Maybe if you need to continue using the comments on a file-by-file basis (instead of configuring it for the whole project like in my example), the `@jsxImportSource` comment would help:

https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#customizing-the-automatic-runtime-import

packages/mdx/index.js
```diff
@@ -124,7 +124,7 @@ async function compile(mdx, options = {}) {
 
   const {contents} = await compiler.process(fileOpts)
 
-  return `/* @jsx mdx */
+  return `/* @jsxFrag React.Fragment */
```
karlhorky, line 127: Maybe something like this?

```suggestion
  return `/** @jsxImportSource mdx */
```

sgmheyhey (commented):
```
113   return `/* @jsxRuntime classic */
114 /* @jsxFrag React.Fragment */
115 /* @jsx mdx */
116 ${contents}`
117 }
```
```
129   return `/* @jsxRuntime classic */
130 /* @jsxImportSource mdx */
131 /* @jsx mdx */
132 ${contents}`
133 }
```
it's also not working in ubuntu 20.4

karlhorky:
@sgmheyhey these three lines are incompatible:

```js
/* @jsxRuntime classic */
/* @jsxImportSource mdx */
/* @jsx mdx */
```

As far as I understand, you want EITHER:

1. If you're using the `automatic` runtime:

```js
/* @jsxImportSource mdx */
```

2. If you're using the `classic` runtime:

```js
/* @jsxRuntime classic */
/* @jsx mdx */
```

...but then again, I am not an expert here. @Andarist has been super helpful in the past elsewhere (with the integration of Emotion and Next.js).

Andarist:
> these three lines are incompatible

Yes, they are incompatible.

# @jsxImportSource mdx

This means that there is an `mdx` package that:
- has `jsx` & `jsxs` exports that can be imported from `mdx/jsx-runtime`
- has `jsxDEV` export that can be imported from `mdx/jsx-dev-runtime`
- has `createElement` export that can be imported from `mdx`

From what I understand none of this is true - especially given that `mdx` is not even a package name.

# @jsxRuntime classic + @jsx mdx

This is most likely the most correct answer here for now. We'll probably be lifting the `@jsxRuntime` requirement in Babel in the following weeks but it might take some time before this will finally land in tools like CRA etc. I hope that in the future `@jsx mdx` will be enough to reconfigure the transform to use its "classic" variant.

sgmheyhey:
```
npm WARN @mdx-js/tag@0.15.6 requires a peer of react@^0.14.x || ^15.x || ^16.x but none is installed. You must install peer dependencies yourself
```
What is `@mdx-js/tag`? `mdx/packages/remark-mdx/tag.js` is not related with this issue perhaps?

karlhorky:
@Andarist Ah ok thanks!

Alright, so it sounds like according to [your comment above](https://github.com/mdx-js/mdx/pull/1014#issuecomment-732821243), the `classic` runtime is the stopgap method, until MDX supports the React 17 `automatic` runtime.

> From what I understand none of this is true - especially given that `mdx` is not even a package name

Then what needs to happen for compatibility with the new `automatic` runtime is that the MDX package be reworked to provide these exports, as in your comment above:

- `jsx`
- `jsxs`
- `jsxDEV`
- `createElement`

I'm guessing these changes are probably also similar to the changes in https://github.com/emotion-js/emotion/pull/1970

cc @johno @wooorm 

I've edited my comments above.

wooorm:
> until MDX supports the React 17 automatic runtime.

I’m working on trying to see if ditching JSX completely is feasible, which would compile MDX to `React.createElement` (or relevant preact/vue/etc) calls instead.

wooorm:
OK, so to solve this, I think we should add, to these two lines https://github.com/mdx-js/mdx/blob/37835548c9fbc3b2202c195d228ebcb6d58e32d5/packages/mdx/index.js#L9-L10.

This code: `/* @jsxFrag mdx.Fragment */`.

Then, in `packages/react`, `mdx.Fragment` should be exposed and set to `React.Fragment`

For `packages/vue`, well, vue 2 doesn‘t have fragments, so that’s complex. Maybe exposing `mdx.Fragment` there as `"div"` might work? Or set it to `null`, and then in Vue’s `createElement`, handle the first argument (`type`) set to `null` by returning the array of children (if that’s supported in Vue) 🤷‍♂️

Andarist:
Note - I will hopefully file a PR to Babel this week to somewhat make `@jsxRuntime` pragma obsolete.

wooorm:
@Andarist Oh, that sounds interesting! I’d love to read that PR if you’ve posted it.

Andarist:
Sure, we'll try to remember to cc you there - just need to find time for a rebase and adding few tests.

karlhorky:
> OK, so to solve this, I think we should add, to these two lines ... `@jsxRuntime classic`

> Note - I will hopefully file a PR to Babel this week to somewhat make `@jsxRuntime` pragma obsolete

I suppose any usage of the classic runtime is only a temporary workaround? Since there's downsides to it and the new JSX transform looks like the way to go, going forward?

wooorm:
> This upgrade will not change the JSX syntax and is not required. The old JSX transform will keep working as usual, and there are no plans to remove the support for it.
> https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

okonet:
Not sure if it's related to this one but I'm trying to use Theme UI with MDX and I'm getting:

```
importSource cannot be set when runtime is classic.
```

with the following `next.config.js` config:

```
{
  "presets": [
    [
      "next/babel",
      {
        "preset-react": {
          "throwIfNamespace": false,
          "runtime": "automatic",
          "importSource": "theme-ui"
        }
      }
    ]
  ],
  "plugins": ["inline-react-svg", "@emotion/babel-plugin"]
}
```

wooorm:
I don’t think the `automatic` runtime is supported (or even can be, MDX overwrites `React.createElement`).
Can you use `classic` as a runtime, and use `theme-ui` some other way?

Andarist:
The problem is very similar to not being able to use 2 classic pragmas at once. I believe that MDX could use the automatic runtime just fine (it's just a different way of declaring the overwrite) but it still would not play OK with other custom automatic runtimes - as there is no built-in way to compose them.

karlhorky:
Oh, so you mean that there is currently no way to specify multiple `importSource`s like the following non-working code?

```json
// NOT WORKING
{
  "presets": [
    [
      "@babel/preset-react",
      {
        "runtime": "automatic",
        // DO NOT USE, DOES NOT WORK
        "importSource": ["theme-ui", "mdx"]
      }
    ]
  ]
}
```

Interesting drawback - is there anything being worked on to add this capability? Or to do something like this, would a new wrapper library need to be created (eg. `mdx-with-theme-ui`)?

wooorm:
@Andarist “The problem is very similar to not being able to use 2 classic pragmas at once.”

The same can be said for two automatic runtimes too though, right? Or generally: *two runtimes*?

@karlhorky “is there anything being worked on to add this capability to compose runtimes?”

How would that work if both react and preact are used?

React/preact/vue/other hyperscripts are actual renderers (“`h` interfaces”), whereas `emotion` (through `theme-ui`) and `@mdx-js/react` and friends hijack that to change some props and do other things, so your proposed feature might work with that, but then I’d propose `jsxTransformSources` or so.

Emotion really needs to add props to all elements, but I’m personally of the opinion that MDX does not need a runtime/renderer/`h` part: it’s hijacking the renderer so that users can pass components in through providers, whereas I think that those could just as well be passed as a prop. This would then solve @okonet’s question. An alternative is indeed a wrapper like `mdx-with-theme-ui`, although that [seems already provided by theme-ui](https://github.com/system-ui/theme-ui/tree/develop/packages/mdx)?

okonet:
Hmm that's an interesting drawback of this design... I'm wondering how it's going affect tooling.

I agree it feels like Theme UI shouldn't need pragma for their `sx` prop since it's using their own components anyways.

I'm going to take a look at the package you mentioned but the problem there are so many things that start to depend on each other and debugging it is a nightmare.

karlhorky:
> How would that work if both react and preact are used? React/preact/vue/other hyperscripts are actual renderers

Yeah, good question. I'm not as deep into this architecture as I should be to seriously propose anything, but FWIW I guess my idea was based on functional composition, where you can put two functions that have a similar input and output API and string them together (eg. `compose(fn1, fn2)`). But, as you mentioned, in the case of `react` and `preact`, there is no value to have both of them running, so I guess this would not be a use case for this particular API.

So I suppose I was probably thinking about something similar to what you were mentioning with `jsxTransformSources` - the ability to pass in composable transformer functions to either: 

1. add or modify a tree of existing JSX elements (after the runtime has run)
2. (maybe) "plug in to" and replace a part of an existing runtime

Andarist:
> Or to do something like this, would a new wrapper library need to be created (eg. mdx-with-theme-ui)?

Probably this. And it seems that such a package already exists - as @wooorm has found out.

> The same can be said for two automatic runtimes too though, right? Or generally: two runtimes?

Yes, that's exactly what I have meant - sorry if the intention was not clear.

PR #1017, opened 2020-04-17
Remove layoutProps from compiled output

### LINKED ISSUE #742: Remove `layoutProps` from transpiled output

silvenon:
There is no need to collect exports into `layoutProps` just to spread it in the layout component, these variables are already in scope so we can just add them directly.

Unfortunately, this doesn't fix the Next.js issue outlined https://github.com/mdx-js/mdx/issues/742#issuecomment-612652071. Maybe MDX should allow providing a blacklist for certain export names, which won't get added to the layout component.

Fixes #742.


silvenon:
I just saw #778 🤦 I'll take some time to get acquainted with plans for v2.

PR #1018, opened 2020-04-19
Export complete MDX unified processor

silvenon:
This will allow advanced usage besides just compiling MDX string to JSX, like providing a vfile and attaching various data to it via remark and rehype plugins.

### DIFF
--- a/docs/advanced/api.mdx
+++ b/docs/advanced/api.mdx
@@ -13,6 +13,7 @@ the JSX string is created.
 
 - [Async API](#async-api)
 - [Sync API](#sync-api)
+- [Compiler](#compiler)
 
 ## Async API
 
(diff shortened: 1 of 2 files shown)


silvenon:
My use case: I'm extracting various metadata from MDX files, like YAML frontmatter, in Next.js [`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) so I can generate my blog posts using [dynamic routes](https://nextjs.org/docs/routing/dynamic-routes).

Adding `.mdx` files directly to `pages` directory just doesn't give me enough control.

wooorm:
Btw, what are you missing if you do: `unified().use(remarkParse).use(remarkMdx)`?

silvenon:
> Nice! Not necessarily for you/now, but maybe a good idea to add docs for these things?

Done!

> Btw, what are you missing if you do: `unified().use(remarkParse).use(remarkMdx)`?

- I can't call `.process()` because I don't have a compiler, so I have to awkwardly execute phases `parse` and `run`
- MDX adds remark-squeeze-paragraphs and `md-ast-to-mdx-ast.js` before adding remark plugins, so it's not the same
- I can't use rehype plugins (not that I need to, currently)
- having this abstraction would generally make me less anxious about whether I'm doing it right

silvenon:
I'll assume that I'm free to merge this 😃

PR #1023, opened 2020-04-21
feat: bump all (dev)Dependencies, update related usages

JounQin:
bump all possible upgradable (dev)Dependencies, I've tested all examples locally and all tests pass except `zero-example` which seems not compatible with monorepo right now.

`prettier` dropped node 8 support, so I remove node 8 from GitHub Actions too.


packages/mdx/package.json
```diff
@@ -34,7 +34,7 @@
     "mdxast"
   ],
   "dependencies": {
-    "@babel/core": "7.8.4",
+    "@babel/core": "7.9.0",
```
lencioni, line 37: It looks like these versions were pinned here https://github.com/mdx-js/mdx/pull/644

But I'm wondering if there's any reason to not use `^` ranges on these dependencies?

PR #1024, opened 2020-04-22
Export the MDXContent component after its instantiated

johno:
This caused a warning in Firefox because the
isMDXContent assignment was after the export
which caused it to result after the component
return in the transpiled output.

Related: gatsbyjs/gatsby#23171

Co-authored-by: Laurie Barth <laurie@gatsbyjs.com>

### DIFF
--- a/packages/mdx/test/index.test.js
+++ b/packages/mdx/test/index.test.js
@@ -56,7 +56,7 @@ it('Should match sample blog post snapshot', async () => {
 
             const layoutProps = {};
             const MDXLayout = \\"wrapper\\";
-            export default function MDXContent({ components, ...props }) {
+            function MDXContent({ components, ...props }) {
               return (
                 <MDXLayout
                   {...layoutProps}
@@ -70,6 +70,8 @@ it('Should match sample blog post snapshot', async () => {
             }
 
             MDXContent.isMDXComponent = true;
+
+            export default MDXContent;
             "
       `)
 })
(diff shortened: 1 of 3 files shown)


packages/mdx/test/index.test.js
```diff
@@ -325,97 +327,99 @@ test('Should handle layout props', () => {
...
+\\\\\`
+\`}</code></pre>
+    </MDXLayout>;
+}
+
+;
+MDXContent.isMDXComponent = true;
+
+export default MDXContent"
+`)
```
silvenon, line 424: 😅 I think you can solve this noisy diff by just indenting the content of `toMatchInlineSnapshot`. As a separate change we should probably start using `toMatchSnapshot` because the content is large.

mohsenkhanpour:
Any update on this?

wooorm:
@johno Is there anything blocking this? What can I do to help?

wooorm:
Solved by https://github.com/mdx-js/mdx/pull/1382 btw

wooorm:
Hi all! I’m going to close this as it landed on the `main`, the (now default) branch we’re using to gear up to the MDX@2 release.

The reason is so that it’s more clear which issues still need to be solved already and don’t require further work.

Expect another `next` release soonish, which might include it, or maybe it’s already released as a beta!

For more info, see <https://github.com/mdx-js/mdx/issues/1041>.

PR #1008, opened 2020-04-11
@mdx-js/vue: context, createElement and mdx-js/vue-loader

codebender828:
- Implements component context with MDXProvider component
- Creates defaults for `wrapper` and `inlineCode` component
- Added createElement function to vue package

Example: https://mdx-vue.now.sh/

Sadly I wasn't able get the tests to run for the vue-loader package.

Issue #851, opened 2019-11-22
Vue documentation


wooorm:
A bit related to GH-696 I guess, the readme for mdx-js/vue on npm and GH is missing!

codebender828:
Hey, @wooorm ! I'd like to take this on. I'd be happy to talk to someone to advise me on how to go about this. Any advice?

wooorm:
@codebender828 Heya, sorry for the wait, how can I help?

I’d suggest something like the [React readme](https://github.com/mdx-js/mdx/tree/master/packages/react) but showing a Vue example instead!

codebender828:
Hey @wooorm Sorry for my late response as well. Didn't know you had responded to this comment! I was just talking with @johno about this.

So I guess I can create a README in the mdx-js/vue package with some examples to begin with.

codebender828:
Hey @wooorm ! Greetings. Spun up some docs for the vue-loader package in the README file. I also added a Vue CLI example. Does this spark joy?

wooorm:
It does! 👍

PR #1029, opened 2020-04-23
Vue, rebased

ChristopherBiscardi:
continuation of #1008 , having been rebased, tests fixed up, and example app working


codebender828:
Really great work with this @ChristopherBiscardi ! Haven't been able to continue work on this during the last week coz I had some school projects to complete. Glad you were able to help with this! Were the @vue/testu-utils tests for the Vue `MDXComponent` component able to run as well?

Screen Shot 2020-04-26 at 6 11 33 PM

examples/vue/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vue-example",
-  "version": "1.5.9",
+  "version": "2.0.0",
```
johno, line 3:
```suggestion
  "version": "1.5.9",
```

examples/vue/package.json
```diff
@@ -11,7 +11,7 @@
     "vue": "^2.6.11"
   },
   "devDependencies": {
-    "@mdx-js/vue-loader": "^1.5.9",
+    "@mdx-js/vue-loader": "^2.0.0",
```
johno, line 14:
```suggestion
    "@mdx-js/vue-loader": "^1.5.9",
```

package.json
```diff
@@ -41,7 +41,7 @@
     "@mdx-js/mdx": "^1.5.8",
     "@mdx-js/react": "^1.5.8",
     "@mdx-js/test-util": "^1.5.8",
-    "@mdx-js/vue": "^1.5.8",
+    "@mdx-js/vue": "^2.0.0",
```
johno, line 44:
```suggestion
    "@mdx-js/vue": "^1.5.9",
```

packages/vue-loader/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mdx-js/vue-loader",
-  "version": "1.5.9",
+  "version": "2.0.0",
```
johno, line 3:
```suggestion
  "version": "1.5.9",
```

packages/vue-loader/package.json
```diff
@@ -36,7 +37,7 @@
   },
   "dependencies": {
     "@mdx-js/mdx": "^1.5.9",
-    "@mdx-js/vue": "^1.5.9",
+    "@mdx-js/vue": "^2.0.0",
```
johno, line 40:
```suggestion
    "@mdx-js/vue": "^1.5.9",
```

packages/vue/package.json
```diff
@@ -1,30 +1,42 @@
 {
   "name": "@mdx-js/vue",
-  "version": "1.5.9",
+  "version": "2.0.0",
```
johno, line 3:
```suggestion
  "version": "1.5.9",
```

codebender828:
🔥🔥🔥 So much life

wooorm:
Does anyone in the combining Vue + React world know if there’s a difference in how props must be cased on HTML elements? For example, React accepts `className`, `htmlFor`, and things like `strokeMiterlimit` (lower l). Does Vue require the exact same casings?
As casing is not defined by JSX but by React, I’m wondering if this thus creates discrepancies per framework?

sw-yx:
congrats team! is the vue stuff documented anywhere? interested to take a look.

johno:
It's not currently documented, but https://github.com/mdx-js/mdx/issues/851 is the issue tracking docs. Don't have an exact estimate on when the docs will be added (right now it's sort of a "hidden" feature until we further test it and will become officially supoprted in v2).

Also, we [do have an example app in examples](https://github.com/mdx-js/mdx/tree/master/examples/vue) that you can check out in the meantime.

johno:
@wooorm I'm not familiar with a mapping between React and Vue differences and tried to google it a bit (I'm not well-versed in React). We'll prolly want to create a mapping library for any differences since serialization might differ quite drastically between libraries.

codebender828:
@wooorm I think might have an idea of what you're trying to achieve. In Vue 2 the render function attributes are spread out into `attrs`, `props`, and `domProps`. So it looks something like this:
```js
h({
  class: ['foo', 'bar'],
  style: {
    color: 'red'
  },
  attrs: {
    id: 'foo'
  },
  domProps: {
    innerHTML: ''
  },
  on: {
    click: foo
  },
  key: 'foo',
  ref: 'chakra'
}, this.$slots.default)
```

In Vue 3, the plan is to [flatten the `createElement` signature](https://github.com/vuejs/rfcs/blob/render-fn-api-change/active-rfcs/0008-render-function-api-change.md#flat-vnode-props-format). So it will look something like this.
```js
h({
  class: ['foo', 'bar'],
  style: { color: 'red' },
  id: 'foo',
  innerHTML: '',
  onClick: foo,
  key: 'foo',
  ref: 'chakra'
}, this.$slots.default)
```

In any case, I think @johno 's suggestion would be good to create a map between MDX props to Vue props. The Vue 2 props could be mapped to Vue 3 props as well in the `@mdx-js/vue-loader` depending on the consumer's Vue version.

wooorm:
Some more info here: https://github.com/vuejs/jsx/tree/dev/packages/babel-plugin-transform-vue-jsx#difference-from-react-jsx.

It’s a bit unfortunate that JSX is different based on whether it’s for React or Vue (even with Vue 3, as innerHTML or class for example won’t work)

codebender828:
@wooorm . So it seems. The Vue core team seems to be working a JSX implementation of for Vue 3 but I also cannot guarantee that will have the same signature as Vue 2.X. 

https://github.com/vuejs/vue-next#official-libraries-vue-3-support-status

I should also clarify that I don't have enough information about v3's JSX plan. Perhaps someone on the core team could shed some light on whether there will be significant changes to v3's JSX implementation.

If the JSX signatures also has big differences between v2.X and v3, then we can always drop down to the render function layer. It might be simpler to map the render function objects 🙏🏽

wooorm:
The default currently (assuming v2) is to pass things to components directly in, so if a user passes a ‘props’ object in, we won’t touch it. So everything *should* work, except that react mdx and vue mdx are different. We can document that we don’t support Vue’s Babel transform (and be open to a pull request for that if there’s interest?)

“Normal” elements (as in, coming from markdown) are different (think remark or rehype transforms, but also the language flag on code, or the task-list-item class on checklists), these are currently not put into a specific attrs object. In the parser rewrite I did add support for hast-to-hyperscript, which can detect vue or react, and magically switch based on that. However, that’s currently hard coded for React and should be updated. Does the compiler have “knowledge” of whether it’s compiling to React or Vuecurrently?

codebender828:
> Does the compiler have “knowledge” of whether it’s compiling to React or Vuecurrently?

I don't think the compiler is aware of what framework it's compiling to. It just spits out some JSX code. Since Vue doesn't have the JSX pragma support we need to expose the render function within that JSX context block where we call `MDXContent()`.

https://github.com/mdx-js/mdx/blob/34b4f50a1f98b45e35e36c8e0026ddb210c9132f/packages/vue-loader/index.js#L33-L37

It's not very elegant, but it works. I'm not the best JSX expert, but we made a render function that was "aware" of the elements returned by mdx.

https://github.com/mdx-js/mdx/blob/master/packages/vue/src/create-element.js

>  In the parser rewrite I did add support for hast-to-hyperscript, which can detect vue or react, and magically switch based on that.

If this magical switch is possible, then I'm all for it 😄

Issue #1002, opened 2020-04-03
Link to sandbox github is broken


jedierikb:
Link to sandbox github is broken.

Link is from
https://mdxjs.com/playground
to
https://github.com/mdx-js/mdx/edit/master/docs/playground.mdx

wooorm:
Hi there! You’re right, that’s a bug!

Would you be willing and able to submit a pull request to solve it?

PR #1030, opened 2020-04-24
EditLink: Fix trailing slash in pathname

iam-medvedev:
Closes GH-1002

### DIFF
--- a/packages/gatsby-theme-mdx/src/components/edit-link.js
+++ b/packages/gatsby-theme-mdx/src/components/edit-link.js
@@ -8,7 +8,7 @@ export default () => (
   <Location>
     {({location}) => (
       <a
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-mdx/src/components/edit-link.js
```diff
@@ -8,7 +8,7 @@ export default () => (
   <Location>
     {({location}) => (
       <a
-        href={base + location.pathname + '.mdx'}
+        href={base + location.pathname.replace(/\/$/, '') + '.mdx'}
```
silvenon, line 11: Pages with trailing slash are index pages, so this would be the correct fix:

```suggestion
        href={base + location.pathname.replace(/\/$/, '/index') + '.mdx'}
```

PR #1047, opened 2020-05-01
Fix API documentation example

davazp:
The input example was causing mdx API to fail with the current version. Additionally, I updated the output to resemble better the current output, with the user defined export in there.


johno:
Thanks!

PR #1077, opened 2020-05-20
[WIP] MDX V2

johno:
This is a WIP, see https://github.com/mdx-js/mdx/issues/1041 for details.


jaredpalmer:
Can you cut a next.9 release? I think I am blocked by https://github.com/mdx-js/mdx/issues/1081 which I think is making MDX confused / prevents using remark transforms looking for curly braces. In my case, I have plugin to allow for custom heading ids by writing `{#custom-id}` after a heading. Right, now MDX is confused because it thinks that it is private class property.

wooorm:
@jaredpalmer I don’t think cutting a release solves that. That is a private class prop to MDX@next.
Why not do `<h1 myprop="whatever">text</h1>`?

jaredpalmer:
If that's the case then v2 will break any site with `gatsby-remark-autolink-headings` which supports this syntax if you add the `enableCustomId` flag:

```mdx
## Custom Heading {#my-custom-id}

// ... becomes

<h2 id="my-custom-id">Custom Heading</h2>
```

I am working on an existing site that is using css to style markdown and don't have control over the heading the components.

wooorm:
> If that's the case then v2 will break any site with gatsby-remark-autolink-headings which supports this syntax if you add the enableCustomId flag:

That is correct. But you’re getting JS expressions in return.

> I am working on an existing site that is using css to style markdown and don't have control over the heading the components.

I don’t understand this sentence. Are you saying you cannot edit the MDX file?

jaredpalmer:
I figured it out. Sorry for blowing up this issue. It appears that the regex sniffing for the custom curly syntax needed to be adjusted to remove the curly braces.

```diff
const last = node.children[node.children.length - 1];
// This regex matches to preceding spaces and {#custom-id} at the end of a string.
// Also, checks the text of node won't be empty after the removal of {#custom-id}.
- const match = /^(.*?)\s*\{#([\w-]+)\}$/.exec(toString(last));
+ const match = /^(.*?)\s*#([\w-]+)$/.exec(toString(last));
```

wooorm:
Oh good, and no problem!

Could you clarify your comment so that it described where this change should be made (it looks like somewhere in gatsby? 🤔), which might be useful for other folks stumbling on it?

To add: this might again break. Expressions are becoming a first class citizen in MDX, and with it comes that they must contain valid JavaScript expression. Fwiw, Gatsby knows about this: https://github.com/gatsbyjs/gatsby/discussions/25068

wooorm:
For folks finding this: the `next` branch is now the `main` (and default) branch

Issue #1078, opened 2020-05-20
Add TypeScript types to the `next` branch


johno:
@ChristianMurphy do you think we can grab the code changes from #811 and apply to the `next` branch? That's the new working branch for v2 since `v2` got so stale 😬😞. I just tried to cherry pick this commit but there have been some monorepo changes so wasn't sure how the `dtslint` side of things would work.

Would love to get this in there and make sure that you get the code credit!

ChristianMurphy:
Thanks for the ping @johno!
I'll take a look at migrating those over to the `next` branch.

PR #1079, opened 2020-05-20
types: add typescript typings for @mdx/loasder, @mdx-js/mdx, @mdx-js/react, @mdx-js/runtime, @mdx-js/vue, @mdx/vue-loader

ChristianMurphy:
re-applies #811
is part of #1078

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
     "publish": "lerna publish --force-publish=\"*\"",
     "publish-ci": "lerna publish -y --canary --preid ci --pre-dist-tag ci",
     "publish-next": "lerna publish --force-publish=\"*\" --dist-tag next --preid rc",
-    "test": "jest --runInBand --detectOpenHandles && nyc --reporter lcov tape packages/remark-mdx/test/index.js"
+    "test": "yarn test-jest && yarn test-types",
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -50,6 +52,7 @@
     "babel-plugin-macros": "2.8.0",
     "babel-plugin-remove-export-keywords": "^1.6.2",
     "babel-plugin-transform-vue-jsx": "4.0.1",
+    "dtslint": "3.6.3",
```
ChristianMurphy, line 55: pulled dtslint to the top level, since the same version can and should be used to test all packages

packages/loader/package.json
```diff
@@ -31,6 +34,9 @@
     "webpack",
     "loader"
   ],
+  "scripts": {
+    "test-types": "dtslint types"
```
ChristianMurphy, line 38: dtslint doesn't seem to support patterns.
I tried adding `dtslint packages/*/types` to handle all packages at once, that didn't work, so opted to add the script at the package level, and use lerna to run all of them.

ChristianMurphy:
ci failure matches current `next` branch error https://github.com/mdx-js/mdx/runs/694041854

johno:
Awesome, thanks! I'll go ahead and fix the error in `next` after I merge this!

PR #1082, opened 2020-05-21
types: add typescript typings for remark-mdx, remark-mdx-remove-expor…

ChristianMurphy:
…ts, remark-mdx-remove-imports, @mdx-js/util

### DIFF
--- a/packages/remark-mdx/package.json
+++ b/packages/remark-mdx/package.json
@@ -2,7 +2,24 @@
   "name": "remark-mdx",
   "version": "2.0.0-next.0",
   "description": "remark plugin to add support for mdx",
+  "repository": "https://github.com/mdx-js/mdx/tree/master/packages/remark-mdx",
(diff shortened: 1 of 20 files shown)


packages/remark-mdx/package.json
```diff
@@ -2,7 +2,24 @@
   "name": "remark-mdx",
```
ChristianMurphy, line 2: prettier reordered the properties when this was updated with the type metadata

ChristianMurphy:
@johno @wooorm I noticed there are both `remark-mdx` and `remark-mdxjs` plugins.
I added types for `remark-mdx` as it appears to be stable (?)
`remark-mdxjs` has a warning about being in alpha, is that going to be published as part of v2?

johno:
I’m thinking we will recombine them for v2 to lessen breaking changes, so you can prolly hold off on typing `remark-mdxjs`.

Thanks for the PR!

PR #1084, opened 2020-05-24
Add support for namespace import

### LINKED ISSUE #1050: Add support for namespace imports
```js
import * as Foo from './bar'
```

ankeetmaini:
Closes #1050

### DIFF
--- a/packages/babel-plugin-apply-mdx-type-props/index.js
+++ b/packages/babel-plugin-apply-mdx-type-props/index.js
@@ -12,7 +12,7 @@ class BabelPluginApplyMdxTypeProp {
 
       return {
         visitor: {
-          JSXOpeningElement(path) {
+          JSXOpeningElement(path, state) {
             const jsxName = path.node.name.name
 
             if (startsWithCapitalLetter(jsxName)) {
@@ -24,6 +24,15 @@ class BabelPluginApplyMdxTypeProp {
                   t.stringLiteral(jsxName)
                 )
               )
+
+              const {namespacedImports} = state.opts
+              const isNamespaceImport =
+                namespacedImports &&
+                namespacedImports.find(s => s.includes(jsxName))
+
+              if (isNamespaceImport) {
+                path.node.name.name += '.default'
+              }
             }
           }
         }
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -134,13 +134,17 @@ MDXContent.isMDXComponent = true`
     const babelPluginApplyMdxPropInstance = new BabelPluginApplyMdxProp()
     const babelPluginApplyMdxPropToExportsInstance = new BabelPluginApplyMdxProp()
 
+    const namespacedImports = importStatements
+      .split('\n')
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-apply-mdx-type-props/index.js
```diff
@@ -24,6 +24,15 @@ class BabelPluginApplyMdxTypeProp {
...
                 )
               )
+
+              const {namespacedImports} = state.opts
+              const isNamespaceImport =
+                namespacedImports &&
+                namespacedImports.find(s => s.includes(jsxName))
+
+              if (isNamespaceImport) {
+                path.node.name.name += '.default'
```
JounQin, line 34: What if it's a commonjs module?

ankeetmaini: This will still work. I tested this using parcel example.

> a component exported using CommonJS
```jsx
// Counter.js
import React from 'react'

const Counter = () => {
  return <div style={{fontSize: 40}}>Not a counter</div>
}

module.exports = Counter
```

> mdx file importing it as a namespaced import

```mdx
import * as Counter from './Counter'

# Hello, world!

<Counter />
```

> Used parcel example to test this.

Screenshot 2020-05-25 at 12 08 35 PM

ankeetmaini: This will still work if the export was done on a commonjs module.

wooorm:
I don’t have enough knowledge about how different bundlers handle default exports to properly review whether this would work. But: more tests with CJS and more types of ESM might help?

wooorm:
@ankeetmaini Sorry for the wait, thanks for your patience.

I’m not convinced: I’m not sure what problem this is solving. And @JounQin’s comment rings true to me.
We’ll soon also have a JS AST: https://github.com/mdx-js/mdx/pull/1382.

What is this trying to solve?

wooorm:
I can confirm that `default` is a field on namespace imports of both ESM and CJS files. Although: in Node they are, Webpack (4) does weird things, so it might not work there:

`a.mjs`:

```js
import * as b from "./b.mjs"
import * as c from "./c.cjs"

console.log('xxx:', b, c);
```

`b.mjs`:

```js
export const x = () => 1
export default () => 2
```

`c.cjs`:

```js
module.exports = () => 2
module.exports.x = () => 1
```

Now, running `node a.mjs` yields:

```
xxx: [Module] { default: [Function: default], x: [Function: x] } [Module] {
  default: [Function (anonymous)] { x: [Function (anonymous)] },
  x: [Function (anonymous)]
}
```

Note that both are an object with a `default` field.

---

However, I don’t think we should take this identifier and add `.default` to it. That seems like your responsibility as an MDX author.

But, as @johno’s WIP PR is open (https://github.com/mdx-js/mdx/pull/953), solving a similar issue, I’m now lost at what #1050 is about. John, could you clarify what is broken? The issue (https://github.com/mdx-js/mdx/issues/1050) is rather short

JounQin:
close due to inactive.

Please contribute to `main` branch instead.

PR #1107, opened 2020-06-10
Fix npm command to start zero example

iwgx:
Because in `package.json` script, looks like we run it with start command


johno:
Thanks!

PR #1109, opened 2020-06-10
fix: add useEffect to Web Components example

marcysutton:
A user was confused about how to use the Web Components example on Twitter, as the call to `window` was causing issues with SSR. This PR adds useEffect to the example to prevent against it.


johno:
Thank you!

PR #1134, opened 2020-07-09
ci: Github CI Fixes

palfrey:
This PR fixes various things that were breaking the Github CI
* Targeted fairly minimal React upgrades so that the `yarn --frozen-lockfile` works again
* Fix all the Prettier errors


JounQin:
Thanks!

palfrey:
Master CI is still broken, but that's in the `Publish CI tag to npm` step that wasn't tested here, and I can't easily test myself!

PR #1156, opened 2020-07-17
fix: `options.file` could be `undefined` error in `toJSX`

dustinsavery:
`toJSX` could be called outside, see [storybook](https://github.com/storybookjs/storybook/blob/next/addons/docs/src/mdx/mdx-compiler-plugin.js#L314), where `options.file` could be unavailable.

### DIFF
--- a/packages/mdx/mdx-hast-to-jsx.js
+++ b/packages/mdx/mdx-hast-to-jsx.js
@@ -120,8 +120,9 @@ MDXContent.isMDXComponent = true`
 
     // Check JSX nodes against imports
     const babelPluginExtractImportNamesInstance = new BabelPluginExtractImportNames()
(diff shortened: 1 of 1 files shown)


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -120,8 +120,9 @@ MDXContent.isMDXComponent = true`
 
     // Check JSX nodes against imports
     const babelPluginExtractImportNamesInstance = new BabelPluginExtractImportNames()
+    let filename = options && options.file && options.file.path ? options.file.path : ""
```
JounQin, line 123:
```suggestion
    const filename = options.file && options.file.path
```

dustinsavery: Confirmed, your idea works as well.

PR #1157, opened 2020-07-17
chore(repo): pin internal package dependencies

### LINKED ISSUE #1090: MDX does not work with mismatched version of mds-utils
### Subject of the issue

When consuming MDX dependencies, we sometimes get errors when MDX updates related packages.

We have pinned our versions, but it seems MDX does not.

https://github.com/mdx-js/mdx/blob/master/packages/mdx/package.json#L40

Since version 1.6.1 of mdx react or run-time does not actually work with mismatched versions of utils, would you consider pinning your internal deps?

### Your environment

*   **OS**: MacOX
*   **Packages**: "@mdx-js/react" + "@mdx-js/run-time"
*   **Env**:  yarn: 1.21.1.    node: v12.14.1

### Steps to reproduce

pin your version of react or run-time to `1.6.1`. running yarn install will install `1.6.4` of utils.  This means the rendering no longer works as expected

we caught this error as we test for our class names to be applied to the md component:

```sh
dxInput › raw markdown › renders table correctly as markdown
    expect(received).toContain(expected) // indexOf
    Expected substring: "<table class=\"table--1clGC\">"
    Received string:    "<div class=\"mdxPlacement\" data-id=\"mdx-output\"><div class=\"component--3kq6z isPaddingMedium--3DaAA\" data-qa=\"card\"><div class=\"preview\"><table><thead><tr><th>c1</th><th>c2</th></tr></thead><tbody><tr><td>r1</td><td>r1</td></tr><tr><td>r2</td><td>r2</td></tr></tbody></table></div></div></div>"
```

our table component (the `Table` applies it's own classes and this is what breaks)
```
import React from 'react';
import { Table } from '@clearscore/rainbow.table';

import Spacing from '../spacing';

const BlockWrapper = () => ({ children }) => (
    <Spacing type={Spacing.types.TABLE}>
        <Table>{children}</Table>
    </Spacing>
);

export default BlockWrapper;
```

### Expected behaviour

for a pinned version of mdx react/run-time to work after a release of utils goes out.

### Actual behaviour

tests fails and rendering breaks when a new version of utils is released.

peter-mouland:
fixes #1090


peter-mouland:
@laurieontech @silvenon @johno I see there have been 5 releases in 5 days.  Although I love small and frequent releases, it would be really great to get this PR in so that those releases don't constantly break builds due to the mis-matched versions.

Is there anything I can do to help get this PR in sooner than later?

johno:
Will merge now, thank you @peter-mouland!

johno:
Ugh, looks like lerna automatically turns the interlinked packages back to the caret `^x.x.x`. Gonna need to dig into their docs to see if there's an option to not do that 🤔

peter-mouland:
maybe this is needed?

```
  "exact": true,
```

peter-mouland:
@johno found the docs for this flag https://github.com/lerna/lerna/blob/master/commands/init/README.md#--exact

peter-mouland:
https://github.com/mdx-js/mdx/pull/1167

Issue #1152, opened 2020-07-16
MDX performance


johno:
**☂️ This umbrella issue is for tracking work related to improving performance to MDX.**

I've been working with @pvdz on MDX performance. We've noted a few aspects that add unnecessary work which we should be able to reduce, especially in v2.

## Numerous babel parse and transformation steps

Firstly, we have multiple babel parse steps throughout the MDX transpilation pipeline.

### Imports and exports

- Partitioning imports and exports
- Finding the default export

Peter has done some work here in `gatsby-plugin-mdx` that we can potentially adapt gatsbyjs/gatsby#25437 for usage in core.

### Shortcode generation

We use babel to figure out what imports and exports exist, and then use that to instantiate variables coming from `MDXProvider` with `makeShortcode`. Also related to gatsbyjs/gatsby#25437

### `mdxType`

This is used by the runtime (react/preact/vue) to determine which component to render. This is something we can do from the MDXAST in v2 since the JSX structure is represented.

## Returning a compiled string that inevitably needs to be transpiled

Secondly, to these parse steps we also return a JSX string. In nearly all cases this JSX string is then transpiled to JS and `mdx` pragma function calls. This was originally an intentional output because we wanted to make MDX more palatable and familiar. However, it might make sense to serialize directly to function calls and JS.

This would remove a babel step users need (unless they're using optional syntax or need browser polyfills which is still achievable in user land).

---

You all are welcome to bring up other areas of the codebase we can make more performant or other ideas as well! In fact, we'd love your thoughts.

wooorm:
> However, it might make sense to serialize directly to function calls and JS.

Even if we compile `# This` and `<That />` to function calls, I can still see folks using JSX inside JSX, or inside expressions though, and expect of MDX for them to be handled:

```js
<Heading icon={<Icon />}>

## Heading {something ? <Y /> : <Z />}
```

johno:
> I can still see folks using JSX inside JSX, or inside expressions though, and expect of MDX for them to be handled

Yeah, I think what we can do is transpile JSX to function calls inside expressions potentially. I _should_ still be a lot faster than handling the whole document (of course this is something we will benchmark to be sure).

wooorm:
Probably also faster if we *only* process jsx there, and nothing else, leaving that up to folks. But indeed, wondering on the benchmarks of 100 expressions vs 1 file

pvdz:
[...]

PR #1160, opened 2020-07-17
Improve export name extraction for shortcode generation

johno:
Many different types of export syntax weren't supported,
primarily destructuring for objects and arrays. This adds
export extraction to the import name extraction step so
we can more accurately determine what shortcodes to
generate.

We will eventually want to use something different for
extraction (#1152), but for now let's just make it right before
we make it fast.

### DIFF
--- a/packages/babel-plugin-extract-export-names/index.js
+++ b/packages/babel-plugin-extract-export-names/index.js
@@ -38,7 +38,7 @@ class BabelPluginExtractExportNames {
       }
 
       const handleSpecifiers = node => {
-        const {specifiers = []} = node
+        const {specifiers} = node
 
         if (!specifiers) {
           return
--- a/packages/babel-plugin-extract-export-names/index.js
+++ b/packages/babel-plugin-extract-export-names/index.js
@@ -10,8 +10,11 @@ class BabelPluginExtractExportNames {
       const {types: t} = api
 
       const handleDeclarations = node => {
-        const {declarations} = node.declaration || {}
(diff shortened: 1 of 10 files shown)


packages/babel-plugin-extract-export-names/index.js
```diff
@@ -0,0 +1,65 @@
...
+    this.state = {names}
+
+    this.plugin = declare(api => {
+      api.assertVersion(7)
+      const {types: t} = api
+
+      const handleDeclarations = node => {
+        const { declarations } = node.declaration || {}
+
+        if (!declarations) {
```
laurieontech, line 15: Won’t this always be false? {} is truthy

johno: I should probably write this differently so it's more clear but it _should_ work since `declarations` is being destructured:

```js
> const {declarations} = null || {}
undefined
> declarations
undefined
```

johno: Addressed in https://github.com/mdx-js/mdx/pull/1160/commits/44e3e49270682357997bfa246a337247bd2bd960 to make it more clear.

Also fixed an incorrect handling of the default below which was incorrect 5f73b0b5b961c45ca170204e46fd8525a741851b.

laurieontech: Ah, ya. It’s just a weird behavior to make use of. But I gotcha :)

PR #1167, opened 2020-07-20
pin workspace deps

peter-mouland:
Adding `exact: true` to prevent `^`
https://github.com/lerna/lerna/blob/master/commands/init/README.md#--exact


johno:
Awesome, thank you!

PR #1175, opened 2020-07-22
Implement ability to opt out of MDXProvider components

johno:
Related to #821

### DIFF
--- a/packages/react/src/context.js
+++ b/packages/react/src/context.js
@@ -11,23 +11,24 @@ export const withMDXComponents = Component => props => {
 
 export const useMDXComponents = components => {
   const contextComponents = React.useContext(MDXContext)
-  let allComponents = contextComponents
-  if (components) {
-    allComponents = isFunction(components)
-      ? components(contextComponents)
-      : {...contextComponents, ...components}
+
+  // Custom merge via a function prop
+  if (isFunction(components)) {
+    return components(contextComponents)
   }
 
-  return allComponents
+  return {...contextComponents, ...components}
 }
 
-export const MDXProvider = props => {
-  const allComponents = useMDXComponents(props.components)
+export const MDXProvider = ({components, children, disableParentContext}) => {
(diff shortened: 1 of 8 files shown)


johno:
I believe this change will require a change to types for `@mdx-js/react` and `@mdx-js/preact` since the MDXProvider component can now accept `components={false}`.

ChristianMurphy:
This is the type for the MDXProvider

https://github.com/mdx-js/mdx/blob/f59b174e828538353dbec2189ec361c8f0bd0817/packages/react/types/index.d.ts#L19-L27

the update would look like

```diff

 /** 
  * Prop type that includes a component dictionary 
  */ 
 interface ComponentsProp { 
   /** 
    * Mapping of names for JSX components to React components 
    */ 
   components: ComponentDictionary 

+  /**
+   * description of new property goes here
+   *
+   * @defaultValue true
+   */
+  useParentContext?: boolean
 }

File: packages/react/src/context.js [hunk 1]
```diff
@@ -11,23 +11,28 @@ export const withMDXComponents = Component => props => {
...
-  return allComponents
+  return {...contextComponents, ...components}
 }
 
-export const MDXProvider = props => {
-  const allComponents = useMDXComponents(props.components)
+export const MDXProvider = ({
+  components,
+  children,
+  useParentContext = true
```
ChristianMurphy, line 26: :thinking: 
JSX/HTML generally makes the absences of an attribute false, and the presence true.
Here the absence is true, the presence is also true, and only an explicit `false` will disable it.

This could both cause issues with adopters that use https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md and be not intuitive for adopters to use.

Maybe `disableParentContext` would make sense?
Thoughts? :thought_balloon:

johno: Yeah, that's a fair point. A `disable` style attr would be more ergonomic because that's what a user is _intending_. Will update!

PR #1178, opened 2020-07-23
ci(github): update github actions and dtslint to latest

ChristianMurphy:
dependency updates may help resolve flaky tests which have been occuring on `next` branch.

dtslint updated using `yarn upgrade dtslint`

related to #1172

PR #1180, opened 2020-07-23
ci(github): update github actions to latest versions

ChristianMurphy:
follow up to #1178 also updates lint and publish pipelines to use new checkout and setup node versions.


johno:
Thanks!

PR #1183, opened 2020-07-24
Update sponsors table

wooorm:
A couple of new sponsors and a different “leaderboard”, so here’s an update that I did for the other unified projects but forgot here. The biggest sponsors are a bit big, but this is the best I could get with HTML in GH and fit everyone on there 😅


wooorm:
Readme live: https://github.com/mdx-js/mdx/blob/5311d7f5afdbfcdb07373ef0af9407c37cd0e1c0/readme.md#sponsors

Issue #1210, opened 2020-08-03
skipExport does not exist in type mdx(typescript)


lonyele:
### `skipExport` does not exist in type `Options`
I had to use the code from `Do it yourself` on `Getting Started` and got into the problem where vscode says there is no `skipExport` at `mdx`(in my case, `mdx.sync()`)  
<img width="874" alt="스크린샷 2020-08-03 오후 4 42 17" src="https://user-images.githubusercontent.com/15892571/89158565-4a3e2f00-d5a9-11ea-8e25-e617d4380d4d.png">

### Your environment

*   **Mac 10.15.2**: 
*   **2.0.0-next.7**: 

### Steps to reproduce

Tell us how to reproduce this issue.  Please provide a working and simplified example.

🎉 BONUS POINTS for creating a [minimal reproduction](https://stackoverflow.com/help/mcve) and uploading it to GitHub.  This will get you the fastest support.  🎉

### Expected behaviour

`skipExport` option should have existed at `mdx` api

### Actual behaviour

`skipExport` option doesn't exist

PR #1211, opened 2020-08-03
fix(mdx): add additional typings to mdx api

lonyele:
Issue: #1210

## What I did
Add `skipExport` and `wrapExport` typings to mdx api

## Context
First of all, thanks for adding some typescript support to mdx. I found that as #1079 was landed, it misses some typings. I personally had a problem with `skipExport` but I found that `wrapExport` looks like it needs typing too. Let me know if something needs to be changed

### DIFF
--- a/packages/mdx/types/index.d.ts
+++ b/packages/mdx/types/index.d.ts
@@ -19,7 +19,7 @@ declare namespace mdx {
     skipExport?: boolean
 
     /**
-     * wrap 'export default' statement with provided string when serialize
+     * wrap 'export default' statement with provided string when serializing
      * to JSX
      */
     wrapExport?: string
--- a/packages/mdx/types/index.d.ts
+++ b/packages/mdx/types/index.d.ts
@@ -12,7 +12,7 @@ declare namespace mdx {
     footnotes?: boolean
 
     /**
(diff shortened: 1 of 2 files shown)


lonyele:
hm... both `yarn test` and `yarn test-types` works ok on my machine(Mac 10.15.2, node 14.4.0, yarn 1.21.1) Probably help is needed here

packages/mdx/types/index.d.ts
```diff
@@ -11,6 +11,19 @@ declare namespace mdx {
      */
     footnotes?: boolean
 
+    /**
+     * skip the addition of 'export default' statement when serialize
```
johno, line 15:
```suggestion
     * skip the addition of 'export default' statement when serializing
```

packages/mdx/types/index.d.ts
```diff
@@ -11,6 +11,19 @@ declare namespace mdx {
...
+     * to JSX
+     * @default false
+     */
+    skipExport?: boolean
+
+    /**
+     * wrap 'export default' statement with provided string when serialize
```
johno, line 22:
```suggestion
     * wrap 'export default' statement with provided string when serializing
```

PR #1212, opened 2020-08-03
Add details on how to add MDX plugins with Next.js

samrobbins85:
Adds detail from the Next.js package on how to use MDX plugins


johno:
Thanks!

PR #1213, opened 2020-08-09
Update typescript.mdx

gamliela:
This is a documentation improvement, on the Typescript page.

This change adds default HTML properties and the `metastring` prop to the list of declared props of MDX components.

The existing declaration only allows the `children` prop for MDX components. It will not work, for example, if you'll try to expand the [syntax-highlighting example](https://mdxjs.com/guides/syntax-highlighting) to use `className` or `metastring` props.

I've ran `yarn test` and `yarn docs` to verify this change.

### DIFF
--- a/docs/advanced/typescript.mdx
+++ b/docs/advanced/typescript.mdx
@@ -39,8 +39,8 @@ declare module '@mdx-js/react' {
     | 'ul'
 
   export type ComponentProps = React.HTMLProps<Element> & {
-    metastring?: string;
-  };
+    metastring?: string
+  }
 
   export type Components = {
     [key in ComponentType]?: React.ComponentType<ComponentProps>
--- a/docs/advanced/typescript.mdx
+++ b/docs/advanced/typescript.mdx
@@ -38,8 +38,12 @@ declare module '@mdx-js/react' {
     | 'tr'
     | 'ul'
 
+  export type ComponentProps = React.HTMLProps<Element> & {
+    metastring?: string;
+  };
+
   export type Components = {
-    [key in ComponentType]?: React.ComponentType<{children: React.ReactNode}>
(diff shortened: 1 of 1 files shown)


docs/advanced/typescript.mdx
```diff
@@ -38,8 +38,12 @@ declare module '@mdx-js/react' {
     | 'tr'
     | 'ul'
 
+  export type ComponentProps = React.HTMLProps<Element> & {
+    metastring?: string
+  }
+
   export type Components = {
-    [key in ComponentType]?: React.ComponentType<{children: React.ReactNode}>
+    [key in ComponentType]?: React.ComponentType<ComponentProps>
```
ChristianMurphy, line 46: I'm not sure this is accurate.
The default components are based of html elements, but it can be extended with additional elements which may not be HTML tag based.

ChristianMurphy: For example `inlineCode` and `thematicBreak` are both not directly HTML types, but are part of the `ComponentType` list, which this builds on.

gamliela: Thanks for reviewing this!
Yes, it is not accurate enough. I can replace `ComponentProps` with `any` like [v2](https://github.com/mdx-js/mdx/blob/v2/packages/react/types/index.d.ts) does. Would that be alright?

ChristianMurphy: Sure, that could work.

gamliela: cool, done

gamliela:
@ChristianMurphy, Thanks for taking the time to review this change!

wooorm (approved):

wooorm:
Thanks for your patience @gamliela!

gamliela:
Thanks @wooorm ! good work!

PR #1218, opened 2020-08-10
Runtime - Added documentation for remarkPlugins

dowi:

### DIFF
--- a/packages/runtime/readme.md
+++ b/packages/runtime/readme.md
@@ -47,6 +47,12 @@ const scope = {
   some: 'value'
 }
 
+// Provide optional remark plugins array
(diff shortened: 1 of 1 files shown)


wooorm:
This sounds like `remarkPlugins` is working with runtime, but it’s missing from the `Props` table? https://github.com/mdx-js/mdx/blob/7fefef74a07139a3ca034eef5c161c9797bd03b2/packages/runtime/readme.md#props. If so, it should be described in the table. And do `rehypePlugins` also work there?

I’m not sure about changing the example to include all potential uses. The example should be short. So I’d vote to not change the example.

johno (approved):
Thanks!

PR #1269, opened 2020-09-17
modernize 2.0 preact build

ChristopherBiscardi:
This sets us up to modernize all of our builds by implementing microbundle on the preact package. This produces a commonjs build that is left as main because that's what people currently expect, an es build that gets used by webpack, etc under the module key, and a true es modern build that will get used conditionally by node v14 native esm when `import` is used. When `require` is used in v14, the commonjs build gets used.

After confirming this works in the preact package in the 2.0 release, we can do the same for our other builds. This puts us onto a well documented, maintained build tool for modern releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "node": "12.x"
   },
   "scripts": {
-    "build": "r -f cjs,esm -o dist --exclude 'babel-plugin-*,create-mdx,*loader,mdx,parcel-plugin-mdx,remark-*,*util'",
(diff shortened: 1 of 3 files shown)


johno (approved):
🏗️

johno:
Looks great! We should probably move all our builds over to this.

ChristopherBiscardi:
> We should probably move all our builds over to this.

tracking in https://github.com/mdx-js/mdx/issues/1271

PR #1289, opened 2020-10-05
Add TypeScript error related to mdx imports for search indexing

nbaksalyar:
This is a small addition to the TypeScript documentation which adds the actual error that a user sees when importing `*.mdx` files as modules. 

The main reason to include it is to make sure that this page is picked up and indexed by search engines; I tried searching by the error description and couldn't find anything relevant for MDX.

### DIFF
--- a/docs/advanced/typescript.mdx
+++ b/docs/advanced/typescript.mdx
@@ -56,8 +56,8 @@ improve the developer experience for TypeScript users.
 ---
 
 If you’re getting errors from TypeScript related to imports with an `*.mdx`
-extension, create an `mdx.d.ts` file in your types directory and include it
-inside your `tsconfig.json`.
+extension (e.g., `Cannot find module 'typescript.mdx' or its corresponding type declarations`),
+create an `mdx.d.ts` file in your types directory and include it inside your `tsconfig.json`.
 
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
Sorry you ran into a spot of trouble @nbaksalyar, thanks for sharing your findings.
Something to be aware of MDX 2 will come with typings built in so these steps will be lessened/elimated.
You can find more on MDX 2 here: https://github.com/mdx-js/mdx/issues/1041

> The main reason to include it is to make sure that this page is picked up and indexed by search engines

I'm cautious about this,
this change by itself could be okay, but, it seems to set a precedent of trying to enumerate (all) possible errors for SEO.
I'm not sure that's something that would be sustainable to keep up-to-date.

nbaksalyar:
Excellent news about MDX 2 and great to know that it will support TypeScript natively.

> it seems to set a precedent of trying to enumerate (all) possible errors for SEO.
> I'm not sure that's something that would be sustainable to keep up-to-date.

Now that you've mentioned it, I think Stack Overflow would be the best choice for this as they also have great search engine visibility.

So I'm going to close this PR. Thanks for reviewing!

PR #1300, opened 2020-10-17
fix: resolve dependencies to avoid relying on hoisting

merceyz:
**What's the problem this PR addresses?**

The `@mdx-js/*` family of packages relies on hoisting to bring all dependencies to the top level which is not guaranteed to happen and is undefined behaviour.

For a refresher on hoisting, I suggest giving this a read https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

**How did you fix it?**

- Add missing peer dependency `react` to `@mdx-js/loader` (inherited from `@mdx-js/react`)
- `require.resolve` default renderer dependencies in `@mdx-js/loader`
- `require.resolve` injected imports `react` and `@mdx-js/react` in `@mdx-js/parcel-plugin-mdx`
- `require.resolve` injected imports `@mdx-js/vue` in `@mdx-js/vue-loader`

### DIFF
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -35,5 +35,8 @@
     "@mdx-js/mdx": "1.6.19",
     "@mdx-js/react": "1.6.19",
     "loader-utils": "2.0.0"
+  },
+  "peerDependencies": {
+    "react": "^16.13.1"
   }
 }
--- a/packages/loader/index.js
+++ b/packages/loader/index.js
@@ -1,10 +1,23 @@
 const {getOptions} = require('loader-utils')
 const mdx = require('@mdx-js/mdx')
+const path = require('path')
 
-const DEFAULT_RENDERER = `
-import React from 'react'
-import { mdx } from '@mdx-js/react'
+let defaultRenderer = null
+
+function getDefaultRenderer() {
+  if (!defaultRenderer) {
+    defaultRenderer = `
+import React from '${path.dirname(
+      require.resolve('react/package.json').replace(/\\/g, '/')
+    )}'
+import { mdx } from '${path
+      .dirname(require.resolve('@mdx-js/react/package.json'))
+      .replace(/\\/g, '/')}'
 `
+  }
+
+  return defaultRenderer
+}
 
 const loader = async function (content) {
   const callback = this.async()
@@ -20,7 +33,7 @@ const loader = async function (content) {
     return callback(err)
   }
 
-  const {renderer = DEFAULT_RENDERER} = options
+  const {renderer = getDefaultRenderer()} = options
 
   const code = `${renderer}\n${result}`
   return callback(null, code)
--- a/packages/parcel-plugin-mdx/src/MDXAsset.js
+++ b/packages/parcel-plugin-mdx/src/MDXAsset.js
@@ -1,6 +1,7 @@
 const {Asset} = require('parcel-bundler')
 
 const mdx = require('@mdx-js/mdx')
+const path = require('path')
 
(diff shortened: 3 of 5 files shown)


packages/loader/package.json
```diff
@@ -35,5 +35,8 @@
     "@mdx-js/mdx": "1.6.18",
     "@mdx-js/react": "1.6.18",
     "loader-utils": "2.0.0"
+  },
+  "peerDependencies": {
+    "react": "^16.13.1"
```
merceyz, line 40: Based on the usage I assume this is optional, is that correct?

ChristianMurphy:
@merceyz could you explain a bit more why this is needed?
The yarn team has opinions, but is there an actual issue that has been encountered?

merceyz:
> @merceyz could you explain a bit more why this is needed?

Sure and sorry, I assumed the detailed explainer on how hoisting works that I linked to would be enough. 

> The yarn team has opinions

The section of the docs I linked to explains hoisting in general, not Yarn, the same applies to npm and pnpm as well.

> is there an actual issue that has been encountered?

The issue is that the loaders are injecting imports into the users files. Based on the examples folder and the lack of a peerDependency it doesn't seem like the user is expected to declare `@mdx-js/react` and/or `@mdx-js/vue` as their own dependency. This means the loaders needs to resolve them before injecting them to ensure the correct version is used.

wooorm:
I don’t really see what you’re solving either: your code changes the require algorithm from runtime (before) to build time (proposed). I don’t see what that has to do with hoisting. At time of building, it’s fine to not have runtime libraries, and I think it’s weird to change the code based on what is available at build time?

> Based on the examples folder and the lack of a peerDependency it doesn't seem like the user is expected to declare @mdx-js/react and/or @mdx-js/vue as their own dependency.

I think this is expected, e.g., the [vue-loader example also depends on mdx-js/vue](https://github.com/mdx-js/mdx/blob/a4fb84aa824b66545eacc60c50beef52f7d848f7/packages/vue-loader/package.json#L40).

In fact, because the loaders are build time, I would argue that they shouldn’t have a direct or peer dependency on the runtime libraries.

merceyz:
Right, imagine for a moment that hoisting isn't a thing and I run this example https://github.com/mdx-js/mdx/blob/a4fb84aa824b66545eacc60c50beef52f7d848f7/examples/webpack/package.json

The layout on disk will look like this:
```
package.json
node_modules
  @mdx-js/loader
    node_modules
      @mdx-js/react
src
  hello.mdx
```

Now when I run webpack `@mdx-js/loader` will inject `import { mdx } from '@mdx-js/react'` into `src/hello.mdx` but since `@mdx-js/react` wasn't hoisted webpack can't find it and the build fails.

The changes in this PR solves this by making `@mdx-js/loader` inject `import { mdx } from 'node_modules/@mdx-js/loader/node_modules/@mdx-js/react'` which webpack can find and the build succeeds.

wooorm:
Thanks. I can see where you’re coming from. I still feel it’s not what’s expected. Maybe @johno can shed a light on this.

wooorm:
I’m sorry but I’m closing this. I think it adds more complexity than expected. And other maintainers are either impartial or unconvinced.

JSX also inserts say, `import { jsx as _jsx } from "react/jsx-runtime";`, not a relative path.

In your runtime, you can always install `@mdx-js/react` yourself, so that it’s available.

I guess it’s actually kind of a bug that `@mdx-js/react` is in the loader. More of a peer dep IMO.

PR #1307, opened 2020-10-20
Add annotation to use classic runtime

ykzts:
React v17 introduces a new syntax, JSX Runtime. MDX uses the traditional `React.createElement`, so if you mix it with the JSX Runtime, it will fail to build. This Pull Request adds an annotation that allows the classic runtime to be used, which is a temporary solution.

ref https://babeljs.io/docs/en/babel-plugin-transform-react-jsx


johno (commented):
Thanks!

PR #1331, opened 2020-11-13
Correct descriptions/tags in Table of components

mrm007:
Fixed tags for mdast `code`, `inlineCode` and `thematicBreak` types.

Preview here: https://mdx-git-patch-1.mdx.vercel.app/table-of-components

### DIFF
--- a/docs/getting-started/-table-of-components.mdx
+++ b/docs/getting-started/-table-of-components.mdx
@@ -16,12 +16,11 @@
 | `tbody`         | [Table body](https://github.com/syntax-tree/mdast#table)             |                                                     |
 | `tr`            | [Table row](https://github.com/syntax-tree/mdast#tablerow)           |                                                     |
 | `td`/`th`       | [Table cell](https://github.com/syntax-tree/mdast#tablecell)         |                                                     |
-| `pre`           | [Pre](https://github.com/syntax-tree/mdast#code)                     |                                                     |
-| `code`          | [Code](https://github.com/syntax-tree/mdast#code)                    | ` ```code``` `                                      |
+| `code`          | [InlineCode](https://github.com/syntax-tree/mdast#inlinecode)        | `` `inlineCode` ``                                  |
+| `pre`           | [Code](https://github.com/syntax-tree/mdast#code)                    | ` ```code``` `                                      |
 | `em`            | [Emphasis](https://github.com/syntax-tree/mdast#emphasis)            | `_emphasis_`                                        |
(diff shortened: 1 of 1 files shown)


wooorm:
Did you confirm that `code` is for inline code? I believe we have an `inlineCode` “element” for inline code: https://github.com/mdx-js/mdx/blob/889eb4b69b0aebdbd92afff4e16aac2db33e6c8b/packages/mdx/mdx-ast-to-mdx-hast.js#L13 (also on next branch)

mrm007:
@wooorm you're right, of course. Fixed that row in the table.

wooorm:
thanks!

I’ve just been doing some work in this repo, which gave me some new insights on this table. Sure MDX supports markdown, but the components in `MDXProvider` can be anything. If you have an `<Icon />` in MDX, you can have an `Icon` in components. Something like that might be good to mention here.
A different insight is that while the source is markdown-esque, and thus linking to mdast has some value, but it’s actually more HTML-like: `<code>` is also used because markdown code maps to `pre` *and* `code`.

So, for this PR: I think `code` has to be added back. And if you’d like to do more, although it could also be a new PR, a bigger refactor of this table, what it links to, and surrounding prose, might be helpful!

docs/getting-started/-table-of-components.mdx
```diff
@@ -17,12 +16,11 @@
 | `tbody`         | [Table body](https://github.com/syntax-tree/mdast#table)             |                                                     |
 | `tr`            | [Table row](https://github.com/syntax-tree/mdast#tablerow)           |                                                     |
 | `td`/`th`       | [Table cell](https://github.com/syntax-tree/mdast#tablecell)         |                                                     |
-| `pre`           | [Pre](https://github.com/syntax-tree/mdast#code)                     |                                                     |
-| `code`          | [Code](https://github.com/syntax-tree/mdast#code)                    | ` ```code``` `                                      |
+| `inlineCode`    | [InlineCode](https://github.com/syntax-tree/mdast#inlinecode)        | `` `inlineCode` ``                                  |
```
johno, line 19:
```suggestion
| `code`          | [Code](https://github.com/syntax-tree/mdast#code)                    | ` ```code``` `                                      |
| `inlineCode`    | [InlineCode](https://github.com/syntax-tree/mdast#inlinecode)        | `` `inlineCode` ``                                  |
```

johno:
Added back code, so going to go ahead and merge this. We should prolly create another issue to refactor this table in its entirety.

PR #1338, opened 2020-11-16
Fix Vue, test and document rest

wooorm:
/cc @codebender828 for the Vue part, @ChristopherBiscardi for the Preact part, any anyone else who wants to take a look (it’s a bit noisy due to the readme changes)

Hi folks! I added some more docs to readmes and added more tests. I couldn’t get the repo to work (latest yarn seems [broken](https://stackoverflow.com/questions/62254089/expected-workspace-package-to-exist-for-sane)), so I locked that. Most of the work was on the tests, trying to get to 100% coverage. I found a lot of tests that were theoretical (e.g., didn’t evaluate a loaded MDX file). We’re now evaluating more MDX code. Specifically, it seems Vue was (mostly) broken.

And: what’s the `withMDXComponents` in (p)react for? It was never tested, and is only used in an old example in the docs.

Changes

* Update readmes
* Fix code style
* Add 100% coverage to renderers, babel plugins, loaders, runtime
* Add todos inline for potentially broken things
* Fix Vue to have tests, support properties, support Vue components
* Lock Yarn because latest is broken
* Soft deprecate (by adding `private: true`) `@mdx-js/util`, `@mdx-js/test-util`, `babel-plugin-html-attributes-to-jsx`: their contents weren’t used or not reused
* Set up tests in each package, to allow for different test methods (e.g., vue)
* Largely not touching code (except for Vue only stylistic changes)


packages/babel-plugin-extract-export-names/index.js
```diff
@@ -38,16 +42,14 @@ class BabelPluginExtractExportNames {
       }
 
       const handleSpecifiers = node => {
-        const {specifiers} = node
-
-        if (!specifiers) {
-          return
-        }
+        /* istanbul ignore next - seems to always exist, maybe for older Babel? */
+        const specifiers = node.specifiers || []
```
ChristianMurphy, line 46: Since specifiers is the only part accessed would it make sense to use:

```js
const handleSpecifiers = ({ specifiers = [] }) =>
```
?

packages/babel-plugin-extract-export-names/test/index.test.js
```diff
@@ -2,55 +2,137 @@ const babel = require('@babel/core')
...
+  })
+
+  // To do: is this correct?
+  test('should capture renames', () => {
+    expect(
+      transform(
+        'var name1 = "a", name2 = "b"; export {name1 as foo, name2 as bar}'
+      )
+    ).toEqual(['name1', 'name2'])
+  })
```
ChristianMurphy, line 67: Good catch.
The names should be aliased as `foo` and `bar`

packages/babel-plugin-extract-export-names/test/index.test.js
```diff
@@ -2,55 +2,137 @@ const babel = require('@babel/core')
...
+
+  test('should capture object destructuring', () => {
+    expect(transform('export var {name1} = {}')).toEqual(['name1'])
+  })
+
+  // To do: is this correct?
+  test('should capture object destructuring w/ rename', () => {
+    expect(transform('export var {name1: bar} = {}')).toEqual(['name1'])
+  })
```
ChristianMurphy, line 76: :thinking: I'm not :100: sure on this one.
I think `bar` would be expected, but I'd need to double check the spec.

packages/babel-plugin-extract-export-names/test/index.test.js
```diff
@@ -2,55 +2,137 @@ const babel = require('@babel/core')
...
+
+  // To do: should be added.
+  test('should *not* capture a function as a default export', () => {
+    expect(transform('export default function name1() {}')).toEqual([])
+  })
+
+  // To do: should be added?
+  test('should *not* capture aggregates', () => {
+    expect(transform('export * from "..."')).toEqual([])
+  })
```
ChristianMurphy, line 106: This would be nice, but could be tricky, as it would require following the import (potentially more than once) to find what the value(s) are.

packages/babel-plugin-extract-export-names/test/index.test.js
```diff
@@ -2,55 +2,137 @@ const babel = require('@babel/core')
...
+  // To do: is this correct?
+  test('should capture reexported as a default whole', () => {
+    expect(transform('export {default} from "..."')).toEqual(['default'])
+  })
+
+  // To do: is this correct?
+  test('should capture reexported default and non-default', () => {
+    expect(transform('export {default as name1, name2} from "..."')).toEqual([
+      'default',
+      'name2'
```
ChristianMurphy, line 135: similar to above `as` aliases should ideally be respected.

packages/babel-plugin-html-attributes-to-jsx/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "private": true,
```
ChristianMurphy, line 2: why is this a private package?

packages/loader/index.js
```diff
@@ -3,10 +3,10 @@ const mdx = require('@mdx-js/mdx')
 
 const DEFAULT_RENDERER = `
 import React from 'react'
-import { mdx } from '@mdx-js/react'
+import {mdx} from '@mdx-js/react'
 `
 
-const loader = async function (content) {
+module.exports = async function (content) {
```
ChristianMurphy, line 9: It seems odd to mix ESM and CJS.
Would is make sense to have this be `export default`?
or alternatively to make the `import`s into `require`s?

packages/preact/test/test.js
```diff
@@ -1,107 +1,224 @@
...
+  })
+
+  test('should not crash if weird values could come from JSX', async () => {
+    // As JSX is function calls, that function can also be used directly in
+    // MDX. Definitely not a great idea, but it’s an easy way to pass in funky
+    // values.
+    const Content = await run('{mdx(1)}')
+
+    expect(render(<Content />)).toEqual('<1></1>')
+  })
```
ChristianMurphy, line 79: How far should this idea be taken?
Should imcomplete JSX crash or not crash? `<a>test</b>`
What about kinda, but not quite import statements? `import Something from './somewhere"`

ChristianMurphy (commented):
Thanks for getting this started @wooorm!

.yarnrc
```diff
@@ -1 +1,6 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
 ignore-engines true
+yarn-path ".yarn/releases/yarn-1.19.0.cjs"
```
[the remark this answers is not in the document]
wooorm: latest yarn can’t install this repo, locking to use an older version does work: https://stackoverflow.com/questions/62254089/expected-workspace-package-to-exist-for-sane

packages/babel-plugin-apply-mdx-type-props/index.js
```diff
@@ -1,6 +1,7 @@
 const {types: t} = require('@babel/core')
 const {declare} = require('@babel/helper-plugin-utils')
-const {startsWithCapitalLetter} = require('@mdx-js/util')
+
+const startsWithCapitalLetter = str => /^[A-Z]/.test(str)
```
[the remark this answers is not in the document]
wooorm: the utils weren’t used or reused, so I inlined them

packages/babel-plugin-extract-export-names/index.js
wooorm: I’d rather remove the default in a major, because I can’t seem to cover it

packages/babel-plugin-extract-export-names/test/index.test.js
wooorm: yeah, think it’s fine to not capture it

packages/babel-plugin-html-attributes-to-jsx/package.json
wooorm: because it’s no longer used. I propose actually deprecating them / removing them

packages/loader/index.js
wooorm: The ESM is for the generated MDX file, which can only be loaded in Babel as it currently includes JSX. The CJS is for the webpack loader.

Most packages in this project are written in CJS and do not require building / bundling, except for Vue and Preact

packages/preact/test/test.js
wooorm: The idea should not go anywhere: there was code handling invalid values, which can’t be passed by JSX, so I couldn’t test it. I didn’t want to remove the code because I may be missing something, and if it can be removed, I’d want to do it in a different PR

wooorm: And: mdx-js/preact, mdx-js/react, and mdx-js/vue are runtime projects, whereas broken JSX or import/exports are compile time problems (in mdx-js/mdx)

packages/vue/src/create-element.js
```diff
@@ -1,61 +1,71 @@
...
-        ...elProps,
-        ...data
-      }
-    }
+    children.unshift(props)
+    props = {}
+  }
+
+  children = children
+    .flatMap(d => (d == null ? [] : d))
```
ChristianMurphy, line 37: this seems a bit surprising.
According to the docs, `children` should be a single level array? (either: https://vuejs.org/v2/guide/render-function.html#createElement-Arguments or https://github.com/syntax-tree/hast#parent)
This implementation indicates a multilevel array.

.eslintignore
```diff
@@ -18,8 +18,7 @@ packages/gatsby-theme-mdx/src/components/search.js
 
 packages/remark-mdx
 
-# tmp
+# tmp - this doesn’t make sense at all
```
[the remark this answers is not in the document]
wooorm: The vue example shouldn’t be ignored from formatting 🤷‍♂️ wanted to do it in a more tooling-related pr

packages/vue/src/create-element.js
wooorm: it is: react-jsx and vue-jsx babel transforms work in mysterious ways. In my testing, I saw vue do `h(type, [undefined])`. I decided to handle both a children array (including if for some reason `null` or `undefined` are passed), but also support the more react-y way of `h(type, props, child, chilN)`

.yarnrc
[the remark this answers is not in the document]
ChristianMurphy: Isn't this what `engines` is for?
https://classic.yarnpkg.com/en/docs/package-json/#toc-engines

packages/vue/src/create-element.js
```diff
@@ -1,61 +1,71 @@
...
+    .map(d =>
+      typeof d === 'number' || typeof d === 'string'
+        ? this.createElement('d', {}, String(d)).children[0]
```
ChristianMurphy, line 40: this seems a bit quirky.
Why does it create an unused element and pick the first child of that element, rather than creating the element directly?
Would
```suggestion
        ? this.createElement(d)
```

work?

packages/vue/src/create-element.js
```diff
@@ -1,61 +1,71 @@
...
-
-    tag = DEFAULTS[type]
-    // Remove components object from attrs
-    const { components, ...attrs } = elProps.attrs
-    elProps = {
-      attrs
-    }
+const own = {}.hasOwnProperty
+
+export default function (type, props, ...children) {
```
ChristianMurphy, line 20: for tracing purposes would there be value in giving the function a name?
maybe:
```suggestion
export default function mdxCreateElement(type, props, ...children) {
```
?

packages/vue-loader/index.js
```diff
@@ -32,10 +19,28 @@ export default {
...
-    h = mdx.bind({ createElement, components: this.components })
-    return MDXContent({ components: this.components })
+    h = mdx.bind({createElement, components: this.components})
+    return MDXContent({components: this.components})
   }
 }
-   `
+`
+
+module.exports = async function (content) {
```
ChristianMurphy, line 28: for debug purposes it could be good to name the exported function

```suggestion
module.exports = async function mdxVueLoader(content) {
```

.yarnrc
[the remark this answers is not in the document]
wooorm: `engines` is for the published package: where it runs, and throws an error or warns if you’re on something else (and this config file seems to ignore it haha)

This is internally: that using yarn on this repo will use a specific locked version for now.
Anyway: it’s just a quick solution for this issue here: https://github.com/yarnpkg/yarn/issues/7807

packages/vue/src/create-element.js
wooorm: It is a bit quirky! The reason is to get a VNode for the text, not for an element

I assumed the above would work the best. It seems there is private function `this._v(d)` we could also use, but I’d rather stay away from hidden stuff.

The reason that we must make them into nodes, is because we’re using `toH` here to handle difference in JSX attribute casing and the complexity of where Vue expects props: https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth.
But, `toH` deals with hast children but here they are already compiled to vue nodes except for the texts

packages/vue/src/create-element.js
wooorm: Good idea! Went w/ `createMdxElement`.

A different issue, but: right now we’re calling everything mdx. Would be good to use `m` / `createMdxElement` for the renderers, and `mdx` for the core which generates serialized JS

ChristianMurphy:
@wooorm have you had a chance to look into why CI is failing?
It looks like several are failing with
```
Error: Cannot find module '/home/runner/work/mdx/mdx/.yarn/releases/yarn-1.19.0.cjs'
```
sounds related to https://github.com/mdx-js/mdx/pull/1338#discussion_r524328805

.eslintignore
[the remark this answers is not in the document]
ChristianMurphy: I interpreted this as a commented out folder
```suggestion
```
but either way this could probably be removed.

.yarnrc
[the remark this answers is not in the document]
ChristianMurphy: This seems to be causing issues on CI

ChristianMurphy (changes requested):

wooorm:
Well, actions is a bit slow but it’s somewhat working.

@ChristianMurphy this error might be something for you: https://github.com/mdx-js/mdx/pull/1338/checks?check_run_id=1418253147#step:5:135

And if someone knows why `lerna run test` or so, while we have `"npmClient": "yarn"` in `lerna.json`, yields:

```
npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1605705373202-0.3365631855975586/node but npm is using /opt/hostedtoolcache/node/10.23.0/x64/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

I’m all ears 🤷‍♂️

.eslintrc.yml
```diff
@@ -43,15 +43,13 @@ settings:
...
-  - files: '*.md'
-    extends: plugin:mdx/recommended
-
   - files: '*.mdx'
     extends:
       - plugin:mdx/recommended
       - plugin:mdx/overrides
     rules:
       import/no-extraneous-dependencies: 0
+      prettier/prettier: 0
```
ChristianMurphy, line 52: :raised_eyebrow: I though prettier supported MDX, or does the eslint-mdx-parser interfere with that?

.github/workflows/lint.yml
```diff
@@ -1,27 +0,0 @@
-name: CI
```
ChristianMurphy, line 1: this was removed because `lint` is now a part of `test` is that accurate?

package.json
```diff
@@ -13,102 +13,53 @@
     "packages/*"
   ],
   "engines": {
-    "node": "12.x"
+    "node": ">=12"
```
ChristianMurphy, line 16: How does this, fit in with CI? https://github.com/mdx-js/mdx/blob/9443e235679f7d37cdfa436ef1b167e4bb68615d/.github/workflows/ci.yml#L14

Would it make sense to use a version union?

```suggestion
    "node": "^10 || ^12 || ^14"
```

packages/loader/index.js
ChristianMurphy: Ah, I missed the template string :sweat_smile: 
My bad, thanks for clarifying.

.eslintrc.yml
wooorm: Something with newlines that prettier says are missing, but are clearly there. Assuming prettier + eslint-mdx + the beta packages in this project don’t like eachother

.github/workflows/lint.yml
wooorm: Yeah, no need to have three separate work flows IMO, because most of the setup is in installing.
And: `lint` requires `build` too (because of microbundle bundling vue, react, and preact now)

package.json
wooorm: I don’t think it does anything because it’s ignored: https://github.com/mdx-js/mdx/blob/e194fc1a61715549be10d56055ea4ad25533fcb1/.yarnrc#L1. And it’s in the main package.json which isn’t published.

I thought unlocking it at least would make more sense that locking to an old version, but would like to revisit it later (IMO remove it)

ChristianMurphy:
> ```
> npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1605705373202-0.3365631855975586/node but npm is using /opt/hostedtoolcache/node/10.23.0/x64/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
> ```

Not sure, and not too worried, since it's a warning not an error.
Yarn has a mix of actual warnings and yarn opinions logged as warnings, it can be hard to tell them apart, my guess is this is an opinion as a warning.
Ideally I'd like to see the project move to `npm` 7, and drop `yarn` entirely, but that's another discussion for another PR.

---

More concerning is:

```
Error: Cannot find module '/home/runner/.dts/typescript-installs/4.0/node_modules/typescript'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at testDependencies (/home/runner/work/mdx/mdx/node_modules/dtslint/bin/lint.js:64:16)
    at Object.<anonymous> (/home/runner/work/mdx/mdx/node_modules/dtslint/bin/lint.js:26:28)
    at Generator.next (<anonymous>)
    at /home/runner/work/mdx/mdx/node_modules/dtslint/bin/lint.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/home/runner/work/mdx/mdx/node_modules/dtslint/bin/lint.js:4:12
```

This might be a network connectivity issue, or it may be an issue with the cache :thinking:

wooorm:
A rerun showed different network timeouts. Maybe what we’re installing is just too much for GH to handle 😅

packages/babel-plugin-extract-export-names/test/index.test.js
wooorm: Confirmed that bar should be caught:

```js
// a.mjs
export var {name1: bar} = {}

// b.mjs
import * as all from './a.mjs'
console.log(all)
```

Yields:

```
[Module] { bar: undefined }
```

packages/babel-plugin-extract-export-names/test/index.test.js
wooorm: I guess this package is used more for figuring out which local values are also exported (not caring about as what they’re exported).

https://github.com/mdx-js/mdx/blob/e194fc1a61715549be10d56055ea4ad25533fcb1/packages/mdx/mdx-hast-to-jsx.js#L153-L163

I’m not 100% whether doing it like this, and at build time, is useful? I’d say checking if this was nully at runtime would be better?

https://github.com/mdx-js/mdx/blob/9443e235679f7d37cdfa436ef1b167e4bb68615d/packages/react/src/create-element.js#L23-L27

wooorm: I think I’ll just leave {import,export}-names and see when I start on mdx-js/mdx whether they can be deprecated

ChristianMurphy:
Thanks @wooorm, looks like github actions is getting happier.
The Vercel build seems to be running into a new issue:

```

06:14:49.259 | Generating JavaScript bundles failed
-- | --
06:14:49.260 | Cannot find module '@mdx-js/mdx/util'
06:14:49.260 | Require stack:
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby-plugin-mdx/loaders/mdx-loader.js
06:14:49.260 | - /vercel/workpath0/node_modules/loader-runner/lib/loadLoader.js
06:14:49.260 | - /vercel/workpath0/node_modules/loader-runner/lib/LoaderRunner.js
06:14:49.260 | - /vercel/workpath0/node_modules/webpack/lib/NormalModule.js
06:14:49.260 | - /vercel/workpath0/node_modules/webpack/lib/NormalModuleFactory.js
06:14:49.260 | - /vercel/workpath0/node_modules/webpack/lib/Compiler.js
06:14:49.260 | - /vercel/workpath0/node_modules/webpack/lib/webpack.js
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby/dist/commands/build-html.js
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby/dist/commands/build.js
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby-cli/lib/create-cli.js
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby-cli/lib/index.js
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby/dist/bin/gatsby.js
06:14:49.260 | - /vercel/workpath0/node_modules/gatsby/cli.js
06:14:49.260 | File: packages/gatsby-theme-mdx/src/pages/playground.mdx
```
https://vercel.com/mdx/mdx/lvagnsynz

wooorm:
I’ll see what I can do. I don‘t know Gatsby. FWIW the tests have been failing for a while now https://github.com/mdx-js/mdx/commits/next, I’d rather start moving to smaller PRs

ChristianMurphy (approved):

PR #1343, opened 2020-11-18
Fix `delete` -> `del` in docs on typescript

freeshineit:

### DIFF
--- a/docs/advanced/typescript.mdx
+++ b/docs/advanced/typescript.mdx
@@ -15,7 +15,7 @@ declare module '@mdx-js/react' {
     | 'a'
     | 'blockquote'
     | 'code'
-    | 'delete'
+    | 'del'
     | 'em'
     | 'h1'
     | 'h2'


wooorm:
Duplicate of #1331

Thanks for taking the time to contribute!

We noticed that this is a duplicate of GH-1331. You may want to subscribe there for updates.

Because we treat our issues list as [the project team’s backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog), we close duplicates to focus our work and not have to touch the same chunk of code for the same reason multiple times. This is also why we may mark something as duplicate that isn’t an exact duplicate but is closely related.

wooorm:
Whoops, sorry: I see that this is somewhat different!

wooorm (approved):

ChristianMurphy (approved):

wooorm:
Thanks @freeshineit!

PR #1344, opened 2020-11-18
fix(docs): correct typos in Vue docs

emmiep:
Fixes two typos in the Vue docs:

- The regexp for matching `.mdx` files is missing a backslash escaping the dot.
- Webpack is configured using `vue.config.js` rather than `webpack.config.js` in Vue CLI (https://cli.vuejs.org/guide/webpack.html)

### DIFF
--- a/docs/guides/vue.mdx
+++ b/docs/guides/vue.mdx
@@ -18,7 +18,7 @@ yarn add @mdx-js/vue-loader
 
 ## Configuring webpack
 
-In a [Vue CLI](https://cli.vuejs.org/) you need to add a `webpack.config.js` that
+In a [Vue CLI](https://cli.vuejs.org/) you need to add a `vue.config.js` that
 uses the MDX loader for vue:
 
 ```js
--- a/docs/guides/vue.mdx
+++ b/docs/guides/vue.mdx
@@ -27,7 +27,7 @@ module.exports = {
     module: {
       rules: [
         {
-          test: /.mdx?$/,
+          test: /\.mdx?$/,
           use: ['babel-loader', '@mdx-js/vue-loader']
         }
       ]
--- a/docs/guides/vue.mdx
+++ b/docs/guides/vue.mdx
@@ -27,7 +27,7 @@ module.exports = {
     module: {
       rules: [
         {
-          test: /\.mdx?$/,
+          test: /\.mdx$/,
           use: ['babel-loader', '@mdx-js/vue-loader']
         }
       ]


docs/guides/vue.mdx
```diff
@@ -18,7 +18,7 @@ yarn add @mdx-js/vue-loader
 
 ## Configuring webpack
 
-In a [Vue CLI](https://cli.vuejs.org/) you need to add a `webpack.config.js` that
+In a [Vue CLI](https://cli.vuejs.org/) you need to add a `vue.config.js` that
```
ChristianMurphy, line 21: This is accurate as is.
It is a webpack loader, so it would go in the webpack config, there currently is not a dedicated vue cli plugin.

ChristianMurphy (changes requested):

docs/guides/vue.mdx
emmiep: I'm sorry I'm not quite following you.
If you're making a "custom" project from scratch using webpack it would definitely go in `webpack.config.js`, but as far as I can tell Vue CLI creates the final webpack configuration for you by merging its configuration with your custom configuration in `vue.config.js`.

The Vue CLI documentation for `vue.config.js` says [`configureWebpack`](https://cli.vuejs.org/config/#configurewebpack) is used to merge your configuration:

> If the value is an Object, it will be merged into the final config using webpack-merge.

As far as I can tell there is no `configureWebpack` in `webpack.config.js`, so the mdx documentation stating this is `vue.config.js` made me confused.

Just to be sure I created [two Vue projects using either `vue.config.js` or `webpack.config.js`](https://github.com/emmiep/mdx-vue-issue-demo) to see if both works, but when using `webpack.config.js` the build fails with this error:

```
 ERROR  Failed to compile with 1 errors                                                  4:24:42 PM

 error  in ./src/components/HelloMDX.mdx

Module parse failed: Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> # Hello MDX!
| 
| Hello!

 @ ./node_modules/cache-loader/dist/cjs.js??ref--12-0!./node_modules/babel-loader/lib!./node_modules/cache-loader/dist/cjs.js??ref--0-0!./node_modules/vue-loader-v16/dist??ref--0-1!./src/App.vue?vue&type=script&lang=js 1:0-49 5:14-22
 @ ./src/App.vue?vue&type=script&lang=js
 @ ./src/App.vue
 @ ./src/main.js
 @ multi (webpack)-dev-server/client?http://192.168.1.144:8080&sockPath=/sockjs-node (webpack)/hot/dev-server.js ./src/main.js
```

I tried moving the configuration out of the `configureWebpack` property, but with no different result.
I would guess the `vue-cli-service` would simply ignore any webpack configuration files and create its own configuration by merging its own base configuration with your custom one from `vue.config.js`.

ChristianMurphy: @emmiep not all Vue users use Vue CLI.
It may make sense to add the reference on how to use with Vue CLI since it is common.
But it does not make sense to remove the more generic and webpack project documentation.

emmiep: Oh, I forgot to mention as it is now the documentation literally states:

> In a Vue CLI you need to add a webpack.config.js that uses the MDX loader for vue

And I agree it would probably be good to mention how to achieve it both with webpack both with and without Vue CLI, but as it is now it's not correct for Vue CLI, which the documentation states.

wooorm: Is this something that should be changed in this PR? Or can it be dismissed?

ChristianMurphy: It would be nice to have a webpack guide, but would could merge this as is.

docs/guides/vue.mdx
```diff
@@ -27,7 +27,7 @@ module.exports = {
     module: {
       rules: [
         {
-          test: /.mdx?$/,
+          test: /\.mdx?$/,
```
wooorm, line 30:
```suggestion
          test: /\.mdx$/,
```

wooorm (approved):

wooorm:
thanks for your patience @emmiep, and the review @ChristianMurphy!

PR #1346, opened 2020-11-18
Add annotation to use classic runtime

michrome:
This is a copy of #1307 but for MDX@2. It's to avoid the error `pragma and pragmaFrag cannot be set when runtime is automatic` which is a result of [React's new JSX Transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).

Related discussion at #1345 .

### DIFF
--- a/packages/mdx/index.js
+++ b/packages/mdx/index.js
@@ -92,7 +92,8 @@ function sync(mdx, options = {}) {
 
   const {contents} = compiler.processSync(fileOpts)
 
-  return `/* @jsx mdx */
+  return `/* @jsxRuntime classic */
+/* @jsx mdx */
 ${contents}`
 }
 
@@ -106,7 +107,8 @@ async function compile(mdx, options = {}) {
 
   const {contents} = await compiler.process(fileOpts)
 
-  return `/* @jsx mdx */
+  return `/* @jsxRuntime classic */
+/* @jsx mdx */
 ${contents}`
 }
 


ChristianMurphy (approved):
Thanks @brandonweiss !

ChristianMurphy (commented):
@brandonweiss when you have a chance could you look into updating the test cases for the generated code https://github.com/mdx-js/mdx/pull/1346/checks?check_run_id=1421073559#step:9:70
to include the new comment/pragma?

wooorm:
Yes, would be good to get that updated fixture in there.
If you’re having troubles with the repo that might be addressed in #1338, in which case I’m fine landing your PR first and I’ll fix it later.

codejunkienick:
Is it possible to publish this any time soon? It will help with projects which use MDX and Emotion, in my case Storybook

wooorm:
@michrome Could you address the conflict, and Christian’s feedback (tests changed a bunch)?

@codejunkienick This PR is currently not mergeable as is.

wooorm:
I cherry-picked all master (v1) changes and applied them on next (v2), which is now the `main` and default branch, so this is no longer needed, once it’s released. Thanks all!

shuding:
Hey @wooorm! Currently MDX v2 is breaking when using React 17 because of this issue. Is it possible to release a new v2 version any time soon?

michrome:
FYI @shuding https://github.com/mdx-js/mdx/commit/b1b2cde7915ae9848d71fe2c22021d679998af67

PR #1347, opened 2020-11-21
Refactor `@mdx-js/mdx` to clean code, test 100%

wooorm:
### What this is

A two part refactor of the core `@mdx-js/mdx` package:
- Rewrite core to remove everything that is untested and undocumented; refactor it a bit; and only real change: remove footnotes (folks can add it themselves if they want)
- Rewrite the tests w/o touching the code to reach 100% test coverage

The rest of the commit bodies for both commits explains in detail what changed and why.

### Why it should land

- The core code and its tests has been touched by many hands over the years, so making it somewhat more coherent is good for the future
- The tests are very focussed on the result of MDX, not what it actually does. I plan on working to make the core output plain vanilla JS instead of JSX, and having the tests check what the MDX format does instead of what it compiles down to, will make that next step easier.

### What to review

Mostly the changes to the actual code are of interest to the folks who’ve worked with mdx-js/mdx and its hidden features to build other things: maybe some things should still be allowed, in which case I can add ’em back with tests and docs


packages/mdx/mdx-ast-to-mdx-hast.js
```diff
@@ -1,102 +1,80 @@
...
+          // To do: this handling seems not what users expect:
+          // <https://github.com/mdx-js/mdx/issues/702>.
+          const meta =
+            node.meta &&
+            node.meta.split(' ').reduce((acc, cur) => {
+              if (cur.split('=').length > 1) {
+                const t = cur.split('=')
+                acc[t[0]] = t[1]
+                return acc
+              }
```
ChristianMurphy, line 35: I'm a bit surprised there is a custom meta handler here.
Since there is so much variation in what user's want/expect, it may make sense to split it out into a plugin?

ChristianMurphy (approved):

packages/mdx/mdx-ast-to-mdx-hast.js
wooorm: All for it. I added a comment above this to link to an issue (and there's another one open as well). This current behavior has some value, but the implementation is not what folks seem to expect. I left it as is to not break anything, but would like to look into it again

PR #1349, opened 2020-11-21
Fix tests on next

wooorm:
* Remove typescript definition tests, they don’t play well with our `yarn.lock`, and are turned off on main too
* “Fix” Gatsby. The site is incredibly purple, but at least it doesn’t fail. I don’t know Gatsby, CSS-in-JS, or theme-ui, so I have no clue how to actually fix it.

### DIFF
--- a/packages/mdx/package.json
+++ b/packages/mdx/package.json
@@ -35,9 +35,10 @@
     "mdxast"
   ],
   "scripts": {
-    "test-api": "jest test --coverage",
+    "test-api": "jest test",
+    "test-coverage": "jest test --coverage",
     "test-types": "dtslint types",
-    "test": "npm run test-api && npm run test-types"
+    "test": "yarn test-coverage && yarn test-types"
   },
   "dependencies": {
     "@babel/core": "7.10.5",
--- a/packages/mdx/package.json
+++ b/packages/mdx/package.json
@@ -22,6 +22,7 @@
   "types": "types/index.d.ts",
   "files": [
     "index.js",
+    "util.js",
     "mdx-ast-to-mdx-hast.js",
     "mdx-hast-to-jsx.js",
     "types/index.d.ts"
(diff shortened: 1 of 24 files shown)


johno (approved):

packages/mdx/package.json
```diff
@@ -35,14 +36,16 @@
...
+    "test-coverage": "jest test --coverage",
     "test-types": "dtslint types",
-    "test": "npm run test-api && npm run test-types"
+    "test": "yarn test-coverage && yarn test-types"
   },
   "dependencies": {
     "@babel/core": "7.10.5",
     "@babel/plugin-syntax-jsx": "7.10.4",
     "@babel/plugin-syntax-object-rest-spread": "7.8.3",
+    "@mdx-js/util": "2.0.0-next.1",
```
ChristianMurphy, line 48: Is this pinned to next 1 for a reason? `2.0.0-next.8` is the latest published in this release line.

wooorm: Somewhat complex to explain the background:
I’m not 100% sure which, but some gatsby-mdx package uses `@mdx-js/mdx/util`, which used to be where the utilities were, which were later moved to `@mdx-js/util`. I’ve deprecated that package. But as we’re using a linked package, gatsby was failing because it was missing.

I added a warning to `@mdx-js/mdx/util` and propose actually removing it (related to https://github.com/mdx-js/mdx/pull/1338), but can’t for now.
The reason for using a lower version was to ensure that lerna doesn’t update it on publish, maybe there’s another way around it.

ChristianMurphy: :raised_eyebrow: won't lerna see an old version and update it anyway?

ChristianMurphy (approved):

packages/mdx/package.json
wooorm: The v1 versions of a couple packages were used in the main `package.json`, and didn’t get updated apparently. Might have to do with only majors, I dunno 🤷‍♂️

PR #1350, opened 2020-11-22
Refactor and test `remark-mdxjs`

wooorm:
A tiny refactor but mostly just removing tests that were testing other packages, and adding more tests for imports/exports


ChristianMurphy:
It looks like this is getting an error similar to https://github.com/mdx-js/mdx/pull/1178 resolved?
Maybe a rebase/reapplication of the changes will resolve CI's issue?

wooorm:
Yeah, once I land my PR fixing the tests, my other two PRs (including this one) should be passing!

packages/remark-mdxjs/test/__snapshots__/test.js.snap
```diff
@@ -1,51 +1,88 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`correctly transpiles 1`] = `
-"import Foo from './bar'
+exports[`Handles an import that is later exported 1`] = `
```
ChristianMurphy, line 3: general question, probably don't require changes.
In some places you opt for tests in others snapshots.
What is the thinking about which should be used where?

ChristianMurphy (approved):

packages/remark-mdxjs/test/__snapshots__/test.js.snap
wooorm: Snapshots were originally there. I can see some use in them, but I personally generally prefer them inline and just as a normal test.

Issue #1341, opened 2020-11-18
`remark-mdx` does not work with `remark@13`


MatthewCaseres:
### Subject of the issue

I am unable to replicate the "Transform Content section of the documentation. https://mdxjs.com/advanced/transform-content

### Your environment

*   Windows 10
*   see package.json in repo
*   node v12.18.2

### Steps to reproduce

https://github.com/MatthewCaseres/mdx-reproduction

### Expected behaviour

I was trying to follow the example here: https://mdxjs.com/advanced/transform-content. I was expecting there to exist a leaf of type import as is shown in the docs

### Actual behaviour

![image](https://user-images.githubusercontent.com/43053796/99467840-d54af700-2904-11eb-85e1-273d3d9c9571.png)

wooorm:
Thanks for the reproduction, @MatthewCaseres!

remark@13 does not work with remark-mdx yet. You can either downgrade to use remark@12 for now. Or you can use `@mdx-js/mdx` directly, which includes `remark-mdx` and supports passing `remarkPlugins`.
What to do, depends on what you want to do!

MatthewCaseres:
Thanks! I'm trying to build a plugin and it seems like I can iterate and check my work faster with a node script. In my current nextjs project I have to stop and restart the server to rebuild the MDXAST and I can't output the transformed stuff to a file for inspection.

I might only need the markdown nodes (I'm modifying headers), I was just surprised I couldn't reproduce the docs. Thanks for the speedy reply!

wooorm:
A node script can also use `mdx-js/mdx`!

wooorm:
Hi all! I’m going to close this as it landed on the `main`, the (now default) branch we’re using to gear up to the MDX@2 release.

The reason is so that it’s more clear which issues still need to be solved already and don’t require further work.

Expect another `next` release soonish, which might include it, or maybe it’s already released as a beta!

For more info, see <https://github.com/mdx-js/mdx/issues/1041>.

Issue #1318, opened 2020-10-31
Nested JSX with hanging > 


ChristopherBiscardi:
### Subject of the issue

Parse error. Something to do with the way children are wrapped inside of JSX I think because the issue only appears with content above the child and also a hanging >

### Your environment

*   **Packages**: `@mdx-js/mdx@2.0.0-next.8`

### Steps to reproduce

```
<Box>
  Original:
  <Box
    as="a"
    href="https://stackoverflow.com/questions/28732670/getting-recently-updated-packages/29104122#29104122"
  >
    s
  </Box>
</Box>
```

repro script for the above file:

```
const mdx = require("@mdx-js/mdx");
const fs = require("fs").promises;

async function process() {
  try {
    const file = await fs.readFile("./thing.mdx", "utf-8");
    const result = await mdx(file);
    console.log(result);
  } catch (e) {
    console.log(e);
  }
}
process();
```

### Expected behaviour

dunno, seems weird that this breaks on parsing.

### Actual behaviour

```
[5:108: Unexpected end of file before attribute name, expected a character that can start an attribute name, such as a letter, `$`, or `_`; whitespace before attributes; or the end of the tag] {
  reason: 'Unexpected end of file before attribute name, expected a character that can start an attribute name, such as a letter, `$`, or `_`; whitespace before attributes; or the end of the tag',
  line: 5,
  column: 108,
  location: {
    start: { line: 5, column: 108, offset: 143, index: 123 },
    end: { line: null, column: null }
  },
  source: 'remark-mdx',
  ruleId: 'unexpected-eof',
  fatal: true
}
```

wooorm:
I'm not 100% why it's parsing as a span instead of a block (early in the am for me), but the > is making a block quote: https://github.com/micromark/mdx-state-machine#742-u003e-greater-than--seen-as-block-quote

wooorm:
I just confirmed that this now renders nicely on `main`. Assuming the shortcodes aren’t passed, the result is:

```html
<div><p>Original:</p><div as="a" href="https://stackoverflow.com/questions/28732670/getting-recently-updated-packages/29104122#29104122"><p>s</p></div></div>
```

Which seems to make sense.

Issue #1316, opened 2020-10-29
Two blank lines within a fenced code block terminates ordered list and breaks code block


epbarger:
### Subject of the issue
Fenced code blocks within a number bullet point break if they contain more one new-line in a row.

### Your environment

*   **OS**: macOS Mojave 10.14.4
*   **Packages**: `@mdx-js/mdx@1.6.4`, via `"gatsby-plugin-mdx": "^1.2.6"`
*   **Env**: yarn 1.22.4, gatsby 2.23.1

### Steps to reproduce
Nest a fenced code block under a number bullet point, with two blank lines in a row within the code block. One blank line does not cause this issue.

````markdown
1. Test
   ```
   broken


   example
   ```

Plain text after the numbered bullet point
````

### Expected behaviour

I would expect this to work fine, parsing the code block correctly and keeping the entire block nested under the numbered bullet point.

### Actual behaviour

The second new line appears to cause the parser to end the bullet point, breaking the code block. The ending triple backticks will then start a new code block, causing chaos with the rest of the page

![Screen Shot 2020-10-29 at 3 41 01 PM](https://user-images.githubusercontent.com/2780987/97624236-374ac780-19fd-11eb-8604-4981c5f6aac3.png)

wooorm:
@epbarger Could you check if this is solved in MDX@2? (see the open issues about that for more info). That’s a big rewrite that might solve this. And there’s another change coming to that’d definitely solve this, by using the new micromark parser in remark

wooorm:
Hi all! I’m going to close this as it landed on the `main`, the (now default) branch we’re using to gear up to the MDX@2 release.

The reason is so that it’s more clear which issues still need to be solved already and don’t require further work.

Expect another `next` release soonish, which might include it, or maybe it’s already released as a beta!

For more info, see <https://github.com/mdx-js/mdx/issues/1041>.

Issue #1283, opened 2020-09-29
[v2] Fenced code blocks within JSX elements with indent > 2 parses incorrectly


jerelmiller:
I've been playing around a bit more with MDX v2 and encountered what I think is a bug, or at the very least, some surprising behavior. I have the need to render fenced code blocks within table cells. Because of the highly nested structure of a table, and to maintain sane indentation, I've placed the code block indented within the `<td />`. Unfortunately MDX v2 parses this incorrectly.

### Steps to reproduce

For example, here is a simple table with a single cell that outputs a code block in one of the `td`s:

````mdx
<table>
  <tbody>
    <tr>
      <td>
        ```
        console.log(test)
        ```
      </td>
      <td>
        Some more text
      </td>
    </tr>
  </tbody>
</table>
````

This parses into an AST like the following:

```json
{
  "type": "code",
  "lang": null,
  "meta": null,
  "value": "console.log(test)\n```\n",
  "position": {
    "start": {
      "line": 5,
      "column": 1,
      "offset": 38
    },
    "end": {
      "line": 8,
      "column": 7,
      "offset": 94
    },
    "indent": [
      1,
      1,
      1
    ]
  }
}
```

Notice how the `value` ends with the fence literal and a newline. This results in text that renders to the screen as:

````
console.log(test)
```
````

If I remove the indentation in front of the code block, it seems to parse and output the correct text:

**Markdown**
````mdx
<table>
  <tbody>
    <tr>
      <td>
```
console.log(test)
```
      </td>
      <td>
        Some more text
      </td>
    </tr>
  </tbody>
</table>
````

**AST**
```json
{
  "type": "code",
  "lang": null,
  "meta": null,
  "value": "console.log(test)",
  "position": {
    "start": {
      "line": 5,
      "column": 1,
      "offset": 38
    },
    "end": {
      "line": 7,
      "column": 4,
      "offset": 63
    },
    "indent": [
      1,
      1
    ]
  }
}
```

**Output**
```
console.log(test)
```

You can see the result of each of these using the [playground](https://mdx-git-next.mdx.vercel.app/playground/)

### Expected behaviour

I would expect that regardless of the indentation, the code block's `value` would result in the same thing.

Is this intentional behavior or a bug? I find it a bit surprising at the least.

### Minimal fix

I've been able to "fix" this issue by creating my own MDX remark transform:

```js
const visit = require('unist-util-visit');

[...]

PR #1367, opened 2020-12-11
Update to remark@13 (micromark)

wooorm:
This updates MDX to use and support remark@13, which comes with a new internal parser (micromark), and supports CommonMark. See <https://github.com/remarkjs/remark/releases/tag/13.0.0> for more information. In short, this means MDX parses markdown closer to what folks expect. And it means all latest plugins work (again).

But it also means that parsing MDX syntax (JSX, expressions, ESM) got an update. See: <https://github.com/micromark/micromark-extension-mdxjs> and <https://github.com/syntax-tree/mdast-util-mdx> for the syntax. In short, this means MDX parsing is now JavaScript-aware: import/exports are actually parsed for being valid JavaScript. Expressions previously counted braces, but now can include braces in strings or comments or whatnot. This also means we can drop Babel (in a future PR) because we already have a JavaScript AST.

This also deprecates the packages `remark-mdxjs` (which is now the default in `remark-mdx`), `remark-mdx-remove-exports`, and `remark-mdx-remove-imports`.

Related to GH-704.
Related to GH-1041.

Closes GH-720.
Closes GH-1028.
Closes GH-1050.
Closes GH-1081.
Closes GH-1193.
Closes GH-1238.
Closes GH-1283.
Closes GH-1316.
Closes GH-1318.
Closes GH-1341.


packages/remark-mdx/test/fixtures/block-parsing.md
```diff
@@ -41,8 +41,8 @@ Weeeeee!
...
-<Foo>{
-  const num = 1+1
-
-  return <h2>{num}</h2>
-}</Foo>
+<Foo>
+  {
+    <h2>{1+1}</h2>
+  }
+</Foo>
```
wooorm, line 48: This was never one expression: JSX only allows expressions, not whole programs.
And I couldn’t get `<Tag>{` to play nice with micromark, hence the line between.

.remarkrc.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
   plugins: [
     './packages/remark-mdx',
-    './packages/remark-mdxjs',
```
wooorm, line 4: deprecated (now in remark-mdx per john’s suggestion)

docs/advanced/typescript.mdx
```diff
@@ -15,8 +15,6 @@ improve the developer experience for TypeScript users.
 - `@mdx-js/runtime`
 - `@mdx-js/vue`
 - `remark-mdx`
-- `remark-mdx-remove-exports`
-- `remark-mdx-remove-imports`
```
wooorm, line 19: deprecated (ESM is just one node now: we could create a new package but it’s also a one-liner that folks can do manually)

packages/mdx/mdx-ast-to-mdx-hast.js
```diff
@@ -53,25 +53,20 @@ function mdxAstToMdxHast() {
...
         },
-        mdxBlockExpression(h, node) {
-          return Object.assign({}, node, {type: 'mdxBlockExpression'})
+        mdxFlowExpression(h, node) {
+          return node
         },
-        mdxSpanExpression(h, node) {
-          return Object.assign({}, node, {type: 'mdxSpanExpression'})
+        mdxTextExpression(h, node) {
+          return node
```
wooorm, line 69: different node types, mostly ESM and expressions can be passed right through

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -1,12 +1,17 @@
 const {transformSync} = require('@babel/core')
 const uniq = require('lodash.uniq')
-const {serializeTags} = require('remark-mdx/lib/serialize/mdx-element')
-const serializeMdxExpression = require('remark-mdx/lib/serialize/mdx-expression')
+const encode = require('stringify-entities/light')
```
wooorm, line 3: The changes in this file are to make micromark and the estrees work, but otherwise not produce noise by keeping stuff as close as it already was. A future PR will compile compile to JS file w/o JSX.

packages/mdx/test/index.test.js
```diff
@@ -257,7 +258,9 @@ describe('@mdx-js/mdx', () => {
   })
 
   it('should support tables in markdown', async () => {
-    const Content = await run('| A | B |\n| :- | -: |\n| a | b |')
+    const Content = await run('| A | B |\n| :- | -: |\n| a | b |', {
+      remarkPlugins: [gfm]
+    })
```
wooorm, line 263: micromark/remark is commonmark compliant, and GFM (strikethrough, tables, literal URLs, tasklists) are enabled by using a plugin

packages/remark-mdx/test/fixtures/block-parsing.out
```diff
@@ -4,9 +4,9 @@ From [GitHub](https://github.com/mdx-js/mdx/issues/195#issuecomment-506745859).
 
 <Row>
   <Col>
-    -   some md
-    -   asd
-    -   asd
+    *   some md
+    *   asd
+    *   asd
```
wooorm, line 9: `remark-stringify` has some new output defaults, which are producing some noise here but are otherwise syntactically the same

packages/remark-mdx/test/fixtures/block-vs-inline.md
```diff
@@ -3,25 +3,22 @@
 To make this example more readable, uppercase letters are used when JSX is
 parsed as a block, wheres lowercase letters are used for inline JSX.
 
-This is inline, it’s a paragraph, because `</x>` is followed by `.`.
+This is inline, it’s a paragraph, because `<x>` is followed by `_`.
```
wooorm, line 6: I changed my mind here. I no longer think `<x># is this a heading?</x>` should be a heading. If you want blocks in tags, use blank lines.

packages/remark-mdx/test/fixtures/character-reference.out
```diff
@@ -4,15 +4,13 @@ The characters `>` and `}` don’t need to be handled differently: >, }.
 
 Those characters are also fine in JSX elements: <>> and }</>.
 
-{
-  /* starts an expression, it must be closed: */
-}
+{/* starts an expression, it must be closed: */}
 
-If you don’t want an expression, encode it, like so: &#x7B;.
+If you don’t want an expression, encode it, like so: \{.
```
wooorm, line 9: Escapes were a bit broken in old markdown and you had to resort to character references (“entities”). Since CommonMark, backslash escapes can now be used for all punctuation.

packages/remark-mdx/test/fixtures/expressions.md
```diff
@@ -14,7 +14,7 @@ Can contain more braces: <x y={{z: true}}/>
 
 ## Attribute expression
 
-<X {/*stuff!*/}/>
+<X {/*stuff!*/...d}/>
```
wooorm, line 17: Expressions now need to be valid expressions. JSX (e.g., in Babel) doesn’t allow just a comment here, it must be a valid spread. And we’re now matching that.

packages/vue/test/test.js
```diff
@@ -71,13 +71,13 @@ describe('@mdx-js/vue', () => {
...
-        '      <li>c</li>\n' +
-        '    </ol>\n' +
         '  </blockquote>\n' +
+        '  <ul>\n' +
+        '    <li>b</li>\n' +
+        '  </ul>\n' +
+        '  <ol>\n' +
+        '    <li>c</li>\n' +
+        '  </ol>\n' +
         '</div>'
```
wooorm, line 81: this was a bug in remark, fixed by micromark

packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -85,7 +83,9 @@ MDXContent.isMDXComponent = true`
   // Check JSX nodes against imports
   const babelPluginExtractImportNamesInstance = new BabelPluginExtractImportNames()
   const babelPluginExtractExportNamesInstance = new BabelPluginExtractExportNames()
-  const importsAndExports = [importStatements, exportStatements].join('\n')
+  const importsAndExports = []
+    .concat(importStatements, exportStatements)
```
ChristianMurphy, line 87: was this causing an issue?
Would spread syntax be appropriate here?
```suggestion
  const importsAndExports = [...importStatements, ...exportStatements]
```

packages/remark-mdx-remove-imports/package.json
```diff
@@ -42,8 +43,8 @@
     "unist-util-remove": "2.0.0"
   },
   "devDependencies": {
-    "remark-parse": "^8.0.0",
-    "remark-stringify": "^8.0.0",
+    "remark-parse": "^9.0.0",
+    "remark-stringify": "^9.0.1",
```
ChristianMurphy, line 47: is v9.0.1 specifically needed?
Would the broader `^9.0.0` range work?

packages/remark-mdx/package.json
```diff
@@ -40,22 +39,22 @@
...
   },
   "devDependencies": {
     "nyc": "^15.1.0",
     "prettier": "^2.0.5",
-    "remark-cli": "^8.0.0",
-    "remark-parse": "^8.0.2",
-    "remark-preset-wooorm": "^7.0.0",
+    "remark-cli": "^9.0.0",
+    "remark-parse": "^9.0.0",
+    "remark-stringify": "^9.0.1",
```
ChristianMurphy, line 51: same question as above

packages/mdx/mdx-hast-to-jsx.js
wooorm: they used to be two strings, now they’re two arrays.
I’m fine with spread or whatnot but there doesn’t seem to be a well defined baseline on what is and isn’t allowed, so I just did what I was familiar with!

packages/remark-mdx-remove-imports/package.json
wooorm: This particular change was because I pushed 9.0.1 while developing and yarn wasn’t picking up on the publish
This package is dropped so a non-issue: i’ll remove the deprecated packages in a next PR.

ChristianMurphy:
question on intended behavior.
In mdx 1, the following mdx file:
```markdown
{
```
would return `{` as a character with no errors.
In this PR the same file results in
```
{
  reason: 'Unexpected end of file in expression, expected a corresponding closing brace for `{`',
  line: 1,
  column: 2,
  location: {
    start: { line: 1, column: 2, offset: 1, _index: 1, _bufferIndex: -1 },
    end: { line: null, column: null }
  },
  source: 'micromark-extension-mdx-expression',
  ruleId: 'unexpected-eof'
}
```

Is this the expected behavior?
My impression was a JSX expression had to be inside a JSX tag/element block, is this no longer the case?
If it is the case, would it make sense to gracefully fallback to displaying the raw text similar to an unclosed link  `[](`?

ChristianMurphy:
Another open question, how should mdx expressions interact with other MDX extensions like frontmatter?
for example:
```mdx
---
property: <>{"hello world"}</>
---

Should the value of the yaml attribute `property` be `hello world` or `<>{"hello world"}</>`?

**Comment by ChristianMurphy** [comment]
Another curious edge case for you.
this is considered valid JSX/MDX:
```
<Icon name="hero
" />
```

this is not:

```
- <Icon name="hero
" />
```

failing with

```
{
  reason: 'Unexpected end of file in attribute value, expected a corresponding closing quote `"`',
  line: 2,
  column: 1,
  location: {
    start: { line: 2, column: 1, offset: 18, _index: 2, _bufferIndex: -1 },
    end: { line: null, column: null }
  },
  source: 'micromark-extension-mdx-jsx',
  ruleId: 'unexpected-eof'
}
```

**Comment by wooorm** [comment]
`{` and `<` on their own were already implemented and documented in the previous PR: https://github.com/mdx-js/mdx/pull/1039 (see “Most likely breaking changes from MDXjs 1”).

> My impression was a JSX expression had to be inside a JSX tag/element block, is this no longer the case?

This was also implemented there and further refined in the micromark rewrite. It was a common request for folks to embed expression in text (such as `## My {prop} heading`)

> If it is the case, would it make sense to gracefully fallback to displaying the raw text similar to an unclosed link [](?

I think the clearest rule is: with `<` and `{` you’re in JSX. You get errors. Otherwise you’re in markdown, you don’t get errors.
If you’re going to be silent about some cases, then: where’s the line? `<` + EOF? `<` + whitespace + EOF? `<` + `x` + EOF?

Markdown and JSX are very different languages. With JSX, folks are programming and expect errors, for `</x/>` to throw an error instead of display as text.

**Comment by wooorm** [comment]
> Another open question, how should mdx expressions interact with other MDX extensions like frontmatter?

Frontmatter is just a buffer for markdown, not a data structure. Similar how footnotes aren’t added inside frontmatter, JSX / expressions / ESM shouldn’t be added there IMO either.

Folks coming from markdown seem to want to use frontmatter, but MDX also has its ESM for data. I believe some of the other MDX maintainers prefer using ESM instead of frontmatter with MDX.

**Comment by wooorm** [comment]
For:

```mdx
- <Icon name="hero
" />
```

This matches how other multiline blocks in markdown work, take this markdown:

```markdown
- ```js
asdasd
```

Yields:

- ```js
asdasd

ChristianMurphy:
> This matches how other multiline blocks in markdown work, take this markdown

The reason I thought it interesting, it (seems) to conflict a bit with:

> I think the clearest rule is: with `<` and `{` you’re in JSX. [...] Otherwise you’re in markdown,

here we're in `<` (jsx) but we're applying a markdown block rule to it.

ChristianMurphy (approved):

packages/mdx/mdx-hast-to-jsx.js
ChristianMurphy: Node 8-15 support spread syntax, as well as every major browser (except IE) https://kangax.github.io/compat-table/es6/

wooorm:
> The reason I thought it interesting, it (seems) to conflict a bit with: ...
> here we're in < (jsx) but we're applying a markdown block rule to it.

I’m sure that’ll trip folks up yeah (although: it’s a bit of a weird case so it probably won’t occur much). Reason is that the list has to be dealt with before (or in fact, simultaneously as) the JSX tag. The list “knows” it’s not in a paragraph so therefore, when it sees lazy text, exits the list.
Alternative would be to “hack” something on top of markdown to allow lazy lines in both paragraphs and tags, but I don’t think that’s worth it.

Issue #1219, opened 2020-08-11
Improper rendering of tables with escaped HTML entities


akrantz01:
### Subject of the issue

When I try to put an escaped HTML entity in a code block within a table, it throws an error. I ran across this when using [Docusaurus](https://github.com/facebook/docusaurus) and needed to use the pipe character in a code block in a table. When I try to use an escaped HTML entity in a code block in a table cell (i.e. `<code>test with HTML and entities &amp; &#124;</code>`) in the playground, it throws the error `TypeError: e.trim is not a function`.

### Your environment

*   **OS**: Windows 10.0.18362
*   **Packages**: MDX 1.6.16, MDX React 1.6.16, Docusaurus v2.0.0-alpha61
*   **Env**: Node 12.18.2, NPM 6.14.5, Yarn 1.22.4

### Steps to reproduce

For Docusaurus, I made an example gist with the config and markdown to use: https://gist.github.com/akrantz01/05a917f9dea75fb73d9991beddc58c84. (I had initially raised this issue in facebook/docusaurus#3246).

In the playground, trying to render/parse the following causes an error:
```
| Number | Test | Expected |
|-|-|-|
| 1 | `&amp` | Preformatted text w/o rendered entity |
| 2 | <code>&amp;</code> | Preformatted text w/ rendered entity |
| 3 | `test with backticks` | Preformatted text |
| 4 | `test with backticks and entities &amp; &#124;` | Preformatted text w/o rendering the entity |
| 5 | <code>test with HTML</code> | Preformatted text |
| 6 | <code>test with HTML and entities &amp; &#124;</code> | Preformatted text w/ rendering the entity |
```

However, if you remove the last line, it displays with no errors.

### Expected behaviour

I expect for it to render the entire table with each of the HTML entities replaced as necessary.

![expected](https://imgur.com/GwkBPpI.png)

### Actual behaviour

The parser throws an error and it does not get displayed.

![actual](https://imgur.com/PS6kTl2.png)

tnorlund:
I'm having this issue too. Can we get an update on this?

Issue #1028, opened 2020-04-23
Support Class and decorator


YuunYang:
### Subject of the feature

Support writing decorator and class in MDX file

### Problem

Now, if I want to write class or decorator in mdx file, it will report an error. maybe we need add a feature to enable those grammar 

Like if I write a class in mdx file it will throw an error from '@mdx-js/loader/index.js': 
`Support for the experimental syntax 'classProperties' isn't currently enabled`,

It is actually not a bug in mdx, but a problem caused by lack of babel-loader plugin, I have tried to add those plugins in `mdx/mdx-hast-to-jsx.js`, solved those errors but caused another error -.-
![image](https://user-images.githubusercontent.com/15278220/80101005-5cb8cd80-85a3-11ea-8368-d4f70ac9f200.png)

### Expected behavior

Now, the mdx seems not support class declaration and other experimental syntax, So I want mdx to support more experimental

### Alternatives

No alternatives

wooorm:
Could you fill out expected behavior and alternatives? Could you provide example code of what you’d like to be supported?

YuunYang:
@wooorm Thanks for attention, I just updated the text

wooorm:
Thanks. Is this issue about:
1. Allowing fancy new features in MDX, by letting you define your own Babel plugins?
2. These two features specifically?

YuunYang:
Yes, because some experimental feature is used in many times, and it is much better if we can add custom babel or other configs

wooorm:
Alright, so I’m assuming you want a solution to 1) instead of 2).

There are some caveats with solving this because some parts of the babel config are required for MDXjs to work, so we can’t give users *all* control.

wooorm:
Hi @YuunYang, thanks for your patience!

The MDX loader is used alongside babel: https://github.com/mdx-js/mdx/tree/master/packages/loader. Have you tried configuring that?

To dig into this some more, I’d appreciate if you could update your issue with a minimal reproduction: something that a maintainer or contributor can run to reproduce your problem. Thanks!

Issue #606, opened 2019-06-18
Backslash sometimes inserted when using $ in text


lourd:
I noticed that on some parts of my portfolio site (made with gatsby and gatsby-mdx), there are extra backslashes being inserted in some places where dollar signs are used. For example:

- Go to https://descioli.design/grove and Find for `\$` — there are 6 occurrences on the page. There are 12 usages of `$`, though — not every $ is preceded by a backslash.
- The MDX source for that page is [here](https://github.com/lourd/descioli-design/blob/master/content/stories/grove/grove.mdx). Note: no backslashes.
- The rendering of the page is [here](https://github.com/lourd/descioli-design/blob/master/src/components/Story.tsx#L161-L191). Pretty standard AFAIK.

It only started happening when I switched from Remark to MDX. I didn't notice it when I did the initial switch.

### Environment

*   macOS and ubuntu, latest version
*   @mdx-js/mdx@1.0.20, gatsby-mdx@0.6.3
*   node@10.1.0, yarn@1.13.0

### Steps to reproduce

Clone my portfolio repo, https://github.com/lourd/descioli-design, and run it. No additional configuration needed.

### Expected behaviour

No extra backslashes.

### Actual behaviour

Extra backslashes

lourd:
This ended up being due to prettier. I'd set it up to run on pre-commit and missed the fact that it was also formatting .mdx files 🤦‍♂  Sorry to waste your time.

johno:
I'm going to reopen since MDX output is in fact different than remark's. It's a bug in the template literal escaping that we'll need to address.

lourd:
Just to be clear, it was Prettier that did the extra escape, not Remark. I
switched to MDX and accidentally had Prettier format my .mdx file in the
same commit, and incorrectly attributed the inserted backslash to the
Remark-MDX switch.

millette:
Should mdx handle \$ and remove the \ when output, since $ must be escaped due to math support - see <https://github.com/prettier/prettier/issues/6213>.

johno:
Yeah, ultimately we should detect the presence of `\$` and not escape it [twice](https://github.com/mdx-js/mdx/blob/master/packages/util/index.js#L30-L33).

In the ideal world Prettier wouldn't add the `$` escaping when there isn't inline math being used (or it at least wasn't the default), but it seems like we've reached the point where we need to solve it on our end.

millette:
I even tried something like `cost: 5{"$"}` but that was output litteraly.

[...]

Issue #741, opened 2019-08-21
RFC: MDX plugins


johno:
## Summary

The plugins from remark and rehype have been great to add in features and functionality, however I think MDX-specific plugins can take advantage of features more inline with the MDX language.

## Motivation

There are many common needs in the MDX ecosystem (frontmatter, table of contents, syntax highlighting, oembed, etc.) which don't make sense in core but are complex to set up or configure. Not to mention, there's some fragmentation developing around how these things are accomplished in different ecosystems (like Gatsby and Next.js).

Providing a plugin-based approach for these common pieces of functionality can help ensure that the best possible solution can be arrived upon and shared amongst a large group of folks. Not to mention, this could also provide a path for folks to experiment and innovate much more in userland.

A table of contents is the perfect example to show how a piece of functionality in an MDX-based stack needs a "vertical slice" of functionality. This allows a single plugin to set up everything it needs across all transformations `MDX => MDAST => HAST => JSX => React|Vue`.

### The way it works today

First, let's consider the ecosystem as it currently stands. For this RFC we will use adding a table of contents as the example functionality. Without MDX plugins, the current flow looks like:

- Add `remark-slug`
- Install `github-slugger` which `remark-slug` uses
- Customize `h1-h6` to be linkable using `github-slugger`
- Create a custom plugin that would assign headings for toc as an export (or use a custom loader, `gatsby-plugin-mdx`, etc)
- Build a component for import/render that turns the toc into a linked list
- Know the data export for the toc and pass to component

One could also use `remark-toc` to achieve some of this, but it doesn't give you much flexibility in how/when the toc is rendered and it doesn't allow the heading information to be imported from outside the document.

This isn't the end of the world, but it results in a lot of moving parts where a _single_ plugin could handle a lot of this and end up being more flexible.

## Basic (proposed) example

**Note**: This is not intended to be _the_ implementation/usage but an abstract example to illustrate the functionality.

##### Inject table of contents data

By default, `mdx-toc` might process the MDXAST and inject an export of the table of contents as an array.

```md
export const tableOfContents = [
  {
    value: 'Some heading',
    depth: 2
  }
]

# My document

<!-- ... -->
```

##### Import table of contents data for usage at the layout level

The injected data could be used at the layout level to add a table of contents.

```js
import { tableOfContents }, Document from './document.mdx'

export default () => (
  <DocumentLayout>
    <SidebarNav />
    <Main>
      <Document />
    </Main>
    <SidebarTableOfContents>
[...]

PR #1382, opened 2020-12-15
Refactor to generate JS AST

wooorm:
This PR changes the internals of the core `@mdx-js/mdx` package to generate a JavaScript syntax tree instead of a string. This fixes escaping issues such as #1219. It makes `mdx-hast-to-jsx` much more palatable. It also prevents several Babel parses. It paves the way for passing in Babel plugins, which is useful for users, but also for us to compile to `React.createElement`, `_jsx`, or Vue’s `h` calls directly and make MDX’s output directly usable.

* `babel-plugin-apply-mdx-type-props`: add `parentType` (used to be when serializing in mdx)
* `mdx`: use `rehype-minify-whitespace` to remove superfluous whitespace
* `mdx`: use `hast-util-to-estree` to transform hast to estree
* `mdx`: use `estree-to-babel` to transform estree to Babel
* `mdx`: generate estree/Babel instead of strings
* `mdx`: use `@babel/generator` to serialize Babel AST
* `vue`: stop supporting the react transform: (it doesn’t make sense)
* `vue`: fix support for props to components

Related to GH-741.
Related to GH-1152.

Closes GH-606.
Closes GH-1028.
Closes GH-1219.


packages/mdx/mdx-hast-to-jsx.js
```diff
@@ -1,348 +1,312 @@
-const {transformSync} = require('@babel/core')
```
wooorm, line 1: it might be easier to compare this file side by side. Because a lot changed

packages/vue/src/create-element.js
```diff
@@ -17,45 +14,35 @@ const defaults = {
 
 const own = {}.hasOwnProperty
 
-export default function createMdxElement(type, props, ...children) {
-  let node
-
+export default function createMdxElement(type, props, children) {
```
wooorm, line 17: I made it pretty hard for myself last time by thinking it’d be wise to also support the React babel transform, for Vue.
That was stupid. Now our Vue runtime only supports the Vue babel transform.

JounQin:
So does this mean that we can reuse the AST in `eslint-mdx` directly in the future? That's amazing!

wooorm:
Correct!

I now implemented it in the compiler that does hast -> string. But I think we should expose it in the future on the interface of the main [`mdx`](https://github.com/mdx-js/mdx/blob/68ff02c8129e2922f48b59bf51f4b967d248f397/packages/mdx) package

wooorm:
You can already do so with `hast-util-to-estree` btw:

https://github.com/syntax-tree/hast-util-to-estree/blob/c54a90845d12c59c40a8fa637d8efac3361752b1/test.js#L750-L769

packages/babel-plugin-apply-mdx-type-props/test/index.test.js
```diff
@@ -19,7 +19,7 @@ const transform = value => {
 describe('babel-plugin-add-mdx-type-prop', () => {
   test('should add `mdxType` to components', () => {
     expect(transform('var d = <div><Button /></div>').code).toEqual(
-      'var d = <div><Button mdxType="Button" /></div>;'
+      'var d = <div><Button parentName="div" mdxType="Button" /></div>;'
```
ChristianMurphy, line 22: could you expand on why `parentName` needs to be added?

wooorm: Because it’s already added: https://github.com/mdx-js/mdx/blob/3080444e2a40d79e3a3ac61ec1266154f95f77bf/packages/mdx/mdx-hast-to-jsx.js#L179

The reason is so that `ol.li` can be used as a key in `components`, in which case, when a `li` has a `parentName` of `ol`, that component is used

ChristianMurphy: with the move to `hast` + `estree` is it still needed?
Could adopters use https://github.com/syntax-tree/hast-util-select to have an even more flexible API?

wooorm: I'd like to see it removed. I'll open a new issue

JounQin:
> You can already do so with `hast-util-to-estree` btw:
> 
> [syntax-tree/hast-util-to-estree@`c54a908`/test.js#L750-L769](https://github.com/syntax-tree/hast-util-to-estree/blob/c54a90845d12c59c40a8fa637d8efac3361752b1/test.js#L750-L769)

@wooorm I tried this and find there is no `comments/tokens/loc/range` info in the transformed AST which are required by ESLint, can they be exposed or should I get them manually in other ways.

wooorm:
HTML comments are supported in hast-util-to-estree. JS comments are not supported in acorn.
For `token`, I don’t know. How do they work?
`loc` is added on certain things.
`range`: similar to `loc` but offset based (so wel also have it for certain things), so also PRs welcome

JounQin:
> HTML comments are supported in hast-util-to-estree. JS comments are not supported in acorn.
> For `token`, I don’t know. How do they work?
> `loc` is added on certain things.
> `range`: similar to `loc` but offset based (so wel also have it for certain things), so also PRs welcome

According to https://github.com/acornjs/acorn/blob/master/acorn/src/options.js#L117-L122, I got `tokens` by the following:

```js
const doc = `
<>
  {/* MDX is awesome! */}
  <div
    // MDX is awesome!
    style={{ color: "white", backgroundColor: "black", padding: "24px 32px" }}
  ></div>
</>;
`;

const comments = [];
const tokens = [];

const mdast = fromMarkdown(doc, {
  extensions: [
    mdxjs({
      acornOptions: {
        locations: true,
        ranges: true,
        onComment: comments,
        onToken: tokens,
      },
    }),
  ],
  mdastExtensions: [mdxFromMarkdown],
});
```

`tokens` are available here now, but not `comments`.

Acorn itself is working:

```js
let acorn = require("acorn");
const jsx = require("acorn-jsx");

acorn = acorn.Parser.extend(jsx());

const comments = [];

acorn.parse(doc, {
  onComment: comments,
});

console.log(comments);

// output
// [
//   { type: 'Block', value: ' MDX is awesome! ', start: 7, end: 28 },
//   { type: 'Line', value: ' MDX is awesome!', start: 41, end: 59 }
// ]

```

wooorm:
@JounQin One problem though is tokens and positional info from markdown, as your example is just JSX

wooorm:
@JounQin Also, comments are rather non-standard in estree. Different parsers add them differently. ESLint supports the `comments` array on Programs. But Babel and recast do not, so that’s not something I can solve right now (https://github.com/coderaiser/estree-to-babel/issues/3)

JounQin:
@wooorm We don't need comments on AST, but we should be able to detect comments on parsing like `onComment` of `acorn`.

wooorm:
> We don't need comments on AST, but we should be able to detect comments on parsing like onComment of acorn.

That is correct for ESLint, but incorrect for other tools

wooorm:
@JounQin Looks like this can attach comments: https://github.com/davidbonnet/astravel#astravelattachcommentsast-comments--ast, that makes it much easier than having to figure it out myself!

PR #1389, opened 2020-12-17
Sync

wooorm:
This cherry pick all relevant changes from `master` to `next`, so that we can swap the two out:
- `next` -> `main`
- set default to main
- `master` -> `v1`

I ignored dep here, we can do that again


docs/index.mdx
```diff
@@ -1,10 +1,26 @@
...
+<a
+  style={{
+    color: 'white',
+    textDecoration: 'none',
+    fontWeight: 'bold',
+    fontSize: 32
+  }}
+  href="https://blacklivesmatters.carrd.co/"
+>
+  #BlackLivesMatter &rarr;
```
ChristianMurphy, line 14: with mdx 2 will we need an escape char here? to prevent hashtag from being interpreted as a heading?

```suggestion
  \#BlackLivesMatter &rarr;
```

ChristianMurphy (approved):

docs/index.mdx
wooorm: Headings don’t form without the space in commonmark, so that shouldn’t be needed.
Is some tool complaining here or was this your hunch?

ChristianMurphy: that was a hunch

wooorm:
Fast-tracked this because it’s all already been reviewed, merged, and most often released as well

Issue #1267, opened 2020-09-16
[remark-mdx] Relax package.json dependency versions


berickson1:
### Subject of the feature
Relax package.json dependency versions

### Problem
Consumers of packages that define babel as a dependency (like remark-mdx) are forced to pull in a newer version of babel with updating this package. The package.json specifies an individual version of babel that can be used instead of a semver range. Without pulling in the latest version of this module, we're stuck on an older version of babel by default (7.10.5) (due to yarn lockfile de-duplication `yarn-deduplicate --strategy fewer`). After pulling the latest update of this module, we're forced to update to babel 7.11.1.

### Expected behavior

When updating remark-mdx (and other packages), we, as the package consumer, should be able to pick a compatible version of babel with the library instead of being required to select a specific version. If babel were to have a critical update/bug fix, we'd either need to manually change dependency version resolution or wait for this package to update/publish. Likewise, if a bug is introduced in babel and we pulled it in by updating this module, we'd need to downgrade this module.

I would expect your package.json for something like babel to be:
`"@babel/core":"^7.9.0"` (possibly some version slightly higher if there's a good reason)
instead of
`"@babel/core":"7.9.0"`(in the latest case I realize it's 7.11.1, just providing an example)

### Alternatives

Caret, tilda or range versions for downstream dependencies. Ideally any solution would allow the consumer a bit more flexibility to upgrade/downgrade/manage their own dependencies without having to override module version resolution.

wooorm:
I’m personally in favor of what you propose, but several other core contributors specifically want it this way, see: https://github.com/mdx-js/mdx/issues/865.

berickson1:
I'd argue that yarn/npm lock files protect from random breakages in existing projects. The biggest concern is that if any module you depend on has a vulnerability (and mdx packages haven't updated yet), there's little recourse for a consumer of mdx to upgrade it without doing explicit version pinning. Forcing an update of mdx to resolve a downstream vulnerability potentially introduces additional risks for projects as well.

Issue #1385, opened 2020-12-15
An alternative “shortcode” (`mdxType` alternative) proposal


wooorm:
Currently, MDX is walking the whole JS program, and adding an `mdxType` prop to all JSX elements. It does this, so components can be overwritten at runtime.

Take this MDX:

```mdx
* a
* b
```

Which turns into (simplified to get the point across, the actual is more complex):

```jsx
var ol = () => throw "error"
var li = () => throw "error"

<ol mdxType="ol">
  <li mdxType="li" parentName="ol">a</li>
  <li mdxType="li" parentName="ol">b</li>
</ol>
```

After compiling the JSX away, you get:

```js
var ol = () => throw "error"
var li = () => throw "error"

func(ol {mdxType: "ol"}, [
  func(li, {mdxType: "li", parentName: "ol"}, ["a"]),
  func(li, {mdxType: "li", parentName: "ol"}, ["b"]),
])
```

Then, the runtime, supports replacing things through `components`.

It has a `func` that ignores its first argument, instead taking `mdxType` (and `parentName`) to look whether it has a different component defined in `components`, and uses that instead of the either imported or defined throwing components.

The above is the simplified example for “normal” tags. The same happens for actual components (Capitalized):

```js
var MyComponent = () => throw "error"

func(MyComponent, {mdxType: "MyComponent"}, [])
```

Which is then overwritten at runtime...

Even weirder is:

```js
import MyComponent from "place"

func(MyComponent, {mdxType: "MyComponent"}, [])

// Passed in runtime:
{components: {"MyComponent": () => /* this is used instead of the imported component! */}}
```

This ✨ magic ✨ is complex and undocumented. I want to remove that complexity. Here’s one alternative.

We instead compile all standard markdown (html?) things to a namespace object: `p` becomes `mdx.p`, `strong` becomes `mdx.strong`, etc:

```jsx
<mdx.ol>
  <mdx.li>a</mdx.li>
  <mdx.li>a</mdx.li>
</mdx.ol>
```

Then, in `mdxContent` (the wrapper function around all JSX components), we’re being passed runtime components. We take those and use them to fill `mdx.p = components.p || 'p'`.

This might solve #821, because the JSX `<p>x</p>` is not overwritten whereas the markdown paragraph `x` is, depending on where we implement the changing of `p` to `mdx.p`.

[...]

Issue #1015, opened 2020-04-14
Reduce churn


wooorm:
Hi folks! 👋

I just came back to this repo after a while, going through issues, PRs, dependency updates, etc, and this may just be me but to me it feels like there’s a bit much churn to maintain this repo:

* Building is long: takes about 5 to 8 minutes for a CI job to finish; not too important, but it takes 2:30 to install dependencies, between 2 and 4 minutes to store a cache version of them, …
* GH CI sometimes randomly fails: for a PR pushed from the repo, there are 18 CI builds done 2 (PR + Push) * 3 (macos, windows, ubuntu) * 3 (node 8, 10, 12). Looking through the renovate PRs, about 1/3rd of them randomly failed.
  Also seems that GH runs up to ± 6 builds at the same time, so in total it takes 15m-20m to check everything
* Deps are completely version locked in `package.json`, which I think makes sense for the MDX library, but it creates noise for dev-dependencies, examples, and the site: I find the current list of open PRs and list of commits pretty unreadable
* Our tests are based on fixtures, which is fine, but if something [meaningless changes](https://github.com/mdx-js/mdx/pull/980/files#diff-ba661d17926236b9a6e9f243c362dbcc) it causes a failure

---

While tooling is great (hey, I spend all my time building tools), and maybe I’m just horribly out of touch with what modern development is supposed to look like, I think it would help maintainers and contributors if we could make the repo simpler and builds faster and less fragile! A couple of ideas:

- Don’t build on macOS: it’s mostly similar to ubuntu and I believe most maintainers of MDX have a mac
- Don’t build on Node versions except for the outliers (so at the moment, do test on 8 and 12, but not on 10)
- Marie Kondō dev-dependencies / tools
- Move `remark-*` stuff to the @remarkjs org
- Figure out if we can configure renovate better
- Rework tests! (this is a big thing that should be discussed somewhere else)

ChristianMurphy:
> but it takes 2:30 to install dependencies

npm 7 may help, v7 adds support for workspaces which was previously one of the main draws for yarn.
Testing locally a clean yarn install:

```
$ rm -rf **/node_modules
$ node --version
v15.1.0
$ yarn --version
1.22.10
$ yarn
```

takes around 59-68 seconds.
while a clean npm install
```
$ rm -rf **/node_modules
node --version
v15.1.0
$ npm --version
7.0.8
$ npm ci
```

takes between 42 to 48 seconds.

[...]

Issue #865, opened 2019-12-12
Relax dependency versions


NMinhNguyen:
### Subject of the feature

Relax dependency versions.

### Problem

See https://github.com/mdx-js/mdx/pull/858#issuecomment-563336666. Essentially, when installing an MDX package such as `@mdx-js/mdx`, I would rather not have a pinned version of any `@babel/*` dependencies installed - this causes duplicate versions and can plausibly lead to subtle bugs. If you look at https://unpkg.com/@mdx-js/mdx@1.5.1/package.json, you'll see `@babel/core` `7.6.2` gets installed, as well as fixed versions for `@babel/plugin-syntax-jsx` and `@babel/plugin-syntax-object-rest-spread`.

### Expected behavior

Dependencies should use carets (e.g. `^7.6.2`).

### Alternatives

Yarn users are able to use things like Yarn [resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) but it's not ideal to have to resort to resolutions.

/cc @JounQin @johno

timneutkens:
Based on our experience at @zeithq not using pinned deps will lead to the package randomly breaking a *lot*.

NMinhNguyen:
@timneutkens I think that pinning versions can give a false sense of security - e.g. all Babel deps use carets for their own deps, so you get non-deterministic installs with every new install without a lockfile anyway.

timneutkens:
@NMinhNguyen it greatly reduces the vector of breakage though.

NMinhNguyen:
@timneutkens Perhaps you could consider relaxing some common build tooling dependencies that aren’t really meant to be duplicated (like Babel)?

JounQin:
I'm not a fan of pining versions or depeendencies neither personally.

https://github.com/1stG/configs/blob/master/packages/renovate-config/package.json#L21

NMinhNguyen:
@JounQin with the currently published packages, an older version of `@babel/core` gets installed because my other dependencies have a dependency on a later version of `@babel/core` (using a caret), so it’s unavoidable even with deduplication.

JounQin:
> @JounQin with the currently published packages, an older version of `@babel/core` gets installed because my other dependencies have a dependency on a later version of `@babel/core` (using a caret), so it’s unavoidable even with deduplication.

[...]

PR #1392, opened 2020-12-18
Update dependencies, dev-dependencies

wooorm:
This updates the dependencies and dev-dependencies in `packages/`. Unfortunately, either updating to webpack 5 or updating to react 17 crash the webpack loader with a react error, with an [invalid hook call warning](https://reactjs.org/warnings/invalid-hook-call-warning.html) for `useMDXComponents`:

<https://github.com/mdx-js/mdx/blob/dafdf6d70affa5dba0b3b7070f7a310b70bbf775/packages/react/src/context.js#L15>

Which might have to do with the magic of shortcodes (#1385), or something else, I have no clue.

Furthermore, this loosens package dependencies instead of locking them, which relates to GH-865, GH-1015, and GH-1267. It was a long and divided discussion before and the reason for changing now is: While the package currently doesn’t break easily (it was mentioned that unlocking packages might cause that), we are currently *locked* on security vulnerabilities. We’re not getting any patches, and MDX isn’t released that frequently or maintained that actively, so MDX users are stuck. If folks want to lock: npm and yarn have package locks.

Closes GH-1267.
Closes GH-1375.


packages/babel-plugin-apply-mdx-type-props/package.json
```diff
@@ -25,11 +25,8 @@
     "remark",
     "babel"
   ],
-  "peerDependencies": {
-    "@babel/core": "^7.10.5"
-  },
   "dependencies": {
-    "@babel/helper-plugin-utils": "7.10.4"
+    "@babel/helper-plugin-utils": "^7.10.4"
```
ChristianMurphy, line 29: would it make sense to broaden these to
```suggestion
    "@babel/helper-plugin-utils": "^7.0.0"
```
?

ChristianMurphy (approved):
LGTM :+1: 

---

> updating to react 17

In a few other upgrades to React 17, the issue has been transitive dependencies on React 16, causing both React 16 and 17 to be in the tree.
Triggering

> You might have more than one copy of React in the same app

- https://reactjs.org/warnings/invalid-hook-call-warning.html

packages/babel-plugin-apply-mdx-type-props/package.json
wooorm: I’d say maybe later, I’m not too familiar with the internals there to know if we’re depending on internal features released in those minor versions

wooorm:
> In a few other upgrades to React 17, the issue has been transitive dependencies on React 16, causing both React 16 and 17 to be bundled.

Yeah, that might be it indeed. I’ll try again somewhere soon

PR #1393, opened 2020-12-18
Remove code for deprecated packages

wooorm:
Some packages aren’t used anymore on `main`, this removes their code, types, tests, etc, which should speed up tests and linting.

They could also actually be removed, but that might break some links.


packages/test-util/package.json
```diff
@@ -1,35 +1,5 @@
...
-    "mdx",
-    "markdown",
-    "react",
-    "jsx",
-    "remark",
-    "mdxast"
-  ],
-  "scripts": {},
-  "dependencies": {},
   "gitHead": "bf7deab69996449cb99c2217dff75e65855eb2c1"
```
ChristianMurphy, line 4: is gitHead still needed as well?

wooorm: could probably also be removed indeed!

wooorm: done

ChristianMurphy (approved):

Issue #961, opened 2020-02-29
Make `export { default } from "…"` work


wereHamster:
When I define the page wrapper in a different module and want to use it from an .mdx file, I have to write both an import and export declaration, ie.

```
import Wrapper from "…"
export default Wrapper
```

In JavaScript, this is semantically equivalent to:

```
export { default } from "…"
```

But mdx-js does not apply the wrapper in that case.

PR #1395, opened 2020-12-19
Add support for named export default from

wooorm:
Previously, only `export default const a = 1` was supported.
This adds support for `export {default} from "a"` and
`export {a as default} from "b"` declarations.

Closes GH-961.


ChristianMurphy (approved):

johno (approved):
😻

PR #1396, opened 2020-12-19
Remove `x0`, `zero` examples

wooorm:
Two frameworks that have gone stale.


ChristianMurphy (approved):

johno:
IMO we should keep the docs pages but not link to them, and maybe mark as stale on the page itself. Aside from that, looks great!

wooorm:
@johno Added the docs back with a deprecation note!

Issue #1384, opened 2020-12-15
Babel or estree?


wooorm:
### Subject of the discussion

With https://github.com/mdx-js/mdx/pull/1382, we now have a JavaScript syntax tree.

The tree starts out in estree: as markdown + mdx.js is parsed simultaneously, I needed a JavaScript parser in [`micromark-extension-mdxjs`](https://github.com/micromark/micromark-extension-mdxjs), and I chose a small and fast one: acorn. Which comes with estree.
Acorn is small, 30kb minzipped. acorn-jsx is 4kb. `astring` (a generator) is also 4kb.

Previously, in this project, we used Babel for plugins.
Babel is giant. `@babel/core`, which has methods to run Babel plugins, is like 220kb minzipped. `@babel/generator` is 63kb. `@babel/parser` is 60kb.`@babel/traverse` is 165kb (it includes both the parser and the generator).

Estree has the drawback of being a fragmented ecosystem: there are no nice parsers that support comments; there are no tree-wakers or compilers that support JSX. And importantly, as as we use JSX, we’d want to turn JSX into function calls (React/preact/vue), but those are all Babel plugins. We could use estree but then users would still need to run Babel afterwards.

Babel has the drawback of being giant and slow. But the good thing is that the JSX -> JS compilers all live there.

### Problem

What should we go with?
We can’t turn JSX -> JS unless we’re using Babel (well, we could, the babel plugin to turn JSX -> `_jsx()` / `React.createElement` is [800l](https://github.com/babel/babel/blob/main/packages/babel-helper-builder-react-jsx-experimental/src/index.js)). Most users probably want to use Babel plugins to turn their fancy features into whatever.
An estree-only system as a base for MDX would be ✨✨✨.  `@mdx-js/runtime` is now 350kb minzipped. That could go down to 100kb or less?

ChristianMurphy:
> Estree has the drawback of being a fragmented ecosystem: there are no nice parsers that support comments; there are no tree-wakers or compilers that support JSX

ESLint's parser and walker have solid ESTree + Comment + JSX support
https://github.com/eslint/espree
https://github.com/eslint/eslint-visitor-keys

Prettier has espree with Comment + JSX support for code gen https://github.com/prettier/prettier/blob/902d524d2f1776efe0b110c1a24813d4d7fcb9d0/src/language-js/printer-estree.js
escogen is close to having ESTree + JSX support https://github.com/estools/escodegen/pull/391

[...]

Issue #704, opened 2019-08-01
Explore moving to babel standalone or offering a more browser friendly build


johno:
Since `@babel/core` attempts to bring along the `fs` module we're often forced to [do annoying webpack config](https://mdxjs.com/getting-started/webpack#running-mdx-in-the-browser) and other hacks.

For environments where that control might not exist we should look into using `@babel/standalone` or even offering an `@mdx-js/standalone` that things like `@mdx-js/runtime` could use.

JulioC:
Is this available in @mdx-js/mdx@2.0.0-next.1? If not, is there any estimate for its release? 

I'm curious to know how much improvement this will make in my bundle size

Issue #1046, opened 2020-04-30
RFC: Support variable declaration (const/let/var)


johno:
Currently, users leverage `export` syntax in order to instantiate local variables for usage in an MDX file. We could introduce `const`/`let`/`var` to better define an MDX document's API by allowing users to not need to overload exports for local variables.

Nothing else would really differ from the `export` syntax except that we'd read `const`/`let`/`var` directly.

## const

```jsx
const MyComponent = () => <div style={{ padding: 20, backgroundColor: 'tomato' }} />
```

## let

```jsx
let MyComponent = () => <div style={{ padding: 20, backgroundColor: 'tomato' }} />
```

## var

```jsx
var MyComponent = () => <div style={{ padding: 20, backgroundColor: 'tomato' }} />
```

wooorm:
I think there are two alternatives (using a different style to show that they’d allow more too):

- **b)** executable code
  ````markdown
  # heading

  ```js run
  function MyComponent() {
    return <div style={{ padding: 20, backgroundColor: 'tomato' }} />
  }
  ```

  paragraph
  ````
- **c)** expressions
  ````markdown
  # heading

  {
    function MyComponent() {
      return <div style={{ padding: 20, backgroundColor: 'tomato' }} />
    }
  }

  paragraph
  ````

aleclarson:
This feature would be much appreciated, as `export const` breaks Fast Refresh when its value is not a component.

ng-hai:
I'm not sure my issue related to this. I'm using MDX/React with Parcel bundler, somehow it does not interpolate React expression

```mdx
export const name = "world";

# Hello {name.toUpperCase()}

{1+1}

<p>{1 + 1}</p>
```

And how it renders
```
Hello {name.toUpperCase()}
{1+1}
2
```

wooorm:
@ng-hai that is unrelated to this proposal. You show MDX 2 features and Parcel does not support MDX 2 apparently.

wooorm:
Closing as there is little traction on this.
The main problem I see is that we need to make exceptions for more and more words. And what about `function MyComponent`, for example?
There is a good alternative possible, as outlined above: `~~~js eval`. That can be supported in a plugin. The pseudocode for that is in [in this comment](https://github.com/mdx-js/mdx/issues/701#issuecomment-947030041).

PR #1399, opened 2020-12-20
Replace Babel w/ estree

wooorm:
This removes the last three custom Babel plugins we had and replaces them with estree versions. Furthermore, it removes `@babel/generator`.

For the plugins, we were only looking at ESM import/exports, but right now we’re delegating work to `periscopic` to look at which things are defined in the top-level scope. It’s a bit more complex, but this matches better with intentions, fixes some bugs, and prepares for a potential future where other ES constructs are allowed, so all in all should be a nice improvement.

For serializing, we’re switching to `astring`, and handling JSX for now internally (could be externalized later). `astring` seems fast and is incredibly small, but is not very popular. We might perhaps see bugs is serialization in the future because of that, but all our tests seem fine, so I’m not too worried about that.

Estree remains a somewhat fragmented ecosystem, such as that the tree walkers in `periscopic` and `astring` are different, so we might also consider writing our own serializer in the future. Or, when we implement Babel’s React JSX transform ourselves, could switch to another generator, or at least drop the JSX serialization code here.

Because of these changes, we can drop `@babel/core` and `@babel/generator` from `@mdx-js/mdx`, which drops the bundle size of **from 349kb to 111kb. That’s 68%. Pretty nice**. This should improve downloading and parsing time of bundles significantly. Of course, we currently still have JSX in the output, so folks will have to resort to Babel (or `buble-jsx-only`) in another step.

For performance, v2 (micromark) was already an improvement over v1. On 1000 simple files totalling about 1mb of MDX:

* v1: 3739ms
* v2: 2734ms (26% faster)
* v2 (w/o babel): 1392ms (**63% faster**).

Of course, this all really depends on what type of stuff is in your MDX. But it looks pretty sweet!

✨

Related to GH-1046.
Related to GH-1152.
Related to GH-1338.
Closes GH-704.
Closes GH-1384.


johno (approved):
❤️

PR #1415, opened 2020-12-24
Update examples, `react`

wooorm:
* Update `react`, `react-dom`
* Clean Next, Razzle, Vue examples
* Update parcel example, parcel plugin, run tests
* Prefer `.mdx` extension over `.md`
* This does not yet touch CRA
* It’s impossible to get Gatsby working in our workspace (they fork MDX but resolve from their dependency on old MDX)
* I can’t get webpack 5 working either

Related to GH-1292.


docs/advanced/plugins.mdx
```diff
@@ -53,7 +53,7 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.mdx?$/,
+        test: /\.mdx$/,
```
wooorm, line 56: Push our own extension.
Several tools already handle markdown (e.g., parcel), which we shouldn’t overwrite. And MDX is somewhat different, so it’s good to push the different extension

examples/next/pages/_app.js
```diff
@@ -1,12 +1,14 @@
...
-export default ({Component, pageProps}) => (
-  <MDXProvider components={mdComponents}>
-    <Component {...pageProps} />
+const Page = ({Component}) => (
+  <MDXProvider components={components}>
+    <Component />
   </MDXProvider>
 )
+
+export default Page
```
wooorm, line 14: exports must have an inferrable name according to next

examples/terminal/example.js
```diff
@@ -1,23 +1,23 @@
...
-const {render, Color, Box, Text} = require('ink')
+const {render, Box, Text} = require('ink')
 const MDX = require('@mdx-js/runtime')
 const {MDXProvider} = require('@mdx-js/react')
 
 const MDXContent = `
 # Hello, world!
 
-From <Color bgBlack white bold>    MDX!    </Color>
+From <Text backgroundColor="black" color="white" bold>MDX!</Text>
```
wooorm, line 9: Ink changed

examples/vue/src/App.vue
```diff
@@ -1,29 +1,20 @@
 <template>
```
wooorm, line 1: mostly a smaller, MVP, example

packages/parcel-plugin-mdx/package.json
```diff
@@ -33,11 +33,16 @@
...
+    "react": ">=16"
   },
   "dependencies": {
     "@mdx-js/mdx": "^2.0.0-next.8",
     "parcel-bundler": "^1.12.4"
   },
+  "scripts": {
+    "test-api": "jest test",
+    "test-coverage": "jest test --coverage",
+    "test": "yarn test-coverage"
```
wooorm, line 45: this package had tests, but they were never tested

examples/next/pages/_app.js
ChristianMurphy: 1. isn't that a requirement for fast refresh, not for Next as a whole. (https://nextjs.org/docs/advanced-features/codemods#nextjs-9)
2. arrow functions don't have names, to five this a name wouldn't it need to be `function Page ({Component}) {`?

ChristianMurphy (approved):

examples/next/pages/_app.js
wooorm: For 2, I think v8 does infer the .name of arrow functions in a variable now? Or they're walking the AST.
For 1, probably. It was a warning and they suggested this, so I applied it

ChristianMurphy: huh, that seems new, IIRC originally arrow functions never had a name. :man_shrugging: 
Ah well it works now, good to know.

ChristianMurphy (approved):

Issue #365, opened 2018-12-22
Error when trying to run the create-react-app example


tornordqvist:
### Subject of the issue

I get an error when trying to run the create-react-app example.

### Your environment

Windows 10
Node v10.14.0
NPM 6.4.1

### Steps to reproduce

Easy. I do:
git clone
cd mdx/examples/create-react-app
npm i
npm start

### Expected behaviour

The app should run without errors.

### Actual behaviour

I get this error when running the app;

```
Failed to compile
./src/App.js
Module not found: You attempted to import C:\Repos\mdx\examples\create-react-app\node_modules\.cache\mdx.macro\Content.796904d8db.mdx.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
```

tornordqvist:
The mdx.macro should read the Content file from /src but seems to look locally in the node_modules/.cache/mdx.macro folder.

briedel1:
I am seeing the same behavior. Did you have any luck?

briedel1:
This worked https://www.npmjs.com/package/mdx-loader.

KeziahMoselle:
I got this error, any fix planned without ejecting ?

johno:
My guess is it's [this line](https://github.com/frontarm/mdx-util/blob/master/packages/mdx.macro/mdx.macro.js#L24) in `mdx.macro` that doesn't work with Windows environments.

aaronSig:
I've managed to work around by adding the below to my .env file

```
NODE_PATH=src 
```

now instead of 

`import X from "../components/X"`  use `import X from "components/X"`

freaksauce:
I am experiencing the same issue on macOS 10.14.3 so it's not a Windows issue. I tried setting up a new CRA app using the latest mdx package with a simple import - [issue raised for mdx.macro](https://github.com/frontarm/mdx-util/issues/44)

devuxer:
I'm getting this error on Windows with a pristine CRA --typescript project.

aress31:
Getting this issue then on a newly created project following the doc - my env is Windows.

nicobrinkkemper:
I don't understand why this issue is still open since Dec 22, 2018. 
```
var _jsxFileName = "C:\\Users\\nico\\Documents\\Code\\fourymm\\src\\data\\Welcome.mdx";
import React from 'react';
import { MDXTag } from '@mdx-js/tag';
[...]

Issue #1388, opened 2020-12-16
please remove totally outdated and unmaintained mdx.maco


TheDiveO:
### Subject of the issue

The official mdx-js documentation requests to install mdx.macro along create-react-app. However, mdx.macro has unmaintained and totally outdated mdx-js dependencies to version <1.0.0. The result is that several documented functions of 1.x+ do not work. It took me quite some time to figure this out. Unfortunately, the author of mdx.macro is unable to upgrade the dependencies. When trying to manually overriding the outdated dependencies, the generated MDX code doesn't work due to changes in the API before and after 1.0+.

### Your environment

*   **OS**: Ubuntu 20.04

### Steps to reproduce

`yarn install mdx.macro`

### Expected behaviour

Currently documented mdx features, such as inline components, to work.

### Actual behaviour

Currently documented mdx features, such as inline components, don't work.

ChristianMurphy:
this is in progress at https://github.com/mdx-js/mdx/pull/589, if you want to do the finishing touches on the docs, a follow up PR would be welcome

ChristianMurphy:
this is a duplicate of #365, closing this issue to consolidate the conversation over there.

TheDiveO:
I'm stuck because that PR example don't work for me:

    Unexpected '!' in '!babel-loader!mdx-loader!./About.mdx'. Do not use import syntax to configure webpack loaders  import/no-webpack-loader-syntax

ChristianMurphy:
The message you shared is a lint warning, not a webpack error.
So it may be, and probably is, working?

TheDiveO:
It classifies as "Failed to compile" ... that's not looking like a linting issue.

ChristianMurphy:
Following up on

> closing this issue to consolidate the conversation over there.

Can we please have the conversation at  #365 instead of both here and there

TheDiveO:
Using "// eslint-disable-next-line import/no-webpack-loader-syntax" immediately before the import now results in "Cannot find module '!babel-loader!mdx-loader!./About.mdx' or its corresponding type declarations.ts(2307)"

PR #1422, opened 2020-12-28
Update CRA guide

wooorm:
Create React App is the most looked at resource by users here on GitHub. But it’s suggesting an old, unmaintained, and buggy way to use MDX.

This instead updates the guide to use our maintained projects, without having to eject from CRA.

As CRA itself is an ever-changing “init” tool, which can support MDX by following a couple steps, I don’t think it’s wise to have an example in the project: we want folks to do `npx create-react-app ...`, not clone our custom example.

Not having CRA checked in also makes for a faster `yarn install`.

Perhaps developing our own [CRA template](https://create-react-app.dev/docs/custom-templates) might be nice for the future, but for now I’ve kept it at an up to date and working guide.

Related to GH-1015.
Related to GH-1388.

Closes GH-365.
Closes GH-589.


docs/getting-started/create-react-app.mdx
```diff
@@ -1,50 +1,92 @@
...
+the following:
+
+```sh
+yarn add babel-loader@$VERSION --dev
+```
+
+(where `$VERSION` is the version number printed in the error message, such as
+`8.1.0`).
+
+Then, running `yarn start` again should do the trick! ✨
```
ChristianMurphy, line 88: CRA + TS users may still see an error, due to MDX not having a typing.
It may be worth linking to the TypeScript guide for `@mdx-js/loader` as well

ChristianMurphy (approved):

docs/getting-started/create-react-app.mdx
wooorm: how about this?

ChristianMurphy (approved):
Thanks @wooorm, glad to see this doc update finally make it across the finish line. :checkered_flag:

PR #1439, opened 2021-01-14
Update webpack.mdx

ChrisChinchilla:
The comment 

> handler for `.md` and `.mdx` files:

Doesn't suit the example, so updated the example to match


ChrisChinchilla:
Actually scrape that I realised I was misreading the regex

PR #1460, opened 2021-02-10
Add missing react peer dep to the webpack loader

mskelton:
`@mdx-js/loader` was missing a peer dependency on react causing a warning when installing with Yarn 2.

```sh
➤ YN0002: │ @mdx-js/loader@npm:1.6.22 doesn't provide react (p12550), requested by @mdx-js/react
```

This PR fixes the issue by adding the peer dependency to `@mdx-js/loader` which follows [Yarn's recommendations](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0).

### DIFF
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -40,6 +40,9 @@
     "test-types": "dtslint types",
     "test": "yarn test-coverage && yarn test-types"
   },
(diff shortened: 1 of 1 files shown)


johno:
Thanks!

PR #1468, opened 2021-02-23
Cache MDX compiler in the Webpack loader

remcohaszing:
A new MDX compiler was created for every file that’s processed. This means that any initialization logic in remark or rehype plugins was always run for every file.

By reusing the MDX compiler, remark and rehype plugins can run heavy setup logic once, just like when they are called by unified directly.

In practice I use this in [remark-mermaidjs](https://github.com/remcohaszing/remark-mermaidjs). This starts a browser once to process a bunch of files, but it requires the same unified processor to run to be able to do so.

The same principle is used in [ts-loader](https://github.com/TypeStrong/ts-loader/blob/953358eb1b090dd9add0824c4746220b0ebe56c1/src/instances.ts#L37-L86) to cache the TypeScript compiler.

### DIFF
--- a/packages/loader/index.js
+++ b/packages/loader/index.js
@@ -1,27 +1,39 @@
 const {getOptions} = require('loader-utils')
-const mdx = require('@mdx-js/mdx')
+const {createCompiler} = require('@mdx-js/mdx')
 
 const DEFAULT_RENDERER = `
 import React from 'react'
 import {mdx} from '@mdx-js/react'
 `
 
+const pragma = `
+/* @jsxRuntime classic */
+/* @jsx mdx */
+/* @jsxFrag mdx.Fragment */
+`
+
+const compilerCache = new Map()
+
 module.exports = async function (content) {
+  if (!compilerCache.has(this.query)) {
+    const {renderer = DEFAULT_RENDERER, ...opts} = getOptions(this)
+    compilerCache.set(this.query, [renderer, createCompiler(opts)])
+  }
+
   const callback = this.async()
-  const options = Object.assign({}, getOptions(this), {
-    filepath: this.resourcePath
-  })
+  const [renderer, compiler] = compilerCache.get(this.query)
 
   let result
 
   try {
-    result = await mdx(content, options)
+    result = await compiler.process({
+      contents: content,
+      path: this.resourcePath
+    })
   } catch (err) {
     return callback(err)
   }
 
-  const {renderer = DEFAULT_RENDERER} = options
-
-  const code = `${renderer}\n${result}`
+  const code = `${renderer}${pragma}${result}`
   return callback(null, code)
 }


packages/loader/index.js
```diff
@@ -1,27 +1,39 @@
...
 const DEFAULT_RENDERER = `
 import React from 'react'
 import {mdx} from '@mdx-js/react'
 `
 
+const pragma = `
+/* @jsxRuntime classic */
+/* @jsx mdx */
+/* @jsxFrag mdx.Fragment */
+`
```
ChristianMurphy, line 13: why does pragma need to move to this level? :thinking:

packages/loader/index.js
```diff
@@ -1,27 +1,39 @@
...
+
+const compilerCache = new Map()
```
ChristianMurphy, line 15: Could cache size be a concern?
Would FIFO cache or LRU cache make sense here?
Maybe something like https://www.npmjs.com/package/quick-lru?

remcohaszing: The maximum cache size is the times `@mdx-js/loader` is used across all Webpack configurations of a project. Typically this is 1. In exceptional situations there may be more. I don’t think this will grow very big.

I also tried using `WeakMap`, but that doesn’t support primitive keys, which `query` can be.

packages/loader/index.js
remcohaszing: As can be seen [here](https://github.com/mdx-js/mdx/blob/main/packages/mdx/index.js#L46), the pragma is added by the `compile` function, not by the processor. Because the processor is now invoked by the loader directly instead of the `compile()` function, the pragma needs to be added by the loader manually.

ChristianMurphy (approved):

JounQin (approved):

slorber:
@wooorm has this caching system been removed in v2?

I don't see it anymore here: https://github.com/mdx-js/mdx/blob/main/packages/loader/lib/index.js

wooorm:
I believe xdm was fast enough, it does a lot less work. Do you need it then?

remcohaszing:
The main issue isn’t the MDX compiler, but the plugins that may be slow.

Specifically `remark-mermaidjs`, which uses puppeteer, is slow. A lot of heavy lifting could be ran only once instead of per file when reusing the same browser instance, but this requires the unified pipeline to be reused.

Personally I haven’t been affected by this issue lately, because I disabled `remark-mermaidjs` in development (watch) mode. Still this optimization would be nice for those who do use it in development, and it would also significantly speed up production builds for my project.

wooorm:
Oh right. This was around the xdm time, I believe the plan was to land it there too but it never did?
xdm has proper caching mechanisms for rollup, esbuild, and node, which have been ported over to mdx-js/mdx. But not for webpack.

wooorm:
added again!

PR #1475, opened 2021-03-03
Fix a typo of the sample Webpack config in doc

Gowee:
The context says:

> ...define the following webpack.config.js extension handler for .md and .mdx files...

> If you only want the loader for `.mdx` files you can _change_ the regex to `/\.mdx$/`.

So I suppose the RegExp in the sample config should have be `test: /\.mdx?$/`.


johno (approved):
Thanks!

PR #1483, opened 2021-03-13
fix a reference error

XeryYue:
well,When i use this plugin,i found an error in vue3,I try to use import { MDXProvider } from '@mdx-js/vue-loader' to resolve my question


JounQin:
@mdx-js/maintainers Anyone can check the Vercel settings? It seems incorrect to require authorization on PR.

PR #1495, opened 2021-03-25
ci: simplify ci workflows, change master to main on ci

JounQin:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   default:
-    name: Build and Test on ${{ matrix.os }} with Node.js ${{ matrix.node }}
+    name: Test on ${{ matrix.os }} with Node.js ${{ matrix.node }}
     strategy:
       matrix:
         os:
@@ -25,11 +25,3 @@ jobs:
           CI: 'true'
       - name: Test
         run: yarn test
-      # To do: macos isn’t used anymore
-      - name: Publish CI tag to npm
-        if: matrix.node == 12 && matrix.os == 'macOS-latest' && github.event_name == 'push' && github.ref	== 'refs/heads/master'
-        run: |
-          npm set //registry.npmjs.org/:_authToken $NPM_AUTH_TOKEN
-          yarn publish-ci
-        env:
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   default:
@@ -8,15 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # for lerna
 
-      - uses: actions/setup-node@v2-beta
+      - uses: actions/setup-node@v2
         with:
           node-version: 14
-
-      - name: Setup yarn
-        run: |
-          curl -o- -L https://yarnpkg.com/install.sh | bash
-          export PATH="$HOME/.yarn/bin:$PATH"
 
       - name: Install Dependencies
         run: yarn --frozen-lockfile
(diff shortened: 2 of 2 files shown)


ChristianMurphy (approved):
LGTM, thanks @JounQin!

.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   default:
-    name: Build and Test on ${{ matrix.os }} with Node.js ${{ matrix.node }}
+    name: Test on ${{ matrix.os }} with Node.js ${{ matrix.node }}
```
ChristianMurphy, line 7: :man_shrugging: it still does build and test https://github.com/mdx-js/mdx/blob/a95e0dd7edcfe9edcdbb8ee5196e2ac9ae91e41e/package.json#L32

JounQin: 😂 This is unobtrusively in the workflow.

wooorm: Can you explain? I tend to agree with @ChristianMurphy

.github/workflows/ci.yml
```diff
@@ -25,11 +25,3 @@ jobs:
...
       - name: Test
         run: yarn test
-      # To do: macos isn’t used anymore
-      - name: Publish CI tag to npm
-        if: matrix.node == 12 && matrix.os == 'macOS-latest' && github.event_name == 'push' && github.ref	== 'refs/heads/master'
-        run: |
-          npm set //registry.npmjs.org/:_authToken $NPM_AUTH_TOKEN
-          yarn publish-ci
-        env:
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
```
wooorm, line 35: Why remove the code for prepublishing, instead of fixing it?

.github/workflows/publish.yml
```diff
@@ -1,23 +1,20 @@
...
+on: [push]
 
 jobs:
   default:
     name: Publish CI tagged release for MDX
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # for lerna
```
wooorm, line 12: This should make checkout out much slower. Why is it needed?

.github/workflows/ci.yml
JounQin: It is in `publish.yml` now.

.github/workflows/publish.yml
JounQin: `lerna` requires commit history

wooorm:
@JounQin Please move slower on your PRs. Give it 48 hours, and address my comments above.

.github/workflows/ci.yml
JounQin: I mean in the workflow file, there is only `yarn test`. `build` is implicit in `package.json`.

Of course, I can add it back if you want.

JounQin:
@wooorm OK, sorry again for that, I'll improve.

.github/workflows/publish.yml
JounQin: https://stackoverflow.com/questions/57597367/how-to-set-up-github-actions-to-publish-a-lerna-monorepo

For `checkout@v1`, it pulls all commits, for `checkout@v2`, it doesn't pull all commits by default.

wooorm: does it require everything all the way back to the start? Could we limit it?

.github/workflows/ci.yml
wooorm: I was assuming that `publish.yml` was publishing tags, and that this file here was publishing canary releases?
Is there still something doing `publish` or `publish-next`?

.github/workflows/ci.yml
wooorm: It doesn’t matter much now... But I would’ve preferred it separately or not at all 🤷‍♂️

.github/workflows/publish.yml
JounQin: I don't know is there any other best practice for it.

Let's keep discussion at https://github.com/mdx-js/mdx/pull/1498#discussion_r601604403

.github/workflows/ci.yml
JounQin: I'd prefer separate too. It actually should be `pretest: yarn build` I think.

.github/workflows/ci.yml
JounQin: Not for now, I don't think you want to publish on CI automatically.

It can be changed to a manual triggered task if you want.

https://github.com/alauda/alauda-ui/blob/master/.github/workflows/publish.yml#L4

Take this for example.

wooorm: ok, sounds good!

PR #1496, opened 2021-03-25
ci: remove .gitattributes file

JounQin:
`lf` is ensured by `.editorconfig` already and it results file difference on build

https://github.com/mdx-js/mdx/runs/2193584817#step:6:15

I don't know if there is a better way to fix this issue. And don't know why types files are effected by building, could be a bug of `microbundle`?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,0 @@
-*.js   text eol=lf
-*.json text eol=lf
-*.jsx  text eol=lf
-*.md   text eol=lf
-*.mdx  text eol=lf
-*.ts   text eol=lf
-*.tsx  text eol=lf


wooorm:
> lf is ensured by .editorconfig already and it results file difference on build

`.editorconfig` is about editors. `.gitconfig` is about Git. They’re different things.

> And don't know why types files are effected by building, could be a bug of microbundle?

`.d.ts` files are generated by typescript (not `microbundle`).

Might adding `.d.ts` to `.gitignore` be a better solution?

JounQin:
> .d.ts files are generated by typescript (not microbundle).

I checked the build script, only `microbundle` is called.

I agree we need to find a better solution.

JounQin:
It is also weird that I cannot reproduce the CI error anymore after removing `.gitattributes` and reverting it back...

ChristianMurphy:
> I cannot reproduce the CI error anymore

When checking this locally I saw the same thing.
I was able to reproduce the error once running `yarn test` at the root level.
But haven't been able to replicate it again, or narrow down which command caused the d.ts files to change. :thinking: :detective:

JounQin:
After digging deep, I found out that it was caused by [microbundle](https://github.com/developit/microbundle/blob/master/src/index.js#L369)

See also https://github.com/developit/microbundle/issues/829

PR #1498, opened 2021-03-25
ci: simplify ci workflows, change master to main on ci

JounQin:
I've reset and reopen this PR.

Blocker: types are incompatible with `lf` setting after building on CI

https://github.com/mdx-js/mdx/pull/1496#issuecomment-806891754

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,8 +23,11 @@ jobs:
       - name: Build
         run: yarn build
 
+      # workaround for types changes after building
       - name: Reset git status
-        run: git reset --hard # workaround for types changes after building
+        run: |
+          git status
+          git reset --hard
 
       - name: Publish CI tag to npm
         if: github.ref	== 'refs/heads/main'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Build
         run: yarn build
 
+      - name: Reset git status
+        run: git reset --hard # workaround for types changes after building
+
       - name: Publish CI tag to npm
         if: github.ref	== 'refs/heads/main'
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -1,23 +1,20 @@
...
+on: [push]
 
 jobs:
   default:
     name: Publish CI tagged release for MDX
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # for lerna
```
JounQin, line 12: https://github.com/actions/checkout/issues/118#issuecomment-806965107

https://github.com/lerna/lerna/issues/2446#issuecomment-585092825

wooorm (approved):

.github/workflows/publish.yml
```diff
@@ -26,8 +23,11 @@ jobs:
       - name: Build
         run: yarn build
 
+      - name: Reset git status
+        run: git reset --hard # workaround for types changes after building
```
JounQin, line 27: @wooorm @ChristianMurphy I add this as workaround temporarily before we determine what's the problem with types after building, how do you think?

wooorm: Maybe add a link in the comment to the action job that had these LF errors?

Otherwise looks good

JounQin: > Maybe add a link in the comment to the action job that had these LF errors?

The action job log will be deleted after days. Or add `git status` command before resetting so that we can always understand what is happening.

wooorm: ah, true, never mind!

wooorm (approved):

ChristianMurphy:
It looks like the publish step is getting stuck after this change, or perhaps it surfaced an issue that was hidden before :thinking: 
https://github.com/mdx-js/mdx/runs/2246873788#step:7:278

PR #1514, opened 2021-04-06
 fix: lint issues, add types back

JounQin:
related to https://github.com/developit/microbundle/issues/829


packages/mdx/types/index.d.ts
```diff
@@ -11,35 +11,35 @@ declare namespace mdx {
     filepath?: string
 
     /**
-     * skip the addition of 'export default' statement when serializing
+     * Skip the addition of 'export default' statement when serializing
```
ChristianMurphy, line 14: Is this linting related?
I'm not seeing a related lint rule in the configuration

packages/loader/types/index.d.ts
```diff
@@ -1,8 +1,8 @@
 // TypeScript Version: 3.4
 
-import {ComponentType} from 'react'
-
+// tslint:disable-next-line: no-single-declare-module
 declare module '*.mdx' {
+  import {ComponentType} from 'react'
```
ChristianMurphy, line 5: out of curiosity, what is the benefit of moving the import inside the `declare module`?

package.json
```diff
@@ -39,15 +39,16 @@
     "@babel/preset-react": "^7.12.0",
     "babel-loader": "^8.2.0",
     "babel-preset-gatsby-package": "^0.12.0",
-    "dtslint": "^4.0.0",
-    "eslint": "^7.13.0",
-    "eslint-config-prettier": "^7.0.0",
-    "eslint-config-xo": "^0.33.0",
+    "dtslint": "^4.0.8",
+    "eslint": "^7.23.0",
```
ChristianMurphy, line 43: these updates appear to already be/have been in semver range?

ChristianMurphy (approved):
Thanks @JounQin!

package.json
JounQin: When debug an issue related to duplicate versions of `@typescript-eslint/eslint-plugin` due to `gatsby`, I tried to upgrade them to latest versions manually.

packages/mdx/types/index.d.ts
JounQin: https://eslint.org/docs/rules/capitalized-comments

Maybe xo enables it, I'm not sure.

packages/loader/types/index.d.ts
JounQin: It is better to write declarations separately for each `module` IMO. Besides, the `disable` is not related.

PR #1530, opened 2021-04-28
[DOCS] Remove MDX Conf 2020 banner

ChrisChinchilla:
Sorry that banner for a conference in 2020 was always annoying me, I have no idea if this is the right way, but I hope so!


johno:
Thank you!

PR #1566, opened 2021-05-30
feat(@mdx-js/vue): mdx for vue 3

Amour1688:


Amour1688:
@wooorm Please review，thanks.  And I can not run vue-jest in this repo.

image

examples/vue-loader/package.json
```diff
@@ -8,11 +8,10 @@
   },
   "dependencies": {
     "core-js": "^3.6.0",
-    "vue": "^2.6.0"
+    "vue": "^3.0.0"
```
ChristianMurphy, line 11: Vue 2 is still supported by the Vue team

> Every time a new major is released, the last minor in the previous major automatically becomes LTS for 18 months, receiving bug fixes and security patches. Then it becomes maintenance mode (security patches only) for another 18 months before entering end of life.

https://github.com/vuejs/roadmap#release-channels-and-lts

Would it be possible to allow `^2.0.0 || ^3.0.0`?

Amour1688: Vue 3 has different virtual dom with  2.  We'd better support Vue 3 only.

wooorm: I agree that Vue 2 is quite hard to combine with MDX and can be dropped.
Vue 3 added a) fragments and b) functional components, which are important for MDX (I did the same with [xdm for vue](https://github.com/wooorm/xdm#vue))

PR #1572, opened 2021-06-03
Fixes syntax error in a sample code

suin:


ChristianMurphy (changes requested):
Thanks @suin!
The docs are generated from the `main` branch.
When you have a chance could you retarget/rebase the PR on that branch?
It also looks like the linter/formatter is unhappy,  running`yarn run format` and committing the change should automatically resolve that.

wooorm:
Thanks for taking the time to work on this, @suin. As work happens on `main`, I’ve instead opened GH-1647 to land this there.

Issue #1578, opened 2021-06-16
jsx approved image alt text


glitteringkatie:
#### tl;dr quotes in image alt text break babel because they don't get appropriately turned into jsx expressions ([since jsx strings don't support `\"` in them](https://github.com/evanw/esbuild/issues/959))

Hi! I found that if someone uses quotation marks in their image alt (`![something really "cool"](https://www.humanesociety.org/sites/default/files/styles/1240x698/public/2020-07/kitten-510651.jpg)`), the outputted jsx is `<img src="https://www.humanesociety.org/sites/default/files/styles/1240x698/public/2020-07/kitten-510651.jpg" alt="something really \"cool\"" parentName="p" />` Which causes babel to blow up because having escaped " marks in a JSX string is not supported turns out! [This comment](https://github.com/evanw/esbuild/issues/959#issuecomment-797714085) really helped me figure that out. So! I think the solution is to make sure that if an image alt has quotes, those should appropriately be wrapped in a jsx expression `{}`.

Disclaimers: I _think_ this should be on MDX since remark is agnostic to jsx. However, I noticed this wasn't an issue with the link text (Like `[link text "here"](blah.com)`) and that gets handled just fine (aka gets wrapped in appropriate curly braces)--unsure if that was because of something in this project or in remark.
### Your environment

*   **OS**: MacOS
*   **Packages**: 
    * "@mdx-js/mdx@2.0.0-next.9"
    * "@mdx-js/react@2.0.0-next.9"
    * "@mdx-js/util@2.0.0-next.8"
*   **Env**
    * "@babel/core@7.13.14"
    * "@babel/preset-env@7.13.12"
    * "@babel/preset-react@7.13.13"

### Steps to reproduce

1. have an mdx file with `![something really "cool"](https://www.humanesociety.org/sites/default/files/styles/1240x698/public/2020-07/kitten-510651.jpg)` in it somewhere
2. parse that file with mdx
3. put that parsed content through babel (`transformAsync` from babel core with presetReact & presetEnv)

🎉 BONUS POINTS for creating a [minimal reproduction](https://stackoverflow.com/help/mcve) and uploading it to GitHub.  This will get you the fastest support.  🎉

### Expected behaviour

Babel is happy and outputs functioning code!

### Actual behaviour

Babel is sad and says "SyntaxError: unknown: Expecting Unicode escape sequence \uXXXX"

### Work around

[...]

PR #1580, opened 2021-06-18
Encode alt text appropriately

glitteringkatie:
Closes GH-1578

This literally copies @wooorm's edits from XDM into this project, as discussed in [the comments](https://github.com/mdx-js/mdx/issues/1578#issuecomment-862820323) of the issue for this. I tried adding a test but given prettier doesn't like unnecessary character codes & the test data wouldn't have the escaped quotes in it (because prettier wants "\"" to be '"'), it was hard to figure out how to write a good test (well, less prettier's issue and more so prettier showing me why I can't write a good test because all of those things would still be equal). I checked XDM and either I didn't look in the right place or there wasn't a test for that either. So here's some code!

### DIFF
--- a/packages/gatsby-theme-mdx/src/components/layout.js
+++ b/packages/gatsby-theme-mdx/src/components/layout.js
@@ -1,5 +1,4 @@
 import React, {useState, useEffect} from 'react'
-import {Link} from 'gatsby'
 import {Global} from '@emotion/core'
 import {ThemeProvider, css} from 'theme-ui'
 
--- a/packages/mdx/estree-to-js.js
+++ b/packages/mdx/estree-to-js.js
@@ -30,7 +30,16 @@ function JSXAttribute(node, state) {
 
   if (node.value != null) {
     state.write('=')
-    this[node.value.type](node.value, state)
+
+    // Encode double quotes in attribute values.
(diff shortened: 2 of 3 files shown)


ChristianMurphy (approved):
Thanks @glitteringkatie! :sparkles: 

> but given prettier doesn't like unnecessary character codes & the test data wouldn't have the escaped quotes in it (because prettier wants """ to be '"'), it was hard to figure out how to write a good test

yeah, prettier can get in the way sometimes, particularly around edge cases.
It can be disabled for specific blocks with https://prettier.io/docs/en/ignore.html#javascript when it causes problems.

glitteringkatie:
Added a test--confirmed it fails when I revert my change--and did a quick removal of an unused import because running yarn test complained about it 🧹

ChristianMurphy (approved):

glitteringkatie:
What's the best way for me to follow along and see when this gets released?

ChristianMurphy:
To know when releases happen. I'd use the "Watch" button at the top of the repository, it should be a dropdown, under "custom" it should allow watching just for releases.

For following the release process and getting an estimate on when a release may happen, that's something which @mdx-js/releasers may be able to comment on.

PR #1593, opened 2021-07-11
Fix example

mvasilkov:
Under "Use MDX for .md files", the regex should match both .md and .mdx

### DIFF
--- a/docs/getting-started/next.mdx
+++ b/docs/getting-started/next.mdx
@@ -35,7 +35,7 @@ The Next.js MDX plugin allows for you to also use MDX parsing for `.md` files:
 
 ```js
 const withMDX = require('@next/mdx')({
-  extension: /\.mdx$/
+  extension: /\.mdx?$/
 })
 
 module.exports = withMDX({


ChristianMurphy (approved):
thanks @mvasilkov!

wooorm:
Why should that be the case?

The question mark used to be there, and was removed here: https://github.com/mdx-js/mdx/commit/6f3dc9cb359fdb1ae29fb4cc898583f6b8e3b8b6. The reason behind it is that `.mdx` and `.md` are not the same. I think it’s a good default to only support `.mdx` by default. If you want `.md` file too, you are of course free to do so.

ChristianMurphy:
The section this titled

```markdown
### Use MDX for `.md` files

The Next.js MDX plugin allows for you to also use MDX parsing for `.md` files:
```

this PR updates the example to match the section description.
It could be that this section should be removed entirely as an alternative.
Thoughts? :thought_balloon:

wooorm (commented):
disregard my previous comment, nice catch!

PR #1600, opened 2021-07-28
Add missing react peer dep to the webpack loader (v1)

Methuselah96:
This is a resubmittal of https://github.com/mdx-js/mdx/pull/1460 for v1. I know this is done for v2, but it's unclear when v2 will get released and it might take a while for it to get integrated into Storybook, so it would be great to just get it fixed in v1 as well.


wooorm:
Sorry for the long wait, appreciate your patience.

The loader is now agnostic to frontend frameworks. That means it can be used with React, Emotion, Preact, Vue, whatnot. As a result, it no longer makes sense to add a React peer dep.
For a solution: make sure you install a recent React yourself, if that’s what you’re using the loader with!

PR #1606, opened 2021-08-08
Added a component snippet

tamojit-123:


wooorm:
Thanks for your patience.
I don’t think this is a great example to add. For example, it doesn’t explain at all where `./CodeBlock` is coming, why `unified-ui`, or what’s going on in the code and why it’s useful?

PR #1615, opened 2021-08-19
Fix TypeScript error: Cannot find name 'Fragment'.

jablko:
Add `import { Fragment } from "react"`, missed in [#1514](https://github.com/mdx-js/mdx/pull/1514/files#diff-c749c3b78a21e2eb84c84458035416c6cc15f688865ea2097c49fe3e5cb68464R68) :heart:

### DIFF
--- a/packages/loader/types/tsconfig.json
+++ b/packages/loader/types/tsconfig.json
@@ -3,7 +3,6 @@
     "module": "commonjs",
     "lib": ["dom", "es6"],
     "strict": true,
-    "skipLibCheck": true,
     "baseUrl": ".",
     "jsx": "react",
     "paths": {
--- a/packages/preact/types/index.d.ts
+++ b/packages/preact/types/index.d.ts
@@ -1,6 +1,13 @@
-// TypeScript Version: 3.4
+// TypeScript Version: 3.5
 
-import {h, Context, AnyComponent, Fragment, FunctionComponent} from 'preact'
+import {
+  h,
+  Context,
+  AnyComponent,
+  Fragment,
+  FunctionComponent,
+  VNode
+} from 'preact'
 
 /**
  * Mapping of names for JSX components to React components
@@ -52,7 +59,7 @@ declare function useMDXComponents(
  */
 declare function withMDXComponents(
   child: AnyComponent<ComponentsProp>
-): ReactElement | null
+): VNode | null
 
 /**
  * Preact hyperscript function wrapped with handler for MDX content
--- a/packages/react/types/index.d.ts
+++ b/packages/react/types/index.d.ts
@@ -2,8 +2,8 @@
 
 import {
   Context,
-  Consumer,
   ComponentType,
+  Fragment,
   FunctionComponent,
   ReactElement,
   createElement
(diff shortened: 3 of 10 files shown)


packages/react/types/index.d.ts
```diff
@@ -4,6 +4,7 @@ import {
   Context,
   Consumer,
   ComponentType,
+  Fragment,
```
jablko, line 7: Fixes:
```Shell
~/mdx/packages/react $ yarn test
Error: /home/nottheoilrig/mdx/packages/react/types/index.d.ts:68:20
ERROR: 68:20  expect  TypeScript@4.2 compile error: 
Cannot find name 'Fragment'.

File: packages/loader/types/tsconfig.json [hunk 2]
```diff
@@ -3,7 +3,6 @@
     "module": "commonjs",
     "lib": ["dom", "es6"],
     "strict": true,
-    "skipLibCheck": true,
```
jablko, line 6: Stop omitting TypeScript errors from automated tests: `sed --in-place /skipLibCheck/d **/tsconfig.json`

packages/preact/types/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// TypeScript Version: 3.4
+// TypeScript Version: 3.5
```
jablko, line 1: Fixes:
```Shell
~/mdx/packages/preact $ yarn test
Error: Errors in typescript@3.4 for external dependencies:
../../../node_modules/preact/src/jsx.d.ts(310,6): error TS2304: Cannot find name 'Omit'.
```

Preact depends on TypeScript 3.5: https://github.com/preactjs/preact/pull/2581

ChristianMurphy (approved):
Thanks @jablko!

packages/loader/types/tsconfig.json
JounQin: This option enabled makes ts ignoring errors in `.d.ts`, right? But what if there are some errors from `node_modules`?

jablko: > This option enabled makes ts ignoring errors in `.d.ts`, right?

Yup.

> But what if there are some errors from `node_modules`?

If a way exists to verify our `.d.ts` without also checking declarations in `node_modules`, I don't know it ... :thinking: Could we use our dependencies/devDependencies and semvers to keep broken types out of `node_modules`? If that works, it would probably be the friendliest thing to our downstream types consumers?

JounQin: OK, maybe we can raise a feature request for TypeScript?

JounQin (approved):

packages/loader/types/tsconfig.json
jablko: I ran into an issue: Since dropping `skipLibChecks` we're checking the following `.d.ts`:
- our explicit `node_modules` dependencies
- every `node_modules/@types` declaration :joy:

https://github.com/mdx-js/mdx/pull/1622 add [`"types": []`](https://www.typescriptlang.org/tsconfig#types), to at least only check our explicit dependencies.

Issue #1172, opened 2020-07-22
Only run test-types once, perhaps as part of its own GitHub Action


johno:
`yarn test-types` appears to be a bit flaky when run in CI (likely as a result of our large test matrix) and causes builds to fail. Perhaps it makes sense to run it in its own action ~~and maybe mark as optional~~?

cc/ @ChristianMurphy @JounQin

ChristianMurphy:
I'd prefer to keep it enabled if possible, type tests are very helpful for verifying the correctness typings.
Having it in a separate job would be okay if desired.
I'm more interesting in finding/fixing what is causing the flakiness.

johno:
Cool, that works for me! Maybe to start we can make it a single action running on ubuntu + node v14 if we can't get to the bottom of the flakiness?

AFAICT it seems to request dependencies and install stuff on the fly which might not be resilient enough to GitHub Action network flapping 🤔

ChristianMurphy:
#1178 has one possible way of addressing this, based of the dtslint readme:

> Package lock files such as yarn.lock and package-lock.json may cause this issue because of our github dependency on "definitelytyped-header-parser": "github:Microsoft/definitelytyped-header-parser#production", which contains the list of typescript versions to install. To fix this, try deleting your lock file and re-installing.

https://github.com/microsoft/dtslint#faq

if the issue persists it may be related to https://github.com/microsoft/dtslint/issues/267

johno:
Gonna push the test out to its own (required) action for now to wait to check stability and then try to figure out what's going wrong with the windows tests.

Thanks for the quick response and PR @ChristianMurphy 🙏

PR #1622, opened 2021-08-28
Add `test-types` script back

jablko:
I ran into *an* issue with `test-types` which this PR solves by adding [`"types": []`](https://www.typescriptlang.org/tsconfig#types) -> `**/tsconfig.json`. If it was *the* issue with `test-types`, perhaps it can be restored -> the CI?

#1179 temporarily moved `test-types` -> its own `types.yml` workflow, suggesting it be moved back -> `test` once it's running stably:
> This will be a temporary change until we see dtslint running stably, then we can move `test-types` back under the `test` command.

`types.yml` was subsequently eliminated in #1338.

The issue I encountered was dtslint implicitly including and checking every `node_modules/@types` declaration:
```
Error: Errors in typescript@3.5 for external dependencies:
../../../node_modules/@types/prettier/index.d.ts(537,14): error TS2456: Type alias 'Doc' circularly references itself.

    at /home/nottheoilrig/mdx/node_modules/dtslint/bin/index.js:206:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nottheoilrig/mdx/node_modules/dtslint/bin/index.js:6:58)
```

This PR solves that by adding `"types": []` -> `**/tsconfig.json`:
```Shell
$ for tsconfig in $(git ls-files '**/tsconfig.json'); do
  git show ":$tsconfig" | jq '.compilerOptions.types = []' > "$tsconfig"
done
```

This will still type check our explicit `node_modules/@types` dependencies, however it will [no longer implicitly include every `node_modules/@types` declaration](https://github.com/mdx-js/mdx/pull/1615#discussion_r692615204).

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "test-api": "lerna run test-api",
     "test-coverage": "lerna run test-coverage",
     "test-types": "lerna run test-types",
-    "test": "yarn build && yarn lint && yarn test-coverage"
+    "test": "yarn build && yarn lint && yarn test-coverage && yarn test-types"
   },
   "dependencies": {},
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "publish-next": "lerna publish --force-publish=\"*\" --pre-dist-tag next --preid next",
     "test-api": "lerna run test-api",
     "test-coverage": "lerna run test-coverage",
-    "test-types": "lerna run test-types",
+    "test-types": "lerna --concurrency 1 run test-types",
     "test": "yarn build && yarn lint && yarn test-coverage && yarn test-types"
   },
   "dependencies": {},
--- a/packages/loader/types/tsconfig.json
+++ b/packages/loader/types/tsconfig.json
@@ -8,6 +8,7 @@
     "jsx": "react",
     "paths": {
       "@mdx-js/react": ["index.d.ts"]
-    }
+    },
+    "types": []
   }
 }
--- a/packages/mdx/types/tsconfig.json
+++ b/packages/mdx/types/tsconfig.json
@@ -8,6 +8,7 @@
     "jsx": "react",
     "paths": {
       "@mdx-js/mdx": ["index.d.ts"]
-    }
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -28,8 +28,8 @@
     "publish-next": "lerna publish --force-publish=\"*\" --pre-dist-tag next --preid next",
     "test-api": "lerna run test-api",
     "test-coverage": "lerna run test-coverage",
-    "test-types": "lerna run test-types",
-    "test": "yarn build && yarn lint && yarn test-coverage"
+    "test-types": "lerna --concurrency 1 run test-types",
```
jablko, line 31: [Concurrent dtslint invocations](https://github.com/lerna/lerna/blob/main/core/global-options/README.md#--concurrency) were trampling each other installing TypeScript versions ... At least I *think* that's what was happening [here](https://github.com/mdx-js/mdx/pull/1622/checks?check_run_id=3451812571#step:5:373) -- and [here](https://github.com/mdx-js/mdx/issues/1172) ... Perhaps [here](https://github.com/microsoft/dtslint/issues/267) too ...

ChristianMurphy (approved):
Sounds alright to me.
Thoughts on this approach @JounQin?

JounQin:
I'm not sure, personally I'd prefer something like `Expect` type utility for `types.test.ts`, and run `tsc --noEmit` for type checking for all `.ts` related files.

But for multiple ts versions, `dtslint` maybe better.

However, I don't think targeting multiple ts versions are worth unless we're not going to always follow latest TypeScript.

jablko:
@JounQin I think most of this change applies equally to `tsc --noEmit` and dtslint: Without `"types": []` we'll implicitly include every `node_modules/@types` declaration in either case, I think?

Can we land this, and roll back the `lerna --concurrency 1` option if/when we drop dtslint?

JounQin (approved):
OK, whether to drop `dtslint` can be discussed later, let's fix current issues first.

ChristianMurphy (approved):

PR #1626, opened 2021-09-02
Update Parcel guide

### LINKED ISSUE #1142: Parcel implementation not working
### Subject of the issue

Parcel demo not working (https://mdxjs.com/getting-started/parcel)
### Your environment
Mac
### Steps to reproduce
Install parcel and mdx.

Create an index.mdx file with ```# test this```

run ```parcel index.mdx```

### Expected behaviour
Should render
### Actual behaviour
Doesn't render

(I'm new to both parcel and mdx)

jablko:
What do you think of this revision to the Parcel guide and example for [Parcel 2](https://v2.parceljs.org/), and deprecating `parcel-plugin-mdx` in favor of [@parcel/transformer-mdx](https://www.npmjs.com/package/@parcel/transformer-mdx)?

I think this fixes #1142.


.eslintrc.yml
```diff
@@ -55,6 +55,7 @@ overrides:
     rules:
       import/no-extraneous-dependencies: 0
       prettier/prettier: error
+      react/no-unescaped-entities: off
```
jablko, line 58: Not sure what's going on here:
```
  ✘  https://google.com/#q=react%2Fno-unescaped-entities

     `>` can be escaped with `&gt;`


     docs/getting-started/parcel.mdx:67:1
     65 | [You need to use][type=module] `<script src="hello.js" type="module"></script>` to reference `hello.js` from `index.html`, otherwise you will see the following error:
     66 | 
   > 67 | > @parcel/transformer-js: Browser scripts cannot have imports or exports.
        | ^
     68 | 
     69 | </Note>
     70 | 
```
`>` should be compiled to `<blockquote>`?

ChristianMurphy: Trying it out in https://mdxjs.com/playground it does compile to `<blockquote>` :thinking: 
/cc @JounQin thought or insights?

JounQin: Markdown inside jsx is still considered as jsx child in `eslint-mdx@v1`, the only workaround is extract the markdown part into another smaller mdx file then import it as Component.

wooorm:
@jablko Could you use a `<blockquote>` for now to work around the problem? 🤷‍♂️

wooorm:
And can you rebase?

jablko:
> @jablko Could you use a `<blockquote>` for now to work around the problem?

@wooorm If the syntax works, it's just a matter of the linter, I'd prefer to temporarily disable the ESLint rule ... However I've now moved the disable from the `.eslintrc.yml` to a configuration comment in this file, and added `--report-unused-disable-directives` so it can be re-enabled when working?
https://github.com/mdx-js/mdx/blob/26385a9f7b688c249b64b03d61bf24cf9d4eab58/docs/getting-started/parcel.mdx#L63

> And can you rebase?

:heavy_check_mark: Done.

.eslintrc.yml
jablko: Got it, thanks for this explanation!

wooorm (approved):

wooorm:
Thanks @jablko!

PR #1632, opened 2021-09-08
Update Actions

wooorm:
* Do not run on Node v14
* Use nvm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,28 +1,26 @@
-name: CI
-
-on: [push, pull_request]
-
+name: main
+on:
+  - pull_request
+  - push
 jobs:
-  default:
-    name: Build and Test on ${{ matrix.os }} with Node.js ${{ matrix.node }}
+  main:
+    name: '${{matrix.node}} on ${{matrix.os}}'
+    runs-on: ${{matrix.os}}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dcodeIO/setup-node-nvm@master
+        with:
+          node-version: ${{matrix.node}}
+      - run: npm install -g yarn && yarn --frozen-lockfile
+        env:
+          CI: 'true'
+      - name: Test
+        run: yarn test
     strategy:
       matrix:
         os:
           - ubuntu-latest
           - windows-latest
         node:
-          - 12
-          - 14
-          - 16
-    runs-on: ${{ matrix.os }}
-    steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,28 +1,26 @@
...
-          - 14
-          - 16
-    runs-on: ${{ matrix.os }}
+  main:
+    name: '${{matrix.node}} on ${{matrix.os}}'
+    runs-on: ${{matrix.os}}
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: dcodeIO/setup-node-nvm@master
```
ChristianMurphy, line 11: please pin a specific major version range

```suggestion
      - uses: dcodeIO/setup-node-nvm@v4
```

.github/workflows/publish.yml
```diff
@@ -1,32 +1,23 @@
...
         with:
           fetch-depth: 0 # for lerna
-
-      - uses: actions/setup-node@v2
-        with:
-          node-version: 14
-
-      - name: Install Dependencies
-        run: yarn --frozen-lockfile
+      - uses: dcodeIO/setup-node-nvm@master
```
ChristianMurphy, line 11: please pin a specific major version range

```suggestion
      - uses: dcodeIO/setup-node-nvm@v4
```

.github/workflows/ci.yml
wooorm: Yeah I dunno, I did think about it a second but:

- this specific action didn’t have “moving” tags before, so I recently saw some older packages still on old commit hashes, that now have to be updated
- it’s a relative churn to go update this every year through every package
- it’s unlikely that this small API surface that’s used would change
- if stuff starts breaking, we’re immediately notified of the API change (or repo removal), and then that’s a good time to go through everything at once?

ChristianMurphy: > it’s a relative churn to go update this every year through every package

renovate manages github actions tags as well, and can/will offer PR(s) when there are updates

> it’s unlikely that this small API surface that’s used would change

Think of package.json, even for minor dependencies, a `^` or `~` range would be used, not `*`.

> if stuff starts breaking, we’re immediately notified of the API change (or repo removal), and then that’s a good time to go through everything at once?

I'd rather it go through renovate.
Where there is a PR to review and a changelog right there accompanying it.
Rather than some unlucky contributor gets stuck fixing CI, because it happened to break when they opened a completely unrelated PR.

PR #1633, opened 2021-09-08
Somplify instructions in interactive example.

hebronwatson:
This change simplifies the language in one of the interactive examples.


wooorm:
- Can you please change the `main` branch? `master` is for v1, all work happens on `main`

wooorm:
A similar patch landed on main: https://github.com/mdx-js/mdx/commit/917c9e144004e1e3d958c31a1acf334c7e7f287d#diff-bb8a7c65bef6e52900370260c910d928619d619663437ee4c6edce2f35843f58R109.

I appreciate that you took the time to work on a fix. I’ll assume that this change satisfies your original problem. If not, could you open a new PR pointing to `main` with what you’d suggest? Thanks

PR #1640, opened 2021-09-12
Add new, RSC + MDX based website

wooorm:
Add new, RSC + MDX based website

* uses xdm for now, will switch to mdx-js/mdx when it has the features
* rsc is the future
* extremely optimized implementation, rich metadata, sitemap, rss feed
* improved playground(s)
* no framework, which makes sure we don’t favor one company over another
* currently a lot of new direct dependencies and custom code, which will
  presumably become less so when RSC hits mainstream as either someone
  else will build stuff that does things like this, or perhaps I will
  externalize it.


.eslintrc.yml
```diff
@@ -48,14 +48,13 @@ settings:
...
-      import/no-extraneous-dependencies: 0
-      prettier/prettier: error
-
+  # - files: '*.{md,mdx}'
+  #   extends:
+  #     - plugin:mdx/recommended
+  #     - plugin:mdx/overrides
+  #   rules:
+  #     import/no-extraneous-dependencies: 0
+  #     prettier/prettier: error
```
ChristianMurphy, line 57: why did a bunch of eslint rules need to be disabled?

docs/_asset/root.client.js
```diff
@@ -0,0 +1,15 @@
+import {Suspense} from 'react'
+
+export function Root(props) {
+  const {response} = props
```
ChristianMurphy, line 4:
```suggestion
export function Root({response}) {
```

docs/_component/editor.client.js
```diff
@@ -0,0 +1,215 @@
...
+            {[
+              'Run',
+              'Compile',
+              'mdast (markdown)',
+              'hast (HTML)',
+              'esast (JS)'
+            ].map(label => {
+              const compile = label === 'Compile'
+              const className =
+                'editor-result-tab' + (compile ? ' editor-result-badge' : '')
```
ChristianMurphy, line 98: it is optional, but often nice to have a utility like https://www.npmjs.com/package/clsx to help manage dynamic classname/string-property construction.

.eslintrc.yml
wooorm: /cc @JounQin

This now uses the MDX@2 parsing, which conflicts with eslint-mdx.
I want to add `eslint-mdx` again when it supports the things mdx@next does and will once v2 is released.

docs/_asset/root.client.js
wooorm: I find that your suggestion makes it harder to debug code

.eslintrc.yml
JounQin: Should `remark-lint` be added back for now in this case?

wooorm: Good idea!

Also Joun: there’s still a lot to do here to port xdm into mdx-js/mdx and write a lot of documentation.
I’m hoping to have a release (candidate) of v2 around in 4 weeks.
Do you have questions or need other information that I can help with for `eslint-mdx`?

JounQin: The problem is that I don't have enough time to work on `next` version for these days. 😂

docs/_asset/root.client.js
ChristianMurphy: I respectfully disagree that it is harder to debug.
And would also note that this pattern is used elsewhere in this PR.

wooorm: Two examples where I find it hard to debug:
- error messages, as there's no name of the parameter
- Logging what else gets passed in props, because there is no props

Do you experience these things and deal with them differently?

And whether being inconsistent: yeah probably 😬😅

ChristianMurphy: I usually handle that by dropping a breakpoint in the debugger to see what is going on at that point in time.
And for React components specifically, the react dev tools will show what props are passed in.

wooorm: hmm, I never use either of those

PR #1654, opened 2021-09-14
Replace yarn, lerna with npm

wooorm:
* yarn is having problems hoisting dependencies
* workspaces are builtin to npm now
* npm has `npm ci` now, which should speed things up

Let’s see if this works and is fast in actions


.github/workflows/ci.yml
```diff
@@ -8,14 +8,18 @@ jobs:
...
-      - run: npm install -g yarn && yarn --frozen-lockfile
-        env:
-          CI: 'true'
-      - name: Test
-        run: yarn test
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: npm-${{hashFiles('package-lock.json')}}
+          restore-keys: npm-
```
ChristianMurphy, line 20: if the setup-node action is used, this could be abstracted away https://github.blog/changelog/2021-09-07-github-actions-setup-node-supports-dependency-caching-for-projects-with-monorepo-and-pnpm-package-manager/

.github/workflows/publish.yml
```diff
@@ -7,15 +7,16 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 0 # for lerna
+          fetch-depth: 0
```
ChristianMurphy, line 10: this is still needed?

wooorm:
Initial time seems at roughly the same time. Now let’s see if caching does something?

ChristianMurphy (commented):
:+1: for leveraging npm workspaces

.github/workflows/publish.yml
wooorm: it should at least make cloning much faster. There have been some giant files historicaly in the repo

.github/workflows/ci.yml
wooorm: I’m assuming the `package-lock.json` will change a ton (it’s 1.4mb, so it’s likely something changes in the ecosystem every day), which is why I’m investigating a different cache strategy.
See also https://github.com/bahmutov/npm-install/issues/108.

Oh, I’m just seeing that `setup node` finally supports “nvm lts syntax” (`lts/erbium`), then we can at least switch to that

wooorm (commented):
✅

PR #1660, opened 2021-09-16
Update `@mdx-js/preact`

wooorm:
* Use ESM
* Add JSDoc based types (and improve the types a bunch)
* Mark `MDXContext` and `withMDXComponents` exports as deprecated
  (they still work)
* Remove `mdx` export, which is no longer needed
* Use `xdm` for a second


wooorm (commented):
✅

PR #1669, opened 2021-09-22
Update `@mdx-js/mdx`

wooorm:
* Import xdm’s core into `@mdx-js/mdx`
* Tons of fixes and updates and a couiple of changes that will be tested and documented later
* Use ESM
* Add JSDoc based types


wooorm (commented):
✅

PR #1681, opened 2021-09-26
Remove examples from monorepo

wooorm:
While useful, the examples pulled in a *lot* of frameworks and packages. This made installing, on CI but also for contributing, slow.
It also resulted in “vulnerabilities” found by npm and such in the package lock, which isn’t an actual problem, but did result in renovate/dependabot and so opening PRs, and npm from showing warning on installation.
My plan is to move them to the website instead soon.


ChristianMurphy:
:+1: for moving example out of core

> My plan is to move them to the website instead soon.

One thing that is/was nice, was having the ability to checkout the repo and run the examples locally.
Having it on the website would require copying and pasting, no?
Would it make sense to create an `mdx-examples` repo that people could checkout and run?
Maybe use that as the source for the examples on the website?

wooorm:
> One thing that is/was nice, was having the ability to checkout the repo and run the examples locally.
> Would it make sense to create an mdx-examples repo that people could checkout and run?

Yes, that’s somewhat nice. Also: codesandbox starters. Which can come from repos.
I do wonder if all the frameworks in one repo will run into the same problems as before though.

> Having it on the website would require copying and pasting, no?

Copy/pasting is always needed. 🤷‍♂️
But it depends on what the goal is.
In my opinion the primary use case should be “How to integrate mdx-js/mdx with an existing X project” rather than the current focus of “How to set up a new X project with mdx-js/mdx”.

Let’s keep on discussing examples either here, while I merge it, or in following up docs/website PRs coming the next few weeks.

ChristianMurphy:
> I do wonder if all the frameworks in one repo will run into the same problems as before though.

Only if the repo combines everything into a single workspace.

> Copy/pasting is always needed. :man_shrugging:

To some extent, yes.
But from

> But it depends on what the goal is.
> In my opinion the primary use case should be “How to integrate mdx-js/mdx with an existing X project” rather than the current focus of “How to set up a new X project with mdx-js/mdx”.

Hmmm :thinking: 
I'm not sure that's quite how I'd frame the current goal.
I see it more as something which people can quickly spin up, see that it works, see how it works, and leverage if they want.
Much in the same way that CodeSandboxes/JSFiddle/{other sandbox services} offer a quick way to try things out.

The value comes in more as a working reference point, to help people debug/reproduce issues.

I don't have a strong preference as to whether that is done as a repo, as codesandboxes which are shared, or another quick-to-run method for sharing code.

It is however something which "How to integrate mdx-js/mdx with an existing X project" doesn't cover, they are different, and both have value in different contexts.

wooorm:
> The value comes in more as a working reference point, to help people debug/reproduce issues.

I’m in favor of that. But how big should the samples be? MVP? Real-world example? The first is easier to maintain. The second could also be useful, but that can perhaps be done by linking to how other people use mdx? (perhaps updating: https://github.com/mdx-js/showcase)

PR #1702, opened 2021-09-29
Fix internal types of `run` to allow strings

wooorm:


packages/mdx/lib/run.js
```diff
@@ -1,14 +1,10 @@
...
- */
-
 /** @type {new (code: string, ...args: unknown[]) => Function} **/
 const AsyncFunction = Object.getPrototypeOf(run).constructor
 
 /**
  * Asynchronously run code.
  *
- * @param {VFile} file JS document to run
+ * @param {{toString(): string}} file JS document to run
```
ChristianMurphy, line 7: would it make sense to use `VFileCompatible` here?
https://github.com/vfile/vfile/blob/b4c9fb85e975be2b32e03e47c4bc8a1501155493/lib/index.js#L13-L22

wooorm: Thanks for the suggestion. Maybe. I saw this suggestion somewhere on twitter recently, for something unrelated, and thought it’d be interesting if it worked, and apparently it did. It does at least work for strings and vfiles. I’m assuming it works for buffers too. It probably fails for the object form of vfiles though, which is a good thing and `VFileCompatible` would not?

ChristianMurphy: > It probably fails for the object form of vfiles though, which is a good thing

Would it be?
I thought anything passed to a pipeline would automatically turn into a vFile?

> `VFileCompatible` would not?

it would not

---

if this is intended, is there a more general util which should be added to vfile or documented for plugin creators?

wooorm: `runSync` is a shorthand for `new Function(string)(options)`, and the more complex async version. So it really has to be something that can be cast to a string.
`VFile` was used before because it can be cast to a string. But it was weird. Because VFiles are not the only things that can strings or can be cast to strings.

The tests were also using `run` to evaluate something, but in this case an actual string, hence the type error: https://github.com/mdx-js/mdx/pull/1702/files#diff-81a626c22b0507b168fd4fce4a156f1909045d4bee13daad9e3972d6b64cdbe0L36

PR #1714, opened 2021-10-06
Add community section

wooorm:
- Add a community section, which houses the support, contribute, about, and projects pages
- rewrite support/contribute to be a bit more apt
- update about with a bunch of more info
- update links to not reference `https://mdxjs.com`, but instead be relative to the current domain
- Add editor section to getting started


wooorm (commented):
✅

PR #1745, opened 2021-10-17
Use `@types/mdx`

remcohaszing:
This removes the need to define `complex-types` for `@mdx-js/react` and `@mdx-js/preact`. The exported `Components` types have been removed, because they can also be imported from `mdx/types`,

A `@ts-expect-error` statement in `@mdx-js/node-loader` is no longer needed and has been removed.


wooorm:
Thanks Remco! Glad it works well!

PR #1754, opened 2021-10-19
Add `schema.org` ldjson about language to home page

ChristianMurphy:
leverage https://schema.org and structured data.
Specifically https://schema.org/SoftwareApplication and https://schema.org/ComputerLanguage to annotated the home page so search engines know how MDX relates to other parts of the knowledge graph and can show a smart card.

Currently only the home page is annotated with Software Application.
In the future we may want to add other smart card enabled types https://developers.google.com/search/docs/advanced/appearance/overview, like HowTo, FAQ, Article, and Event.

This can be tested using https://search.google.com/test/rich-results pointing to the home page

### DIFF
--- a/docs/_component/home.server.js
+++ b/docs/_component/home.server.js
@@ -9,6 +9,11 @@ export const Home = (props) => {
     <div className="page home">
       <NavSiteSkip />
       <main>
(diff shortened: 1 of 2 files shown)


wooorm:
sweeeet!

PR #1758, opened 2021-10-20
Add debounce to playground

ChristianMurphy:
this debounces user inputs to allow smoother typing while still updating the preview.


ChristianMurphy:
> Any reason you like ahooks?

A few reasons:
1. It has a lodash like amount of small composable hooks https://ahooks.js.org/hooks
2. Good docs and examples
3. Native typescript support

PR #1803, opened 2021-11-09
Update links to point to `mdxjs.com`

wooorm:


wooorm (commented):
✅

Fast-tracking as the v2 already moved, but the old references break it

wooorm:
…once tests pass pfft

PR #1804, opened 2021-11-09
Update for Node 17

wooorm:
*   Add support for Node 17 to the node loader
    (backport https://github.com/wooorm/xdm/commit/e72b930)
*   Turn of a `preact/compat` test, which throws in Node 17 as its missing an
    export map
*   Turn tests off for Node 17 in `@mdx-js/register`, as it requires `deasync`
    which hasn’t been updated with a Node 17 build yet.

### DIFF
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -149,13 +149,14 @@ test('compile', async () => {
     '\nnote: the next deprecation is expected (preact is missing an export map)\n'
   )
 
-  assert.equal(
-    render(
-      h(await run(compileSync('?', {jsxImportSource: 'preact/compat'})), {})
-    ),
-    '<p>?</p>',
-    'should support an import source (`@jsxImportSource`)'
-  )
+  // To do: re-enable when `preact/compat` has a correct export map.
+  // assert.equal(
+  //   render(
+  //     h(await run(compileSync('?', {jsxImportSource: 'preact/compat'})), {})
+  //   ),
+  //   '<p>?</p>',
+  //   'should support an import source (`@jsxImportSource`)'
+  // )
 
   assert.equal(
     render(
--- a/packages/node-loader/lib/index.js
+++ b/packages/node-loader/lib/index.js
@@ -2,8 +2,9 @@
  * @typedef {import('@mdx-js/mdx/lib/compile.js').CompileOptions} CompileOptions
  */
 
-import path from 'path'
-import {URL} from 'url'
+import {promises as fs} from 'node:fs'
+import path from 'node:path'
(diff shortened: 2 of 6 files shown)


packages/mdx/test/compile.js
```diff
@@ -149,13 +149,14 @@ test('compile', async () => {
...
   )
 
-  assert.equal(
-    render(
-      h(await run(compileSync('?', {jsxImportSource: 'preact/compat'})), {})
-    ),
-    '<p>?</p>',
-    'should support an import source (`@jsxImportSource`)'
-  )
+  // To do: re-enable when `preact/compat` has a correct export map.
```
ChristianMurphy, line 152: Is there an open issue with preact to address this?
If so could it be noted in a comment?

packages/node-loader/lib/index.js
```diff
@@ -15,12 +16,39 @@ import {createFormatAwareProcessors} from '@mdx-js/mdx/lib/util/create-format-aw
...
 
-  return {getFormat, transformSource}
+  return {load, getFormat, transformSource}
 
+  /* c8 ignore start */
+  // Node version 17.
+  /**
+   * @param {string} url
+   * @param {unknown} context
+   * @param {Function} defaultLoad
```
ChristianMurphy, line 26: Does Node have a more specific type for the default loader?

ChristianMurphy (commented):
Nice! :+1:

packages/mdx/test/compile.js
ChristianMurphy: :thinking: it appears to have an export map https://github.com/preactjs/preact/blob/bd52611cade159a643b617794410f40a9d52eda3/package.json#L23
What issue is being encountered?

packages/node-loader/lib/index.js
wooorm: The API for `defaultLoad` is the same as this `load` function.

packages/mdx/test/compile.js
wooorm: The problem is that they don’t list `compat/jsx-runtime` in their export map

wooorm: reported

packages/node-loader/lib/index.js
ChristianMurphy: Could we create a recursive interface for the type?

wooorm: Sure but I don’t see a bunch of value. It just got redesigned. It’ll be redesigned again. The API is unstable: https://nodejs.org/api/esm.html#loadurl-context-defaultload

PR #1805, opened 2021-11-10
Ignore `react-dom`, `react-server-dom-webpack` in renovate

wooorm:
These package are experimental and contain RSC. Switching to a random public react version as proposed by https://github.com/mdx-js/mdx/pull/1790 and such just breaks everything

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,10 @@
 {
   "extends": ["config:base", ":preserveSemverRanges"],
-  "schedule": "before 3am on Monday"
+  "schedule": "before 3am on Monday",
+  "packageRules": [
+    {
+      "matchPackageNames": ["react-dom", "react-server-dom-webpack"],
+      "enabled": false
+    }
+  ]
 }


renovate.json
```diff
@@ -1,4 +1,10 @@
 {
   "extends": ["config:base", ":preserveSemverRanges"],
-  "schedule": "before 3am on Monday"
+  "schedule": "before 3am on Monday",
+  "packageRules": [
+    {
+      "matchPackageNames": ["react-dom", "react-server-dom-webpack"],
+      "enabled": false
```
ChristianMurphy, line 7: Would it make sense to leverage https://docs.renovatebot.com/configuration-options/#allowedversions instead?
So we can prevent updates for unwanted versions, but still potentially get future updates

wooorm: I’m open to better ways, I don’t know renovate.
But react with RSC is on 0.0.0-experimental. All other versions do not have RSC and break

wooorm: Know of a way to get around this? Or is it fine?

ChristianMurphy (approved):

PR #1807, opened 2021-11-10
Update `loader-utils` in `@mdx-js/loader`

wooorm:
This updates `loader-utils`, which moves a method from it to webpack itself, meaning `loader-utils` can be removed again.

This commit backports: <https://github.com/wooorm/xdm/commit/a574477>

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -6,11 +6,10 @@
  */
 
 import {SourceMapGenerator} from 'source-map'
-import {getOptions} from 'loader-utils'
 import {compile} from '@mdx-js/mdx'
 
 /**
- * A Webpack (4+) loader for MDX.
+ * A Webpack (5+) loader for MDX.
  * See `webpack.cjs`, which wraps this, because Webpack loaders must currently
  * be CommonJS.
  *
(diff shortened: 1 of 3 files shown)


packages/loader/lib/index.js
```diff
@@ -6,11 +6,10 @@
  */
 
 import {SourceMapGenerator} from 'source-map'
-import {getOptions} from 'loader-utils'
 import {compile} from '@mdx-js/mdx'
 
 /**
- * A Webpack (4+) loader for MDX.
+ * A Webpack (5+) loader for MDX.
```
ChristianMurphy, line 12: Does this mean webpack 4 is no longer supported?
This may be okay long term, short term this would be problematic for CRA adopters (pending https://github.com/facebook/create-react-app/discussions/11278 going GA)

ChristianMurphy (approved):

packages/loader/lib/index.js
wooorm: Yes
I’m not sure the RC / xdm works with webpack 4 anyway?

wooorm: I have recently done some stuff locally though. Might be CodeSandbox acting up? 🤷‍♂️

ChristianMurphy (approved):

packages/loader/lib/index.js
ChristianMurphy: Could be?
I spun up a fresh sandbox for these to try to avoid dependency caching issues, but it may still be an issue :thinking:

wooorm: Well if webpack 4 doesn’t seem to work now anyway, then this isn’t a breaking change.
We can see when folks try the RC out / try it out ourselves locally and go from there later?

PR #1815, opened 2021-11-14
Fix broken link

rodrez:
**Issue**: toc url at https://mdxjs.com/docs/using-mdx/#components was returning 404 Not Found.
**Reason**: link was pointing to the wrong URL
**Deatails**: Clicking on the URL was returning 404 Not found as it was pointing towards `docs/table-of-components/` instead of `table-of-components/`
**Solution**: Removed `docs/` from the URL in the respective documentation file.


wooorm:
Thanks @rodrez!

Issue #1774, opened 2021-10-24
MDXProvider components not word


keenwon:
### Subject of the issue

MDXProvider `components` not word

This is *not* a place to ask questions.  For that, go to [Discussions](https://github.com/mdx-js/mdx/discussions)

### Your environment

*   **OS**: Linux
*   **Packages**: 
    * "@mdx-js/loader": "^2.0.0-rc.1"
    *  "@mdx-js/mdx": "^2.0.0-rc.1"
    *  "@mdx-js/react": "^2.0.0-rc.1"
*   **Env**:
    * node 14.18.1
    * npm 6.14.15

### Steps to reproduce

test repo:  https://github.com/keenwon/mdx-test

```
npm install
npm start
```

### Actual behaviour

MDXProvider component not work

PR #1837, opened 2021-11-28
Add providerImportSource instructions in React docs

citypaul:
Hi there,

Thanks for this great lib!

I had an issue just now when trying to use the `MDXProvider` in a next js app. I had a custom webpack config setup but the components passed to the provider were having no effect.

I found this issue, which provided the answer for me: https://github.com/mdx-js/mdx/issues/1774. However, the React docs do not mention this, and because someone else had clearly had the same issue previously, I figured it would make sense to add this information into the docs.

I also spotted a broken link in the same part of the docs I was editing (the link to `use-provider` was broken), so I fixed that too.

### DIFF
--- a/packages/react/readme.md
+++ b/packages/react/readme.md
@@ -86,8 +86,13 @@ directly:
 
 See [¶ React in § Getting started][start-react] for how to get started with MDX
 and React.
+
 See [¶ MDX provider in § Using MDX][use-provider] for how to use an MDX
 provider.
+
+If using a custom webpack setup and you wish to use an MDX provider,
+make sure to setup the `providerImportSource` in your webpack config.
+See [the migration docs for an example][mdx-loader-migration-guide].
 
 ## API
 
@@ -221,8 +226,10 @@ abide by its terms.
 
 [start-react]: https://mdxjs.com/getting-started/#react
 
-[use-provider]: https://mdxjs.com/using-mdx/#mdx-provider
+[use-provider]: https://mdxjs.com/docs/using-mdx/#mdx-provider
 
 [security]: https://mdxjs.com/getting-started/#security
 
 [typescript]: https://www.typescriptlang.org
+
+[mdx-loader-migration-guide]: https://mdxjs.com/migrating/v2/#mdx-jsloader


wooorm:
Hi! Thanks for your contribution!

I think it’s important to note that providers are not required. Providers are useful when you’re importing MDX files inside other MDX files. They’re not needed in other cases. That’s why it’s no longer on by default: you can pass components, no need to provide them. See the article on [Using MDX](https://mdxjs.com/docs/using-mdx/#mdx-provider) for when it’s useful and what the alternative is.

I do agree that documenting how to configure the provider in esbuild/webpack/rollup integrations here makes sense. But it should be agnostic to webpack. And, similar docs should also exist for the preact and vue providers.

I don’t think we should link to the migration docs for `providerImportSource`. We can link to the migration docs in each package, sure, but there are multiple things in multiple packages to migrate, not just this option. So if we care about migration in each package, then it should be linked from multiple places, and should be agnostic to this one particular option.

---

So the above are some ideas on how this should be properly approached for all packages and what this PR is lacking. I’m not saying that you should do everything. What do you think on my above points as a migrating user?

citypaul:
Ok, so to give a bit more context about how I was confused as a newcomer to the library - What I was trying to do was to bring the provider in at the top level of my next js app inside `App.jsx`. My thinking was that then I can bring in any custom styled components used inside my app to create a template for all mdx pages.

I started by using the mdx loader inside my custom webpack config, and was then surprised when I tried to pull in the react package as the MDXProvider didn't appear to work as expected.

It wasn't clear to me from reading the React docs that I would need to set the `providerImportSource` in the loader.

I'm clearly not an expert with this lib as I only stumbled upon it for the first time this afternoon, so I figure just explaining how it was confusing to me as a newcomer might help you to tweak these docs appropriately. If I'm reading the React docs, it should probably mention how to setup the MDX Provider with the loader package somewhere, or at least point to the part of the docs that does explain this.

I understand what you're saying about nested components, but I guess the mental model I had when reading through the docs was that if I'm looking for a single place to pass custom components for all my mdx pages, the MDXProvider was the place to look for that.

wooorm:
Quite understandable

> If I'm reading the React docs, it should probably mention how to setup the MDX Provider with the loader package somewhere, or at least point to the part of the docs that does explain this.

That happens right before your edit! (`See MDX provider in § Using MDX`).
I’ve updated the code examples and added another link to that page.

>  the MDXProvider was the place to look for that.

It can. It doesn’t have to (note: If you’re using Next, it’s required, but that’s Next’s choice)
It’s powerful but it makes your code much much much slower, it costs 3 lines of code instead of 1 line, and it doesn’t do anything: https://mdxjs.com/docs/using-mdx/#mdx-provider

Issue #1821, opened 2021-11-18
@mdx-js/esbuild not resolving relative paths


DCsunset:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

@mdx-js/esbuild: 2.0.0-rc.2

### Steps to reproduce

Create the following directory structure:

```
index.mdx
pages/
    page.mdx
    data.js
```

Suppose `index.mdx` includes `import Component from "./pages/page.mdx"`,
and `pape.mdx` includes `import data from "./data.js"`.

Then use `@mdx-js/esbuild` to build:

```js
import esbuild from 'esbuild'
import mdx from '@mdx-js/esbuild'

await esbuild.build({
  entryPoints: ['index.mdx'],
  outfile: 'output.js',
  bundle: true,
  format: 'esm',
  plugins: [mdx({})]
})
```

### Expected behavior

It's expected to bundle them without any error.

### Actual behavior

An error occurs:

```
pages/page.mdx:3:17: error: Could not resolve "./data.js"
    3 │ import data from "./data.js";
```

### Runtime

Node v16

### Package manager

npm v8

### OS

Linux

### Build and bundle tools

esbuild

PR #1854, opened 2021-12-04
Fix resolve base in esbuild loader

wooorm:
Closes GH-1821.
Related-to: wooorm/xdm@db74ddc

### DIFF
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -212,7 +212,7 @@ export function esbuild(options = {}) {
       assert(file.dirname, 'expected `dirname` to be defined')
 
       // V8 on Erbium.
-      /* c8 ignore next 2 */
+      /* c8 ignore next 9 */
       return {
         contents: value,
         errors,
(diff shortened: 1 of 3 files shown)


packages/esbuild/lib/index.js
```diff
@@ -13,7 +13,9 @@
  * @typedef {ProcessorOptions & {allowDangerousRemoteMdx?: boolean}} Options
  */
 
+import assert from 'node:assert'
 import {promises as fs} from 'fs'
```
ChristianMurphy, line 17: Would it make sense for `fs` to also have a `node:` prefix?

ChristianMurphy (approved):

packages/esbuild/lib/index.js
wooorm: Good point, I’ll remove the `node:` prefixes.
There was some reason why I didn’t apply them here (they are in xdm). Perhaps due to CRA or Next or so?

wooorm: hmm, it’s a bit inconsistent now anyway. I’ve made it more consistent. esbuild and the node loader probably run in a Node version that support it. I’ve “downgraded” one instance of MDX core though.

PR #1894, opened 2022-01-09
Update syntax-highlighting.server.mdx

godmar:
add a note about the necessity of ESM for rehype.

### DIFF
--- a/docs/guides/syntax-highlighting.server.mdx
+++ b/docs/guides/syntax-highlighting.server.mdx
@@ -80,8 +80,8 @@ async function main(code) {
 
 <Note type="important">
   **Important**: `rehype-highlight` requires the use of ESM and will not work with CJS.
-  This means that if using `next.js`, you will need a `next.config.mjs` instead of a `next.config.js`
-  file for your configuration.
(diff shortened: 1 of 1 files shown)


ChristianMurphy (commented):
This generally applies to all remark and rehype plugins. (Since https://github.com/unifiedjs/unified/issues/121)
Not just `rehype-highlight`, it may make sense to add the note to https://mdxjs.com/docs/getting-started/#nextjs

godmar:
This is the first time I've used the "edit" on github feature.  I thought it would be a low overhead way to propose a change - turns out though that github just forked your repo and added a branch in my account!  Anyway, I added another commit to my proposal.

wooorm:
I don’t think this is the place to explain how to use ESM for MDX with each tool. It would become a giant section, which is very similar to the existing Getting started article.

How about instead using ESM for the example in Getting started: https://mdxjs.com/docs/getting-started/#nextjs

godmar:
Then please do it yourself.

I had an afterthought on this: the next.js starter, as of today, doesn't create a `next.config.mjs` file. So for people using the next.js starter, all the documentation for the rehypePlugins etc. is out of date from the get-go.

PR #1912, opened 2022-01-25
Add processor cache in webpack loader

wooorm:
Backports: wooorm/xdm@decafe9.

Related-to: GH-1468.

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -1,12 +1,18 @@
 /**
+ * @typedef {import('vfile').VFileCompatible} VFileCompatible
+ * @typedef {import('vfile').VFile} VFile
  * @typedef {import('@mdx-js/mdx').CompileOptions} CompileOptions
  * @typedef {Pick<CompileOptions, 'SourceMapGenerator'>} Defaults
  * @typedef {Omit<CompileOptions, 'SourceMapGenerator'>} Options
  * @typedef {import('webpack').LoaderContext<unknown>} LoaderContext
+ * @typedef {(vfileCompatible: VFileCompatible) => Promise<VFile>} Process
  */
 
 import {SourceMapGenerator} from 'source-map'
-import {compile} from '@mdx-js/mdx'
+import {createFormatAwareProcessors} from '@mdx-js/mdx/lib/util/create-format-aware-processors.js'
+
+/** @type {WeakMap<CompileOptions, Process>} */
+const cache = new WeakMap()
 
 /**
  * A Webpack (5+) loader for MDX.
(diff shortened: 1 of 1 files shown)


packages/loader/lib/index.js
```diff
@@ -21,20 +27,25 @@ export function loader(value, callback) {
   /** @type {Defaults} */
   const defaults = this.sourceMap ? {SourceMapGenerator} : {}
   const options = /** @type {CompileOptions} */ (this.getOptions())
+  const config = {...defaults, ...options}
```
remcohaszing, line 30: The config object is constructed here. This is always a new object by reference. It’s then used as a key in the weak map cache. Once the loader is done processing a file, the config object is unused and garbage collected, meaning it’s removed from the cache.

This means the cache is never used.

If `options` is always the same object, that could be used.

wooorm: Is `options` a reference to the same object? Seems to weird that `this.getOptions()` would do that.

remcohaszing: I don’t know. That’s why I wrote _if_.

I think it might be the same object if options are passed to the loader as an object, but not when they are passed as a query parameter. I’m really just guessing here though.

wooorm: We can still use `this.query` though?

wooorm: …assuming that worked before, you tested it?

remcohaszing: Yes, I tested it back then. If `query` is the raw object, that’s probably our best option. Just beware it might also be a string, which can’t be used as a key for weak maps.

It was heavily inspired by [`ts-loader`](https://github.com/TypeStrong/ts-loader). They use a similar approach [here](https://github.com/TypeStrong/ts-loader/blob/5c030bfdedbffac475530f1c90086ecef9a5b2b7/src/instance-cache.ts#L12).

Also when I implemented it, this was compatible with Webpack 4. The use of `getOptions()` indicate this is no longer the case. I don’t know if the use of Webpack 5 opens up better alternatives.

wooorm: `ts-loader` has a similar approach but caches on webpack compilers. Should we use those instead?

remcohaszing: I think it doesn’t affect most users, as most users instantiate a single compiler, but it’s still a good idea.

wooorm: TS uses a rather complex hash though: https://github.com/TypeStrong/ts-loader/blob/dcec071c110bff12d9d510a35869395862e4a6eb/src/index.ts#L182.
Which is then passed [through here](https://github.com/TypeStrong/ts-loader/blob/dcec071c110bff12d9d510a35869395862e4a6eb/src/index.ts#L38) as the [`name` here](https://github.com/TypeStrong/ts-loader/blob/dcec071c110bff12d9d510a35869395862e4a6eb/src/instances.ts#L46), which is [finally used here](https://github.com/TypeStrong/ts-loader/blob/5c030bfdedbffac475530f1c90086ecef9a5b2b7/src/instance-cache.ts#L26)

PR #1916, opened 2022-01-25
Add improved webpack cache

wooorm:
This uses crypto (the good one) to create a hash from options, for strong and proper caching.

This is somewhat based on the conversation in https://github.com/mdx-js/mdx/pull/1912#discussion_r791481193. Note though that for us, options could be *anything*, including non-JSON values. So caching with JSON isn’t perfect. But it’s likely better than nothing.

/cc @remcohaszing

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -5,13 +5,20 @@
  * @typedef {Pick<CompileOptions, 'SourceMapGenerator'>} Defaults
  * @typedef {Omit<CompileOptions, 'SourceMapGenerator'>} Options
  * @typedef {import('webpack').LoaderContext<unknown>} LoaderContext
+ * @typedef {import('webpack').Compiler} WebpackCompiler
  * @typedef {(vfileCompatible: VFileCompatible) => Promise<VFile>} Process
  */
 
+import {createHash} from 'node:crypto'
 import {SourceMapGenerator} from 'source-map'
 import {createFormatAwareProcessors} from '@mdx-js/mdx/lib/util/create-format-aware-processors.js'
 
-/** @type {WeakMap<CompileOptions, Process>} */
+const own = {}.hasOwnProperty
+
+// Note: the cache is heavily inspired by:
+// <https://github.com/TypeStrong/ts-loader/blob/5c030bf/src/instance-cache.ts>
+const marker = /** @type {WebpackCompiler} */ ({})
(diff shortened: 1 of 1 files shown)


remcohaszing:
> So caching with JSON isn’t perfect. But it’s likely better than nothing.

:point_up_2: my thoughts exactly

remcohaszing (approved):

packages/loader/lib/index.js
```diff
@@ -37,15 +48,44 @@ export function loader(value, callback) {
...
     callback(null, file.value, file.map)
     return file
   }, callback)
 }
+
+/**
+ * @param {Options} options
+ */
+function getOptionsHash(options) {
+  const hash = createHash('sha256')
```
ChristianMurphy, line 75: Here speed would be of value, it may be worth considering a blake2 hash

```suggestion
  const hash = createHash('blake2s256')
```

ChristianMurphy (approved):

packages/loader/lib/index.js
wooorm: I’m a bit uneasy with a digest format that is [not frequently used](https://github.com/search?l=JavaScript&q=%22blake2s256%22&type=Code).
Searching Google or markdown on GitHub seems to almost exclusively include it in lists of all possible algorithms, not in code like us here or in tutorials or so.

I was also uneasy about a bundled/external dependency. It does seem like `blake2s256` is supported in all maintained versions of Node.js though, as it seems to have landed in OpenSSL 1.1.1. So it seems okay, but I know that there are smaller Node.js builds out there as well?

I’m also assuming that the time factor is fine when using sha256, options objects should be rather small (especially as functions omitted) and performance bottlenecks of webpack and MDX are in other areas.

I do see room for improvement around hashing the options. In the area of preventing crashes on cyclical references in options, adding support for functions, and using a package to hash an options object which likely already exists, 

Given that you’re :+1: on this PR without this change, I’ll merge it as is, but I’m open to discussing improvements to this PRs functionality.

ChristianMurphy: the 2b variant is also an option, and is more common https://github.com/search?q=%22blake2b%22&type=Code
The reason I suggested 2s is it matches the hash size, and it optimized for 16/32 bit processors.
While 2b has a larger hash size and is optimized for 64 bit.

Issue #1772, opened 2021-10-22
Playground on site not catching invalid Javascript expressions error


homumado:
### Subject of the issue

Great Job on the v2 rc release! Noticed a minor bug with error catching on the new Playground page for the new Javascript expressions support.

### Your environment

*   OS: Window 10
* Packages: 2.0.0-rc.1 on https://v2.mdxjs.com/playground/
* Env: Chrome

### Steps to reproduce

Invalid JSX Errors are caught correctly by the Compile tab:

```
Hello, world!
Below is an example of markdown in JSX.

</div>
```

![image](https://user-images.githubusercontent.com/84349818/138506470-5d245629-c0a7-4e4e-a910-90f858928bf8.png)

Errors are NOT caught by the Compile tab when when using the new Javascript expressions syntax {...].

```
Hello, world!
Below is an example of markdown in JSX.

{r}
```

The page crashes from the error, error from chrome console:

![image](https://user-images.githubusercontent.com/84349818/138506603-df319d5f-f515-4c9a-a98a-d6a60f7b1ab7.png)

### Expected behaviour

Errors within Javascript expressions {...} should caught by the Playground Console tab like JSX syntax error.

### Actual behaviour

The page crashes from the error, force you to refresh Playground.

wooorm:
I think this is solved? https://github.com/mdx-js/mdx/issues/1791#issuecomment-1000987901

KuangAnthony:
This isn't solved, editor doesn't have a mechanism to catch invalid expressions. #1791 is related to codemirror state not being in sync with the application state

wooorm:
ah, right, thanks!

PR #1931, opened 2022-02-03
hopefully fixes issue #1772

homumado:
reference https://github.com/mdx-js/mdx/issues/1772

Add a Javascript try/catch to prevent the Playground page from crashing from incomplete JavaScript expressions in Preview.

Closes https://github.com/mdx-js/mdx/issues/1772

### DIFF
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -119,6 +119,17 @@ export const Editor = ({children}) => {
   )
   const stats = state.file ? statistics(state.file) : {}
 
+  //Create a preview component that can handle errors with try-catch block; hopefully catching invalid JS expressions errors
+  const Preview = useCallback(() => {
+    try {
+      return state.file.result();
+    } catch (error) {
+      return (<pre>
+        <code>{String(error.message)}</code>
+      </pre>);
+    }
+  }, [state.file.result]);
+
   return (
     <div>
       <Tabs className="frame">
@@ -234,7 +245,7 @@ export const Editor = ({children}) => {
           <div className="frame-body frame-body-box-fixed-height frame-body-box">
             {state.file && state.file.result ? (
               <ErrorBoundary FallbackComponent={FallbackComponent}>
-                {state.file.result()}
+                <Preview />
               </ErrorBoundary>
             ) : null}
           </div>


wooorm:
And did you try it? Does it work?

homumado:
Testing the Vercel Preview https://mdx-88oauc3yx-mdx.vercel.app/playground/

Passing in unscoped Components no longer cause the page to crash:

image

docs/_component/editor.client.js
```diff
@@ -119,6 +119,17 @@ export const Editor = ({children}) => {
   )
   const stats = state.file ? statistics(state.file) : {}
 
+  //Create a preview component that can handle errors with try-catch block; hopefully catching invalid JS expressions errors
```
ChristianMurphy, line 122:
```suggestion
  // Create a preview component that can handle errors with try-catch block; for catching invalid JS expressions errors
```

ChristianMurphy (approved):

docs/_component/editor.client.js
```diff
@@ -234,7 +245,7 @@ export const Editor = ({children}) => {
           <div className="frame-body frame-body-box-fixed-height frame-body-box">
             {state.file && state.file.result ? (
               <ErrorBoundary FallbackComponent={FallbackComponent}>
-                {state.file.result()}
+                <Preview />
               </ErrorBoundary>
```
wooorm, line 249: Does this remove the need for `ErrorBoundary`?

docs/_component/editor.client.js
```diff
@@ -119,6 +119,17 @@ export const Editor = ({children}) => {
   )
   const stats = state.file ? statistics(state.file) : {}
 
+  // Create a preview component that can handle errors with try-catch block; for catching invalid JS expressions errors
+  const Preview = useCallback(() => {
+    try {
+      return state.file.result();
```
wooorm, line 125: As this component (`state.file.result`) is now wrapped in another component (`Preview`), the benefit if calling it ourselves (`state.file.result()`) instead of using it like any other component (`<state.file.result />`) is lost.
Due to this case, I think it’s clearer to use JSX syntax here:  `<state.file.result />`

docs/_component/editor.client.js
```diff
@@ -119,6 +119,17 @@ export const Editor = ({children}) => {
...
+    } catch (error) {
+      return (<pre>
+        <code>{String(error.message)}</code>
+      </pre>);
+    }
+  }, [state]);
+
```
wooorm, line 132: This file is not formatted. You can run `npm run format` to apply formatting

docs/_component/editor.client.js
```diff
@@ -119,6 +119,17 @@ export const Editor = ({children}) => {
...
   const stats = state.file ? statistics(state.file) : {}
 
+  // Create a preview component that can handle errors with try-catch block; for catching invalid JS expressions errors
+  const Preview = useCallback(() => {
+    try {
+      return state.file.result();
+    } catch (error) {
+      return (<pre>
+        <code>{String(error.message)}</code>
+      </pre>);
```
wooorm, line 129: There is already similar (and improved) code in this file: `FallbackComponent`. Can you use that?

docs/_component/editor.client.js
wooorm: And isn’t it weird that `ErrorBoundary` doesn‘t catch the error in question? Is it a bug there, or something we should change here to make it work?

docs/_component/editor.client.js
homumado: Thanks for the feedback!

docs/_component/editor.client.js
homumado: Seems like ErrorBoundary is unnecessary for Preview here. 

I'm not entirely sure why try...catch can catch the error but ErrorBoundary doesn't either.

wooorm:
Thanks!

homumado:
This is my first ever OSS commit!  

Thanks for being patient and walking me through the code review!

PR #1936, opened 2022-02-09
Add missing dependency

bensmithett:
`@mdx-js/rollup` [depends on vfile](https://github.com/mdx-js/mdx/blob/main/packages/rollup/lib/index.js#L16) but I think it's just working in some cases at the moment because of the transitive dependency via `@mdx-js/mdx`.

My project has another transitive dependency on vfile@^4.0.0, which is the one yarn puts in the root `node_modules` and that `@mdx-js/rollup@2.0.0` ends up importing, then 💥 because it's the old commonjs version.

### DIFF
--- a/packages/rollup/package.json
+++ b/packages/rollup/package.json
@@ -40,7 +40,8 @@
   "dependencies": {
     "@mdx-js/mdx": "^2.0.0",
     "@rollup/pluginutils": "^4.0.0",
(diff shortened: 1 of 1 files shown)


wooorm (approved):

wooorm:
nice, thanks!

wooorm:
Released in 2.1.0! :)

PR #1967, opened 2022-03-08
Mark react as peer dependency

flex-jonghyen:
`@mdx-js/react` that is its dependency has `react as peer dependency.

So, It should have `react` as peer dependency.

I did it.

### DIFF
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -48,6 +48,7 @@
   },
   "peerDependencies": {
     "webpack": ">=4"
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
Duplicate of https://github.com/mdx-js/mdx/pull/1600

and as noted before

> The loader is now agnostic to frontend frameworks. That means it can be used with React, Emotion, Preact, Vue, whatnot. As a result, it no longer makes sense to add a React peer dep.

PR #1974, opened 2022-03-14
fix: vfile import

guoyunhe:
`vfile` is not an es module, and it only has a default export. So named import throw errors:

```
file:///Users/guoyunhe/Git/cratedx/node_modules/@mdx-js/rollup/lib/index.js:16
import {VFile} from 'vfile'
        ^^^^^
SyntaxError: Named export 'VFile' not found. The requested module 'vfile' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'vfile';
const {VFile} = pkg;
```

This PR uses the default export to fix the errors.

### DIFF
--- a/packages/rollup/lib/index.js
+++ b/packages/rollup/lib/index.js
@@ -13,7 +13,7 @@
  */
 
 import {SourceMapGenerator} from 'source-map'
-import vfile from 'vfile'
+import VFile from 'vfile'
 import {createFilter} from '@rollup/pluginutils'
 import {createFormatAwareProcessors} from '@mdx-js/mdx/lib/util/create-format-aware-processors.js'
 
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
VFile is an ES Module and its default export is `VFile` (see https://github.com/vfile/vfile/blob/7339f82689a1f3bf0e30881476caf2485ff7a48e/package.json#L35 and https://github.com/vfile/vfile/blob/7339f82689a1f3bf0e30881476caf2485ff7a48e/index.js#L1)

It is likely a resolution issue in your package.json.
If `vfile` is listed as a dependency, devDepenency, or resolution make sure to either remove that, or update the version to match mdx rollup

https://github.com/mdx-js/mdx/blob/4f5d7ba119bcc7d77369555ca12ab6678caea3cc/packages/rollup/package.json#L44

guoyunhe:
Yes, it is because some other package depends on vfile 4.x. It is strange that mdx didn't resolve its own dependency tree...

wooorm:
That’s probably more related to your package manager (yarn, npm, etc), and not mdx!

PR #1975, opened 2022-03-14
Fix some more playground errors

### LINKED ISSUE #1956: Consistent crash of playground by replacing JSX variable with a double quote
Hi everyone. Hope you're well.

### Initial checklist

_This pertains to [the playground](https://mdxjs.com/playground/), so whatever package/version is in use there._

### Link to runnable example

_See [below](#steps-to-reproduce)._

### Steps to reproduce

Probably the rendering engine and the OS don't matter, but I tried both Firefox and Chrome (latest versions) on macOS, on Intel, running Catalina.

1. visit https://mdxjs.com/playground/
2. select `{{padding: '1rem', backgroundColor: 'violet'}}` with the pointing device (or your finger)
3. overwrite the selection by typing `"`

### Expected behavior

I'm probably not going to get the terminology right here, since JSX and all this is new to me, but I am trying this out because I think I might have a use case for it.

The first thing I wondered about with the playground demo is "how is this different from CSS?" so I tried replacing the object within the JSX variable—the `{{stuff in braces}}`—with a quoted attribute, plain CSS. This is when I experienced the crash.

### Actual behavior

The entire page goes white. The following error messages appear in the developer console:

```
Error: Minified React error #62; visit https://reactjs.org/docs/error-decoder.html?invariant=62 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    React 8
react-dom.production.min.js:193:243
```
```
Uncaught Error: Minified React error #62; visit https://reactjs.org/docs/error-decoder.html?invariant=62 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    React 8
react-dom.production.min.js:50:340
```

### Runtime, Package manager, OS, Build and bundle tools

As noted above, this was Firefox and Chrome on macOS, but I think that probably doesn't matter. The issue pertains to the web site, specifically [the playground](https://mdxjs.com/playground/).

<hr>

Thanks!

homumado:
Closes #1956

Looks like this bug got re-introduced when we removed the Preview ErrorBoundary in favor of a try... catch. Having both would catch both types of bugs.

Bug caught by the try...catch:

* Referencing components that's not within scope: `<Test />`
* Unclosed components: `<Test>`

Bugs caught by ErrorBoundary:

* syntax errors in dom, such as `<div style=""></div>`

### DIFF
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -241,7 +241,9 @@ export const Editor = ({children}) => {
         <TabPanel>
           <noscript>Enable JavaScript for the rendered result.</noscript>
           <div className="frame-body frame-body-box-fixed-height frame-body-box">
-            {state.file && state.file.result ? <Preview /> : null}
+            <ErrorBoundary FallbackComponent={ErrorFallback}>
+              {state.file && state.file.result ? <Preview /> : null}
+            </ErrorBoundary>
           </div>
         </TabPanel>
         <TabPanel>


ChristianMurphy (approved):

wooorm (approved):

wooorm:
Thanks @homumado!!

Issue #1802, opened 2021-11-08
mdx v2: import URL causes fail to compile error 


homumado:
### Subject of the issue

The `import {URL} from 'url'` line in [recma-document plugin](https://github.com/mdx-js/mdx/blob/079c8c7bb953dda5e59cf4c93fb924ad07b15ad1/packages/mdx/lib/plugin/recma-document.js) creates a compile error.

I don't think it's necessary to import the URL class since it's a global object?

### Your environment

*   **OS**: Window 10 64bit
*   **Packages**: v.2.0.0.rc
*   **Env**:  yarn v1.22.15 Next 11.0.1

### Steps to reproduce

`yarn build` a nextjs project.

### Expected behaviour

Project compiles.

### Actual behaviour

Failed to compile due to error.

```
./node_modules/@mdx-js/mdx/lib/plugin/recma-document.js
Attempted import error: 'URL' is not exported from 'url' (imported as 'URL').
```

wooorm:
> I don't think it's necessary to import the URL class since it's a global object?

TypeScript can’t deal with it properly.

> `The import {URL} from 'url' line in recma-document plugin creates a compile error.`

Where? Several people are using xdm and mdxjs/mdx@next without errors. And `URL` should be exported from `'url'`, so this leads me to think its a problem in your bundler?

homumado:
I'm using webpack 5 for production bundling. 

`import {URL} from 'url'` line is the only difference between the recma-document plugins on mdx-js/mdx@next and xdm:

https://github.com/mdx-js/mdx/blob/079c8c7bb953dda5e59cf4c93fb924ad07b15ad1/packages/mdx/lib/plugin/recma-document.js

wooorm:
And how do you configure webpack? For example, I'm currently using all of this swimingly with webpack through next

homumado:
I worked on the problem a bit more overnight and figured out the problem was caused by the version of nextjs I was using. 

The compile error clears when I changed the next package from `"next": "v11.0.1"` to `"next": "latest"`.

cesutherland:
I'm bumping into this too in laravel-mix with the webpack 5.67.0. Afaict it works fine w/o `import {URL} from 'url'` but I'll make sure and submit a patch.

cesutherland:
OK, in my case this is related to some project importing `node-url`, and this issue: https://github.com/defunctzombie/node-url/issues/53 URL not exported.

[...]

PR #1976, opened 2022-03-15
Remove use of `URL` from `url`

zfben:
URL is a global reference, it doesn't need to import before use. And import will break webpack.

Here is a related issue:

https://github.com/mdx-js/mdx/issues/1802

### DIFF
--- a/packages/mdx/lib/plugin/recma-document.js
+++ b/packages/mdx/lib/plugin/recma-document.js
@@ -39,7 +39,6 @@
  *   JSX runtime to use.
  */
 
-import {URL} from 'url'
 import {analyze} from 'periscopic'
 import {stringifyPosition} from 'unist-util-stringify-position'
 import {positionFromEstree} from 'unist-util-position-from-estree'
--- a/website/mdx-config.js
+++ b/website/mdx-config.js
@@ -8,7 +8,6 @@
 
 import path from 'path'
 import process from 'process'
-import {pathToFileURL} from 'url'
 import remarkFrontmatter from 'remark-frontmatter'
 import remarkGfm from 'remark-gfm'
 import remarkGithub from 'remark-github'
--- a/website/mdx-config.js
+++ b/website/mdx-config.js
@@ -8,6 +8,7 @@
 
 import path from 'path'
 import process from 'process'
+import {pathToFileURL} from 'url'
 import remarkFrontmatter from 'remark-frontmatter'
 import remarkGfm from 'remark-gfm'
 import remarkGithub from 'remark-github'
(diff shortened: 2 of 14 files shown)


packages/mdx/lib/plugin/recma-document.js
```diff
@@ -39,7 +39,6 @@
  *   JSX runtime to use.
  */
 
-import {URL} from 'url'
```
wooorm, line 42: There are likely other `URL` imports in other files.
So only changing this line, probably hides one error, but probably shows more errors.
Could you also change the other ones?

wooorm (commented):
Thanks for your PR. Can you change the other `URL` imports too?

Did you also read the conversation you linked to? Because you can configure webpack.

Luckily, TypeScript was fixed to finally support a global `URL`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58277#issuecomment-1026571932. So this can now be changed.

website/mdx-config.js
```diff
@@ -8,7 +8,6 @@
 
 import path from 'path'
 import process from 'process'
-import {pathToFileURL} from 'url'
```
wooorm, line 11: This is broken.

website/post.js
```diff
@@ -1,5 +1,4 @@
 import {promises as fs} from 'fs'
-import {fileURLToPath} from 'url'
```
wooorm, line 2: This is broken.

website/prep.js
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env node
 import {promises as fs} from 'fs'
-import {fileURLToPath} from 'url'
```
wooorm, line 3: This is broken.

website/prerender.js
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 import {promises as fs} from 'fs'
 import path from 'path'
-import {fileURLToPath} from 'url'
```
wooorm, line 4: This is broken.

zfben:
@wooorm I have removed all `URL`.
I know it can be through with configure webpack, but removing URL is a better way currently.
In fact, the first time for this issue is when I use mdx in a vite project.

wooorm (commented):
Nice! Could you remove the changes to `package-lock.json` from this PR? Rest looks good!

zfben:
> Nice! Could you remove the changes to `package-lock.json` from this PR? Rest looks good!

done

wooorm (approved):

wooorm:
Thanks! :)

wooorm:
Released in 2.1.0! :)

PR #1986, opened 2022-03-25
Fix to not add `_missingMdxReference` when thing is in scope

vlad-zhukov:
When using object members as components, for example `<x.y />`, MDX always adds the following checks:

```js
if (!x) _missingMdxReference("x", false);
if (!x.y) _missingMdxReference("x.y", true);
```

It works for basic cases, but breaks when used inside a function expression and `x` is assigned locally:

```mdx
Hello, world!

{(() => {
    const x = { y: () => '' }
    return <x.y />
})()}
```

This example throws `x is not defined` error because the variable is locally scoped. The solution I propose is to add `_missingMdxReference` only when an object (in this case `x`) is not in scope.

### DIFF
--- a/packages/mdx/lib/plugin/recma-jsx-rewrite.js
+++ b/packages/mdx/lib/plugin/recma-jsx-rewrite.js
@@ -146,9 +146,9 @@ export function recmaJsxRewrite(options = {}) {
                 !isInScope ||
                 // If the parent scope is `_createMdxContent`, then this
                 // references a component we can add a check statement for.
-                (parentNode &&
-                  parentNode.type === 'FunctionDeclaration' &&
-                  isNamedFunction(parentNode, '_createMdxContent'))
+                (parentScope &&
+                  parentScope.node.type === 'FunctionDeclaration' &&
+                  isNamedFunction(parentScope.node, '_createMdxContent'))
               ) {
                 fnScope.references[fullId] = {node, component: true}
               }
--- a/packages/mdx/lib/plugin/recma-jsx-rewrite.js
+++ b/packages/mdx/lib/plugin/recma-jsx-rewrite.js
@@ -142,7 +142,6 @@ export function recmaJsxRewrite(options = {}) {
               const parentScope = /** @type {Scope|null} */ (
                 currentScope.parent
               )
-              const parentNode = parentScope && parentScope.node
               if (
                 !isInScope ||
                 // If the parent scope is `_createMdxContent`, then this
--- a/packages/mdx/lib/plugin/recma-jsx-rewrite.js
+++ b/packages/mdx/lib/plugin/recma-jsx-rewrite.js
@@ -136,11 +136,24 @@ export function recmaJsxRewrite(options = {}) {
             const fullId = ids.join('.')
             const id = name.name
 
+            const isInScope = inScope(currentScope, id)
+
             if (!own.call(fnScope.references, fullId)) {
-              fnScope.references[fullId] = {node, component: true}
+              const parentScope = /** @type {Scope|null} */ (
+                currentScope.parent
+              )
+              const parentNode = parentScope && parentScope.node
+              if (
+                !isInScope ||
+                (parentNode &&
+                  parentNode.type === 'FunctionDeclaration' &&
+                  isNamedFunction(parentNode, '_createMdxContent'))
+              ) {
+                fnScope.references[fullId] = {node, component: true}
+              }
             }
 
-            if (!fnScope.objects.includes(id) && !inScope(currentScope, id)) {
+            if (!fnScope.objects.includes(id) && !isInScope) {
               fnScope.objects.push(id)
             }
           }
--- a/packages/mdx/lib/plugin/recma-jsx-rewrite.js
+++ b/packages/mdx/lib/plugin/recma-jsx-rewrite.js
@@ -145,6 +145,7 @@ export function recmaJsxRewrite(options = {}) {
               const parentNode = parentScope && parentScope.node
               if (
                 !isInScope ||
+                // If the parent is a _createMdxContent function, it's a top-level component in MDX.
                 (parentNode &&
                   parentNode.type === 'FunctionDeclaration' &&
                   isNamedFunction(parentNode, '_createMdxContent'))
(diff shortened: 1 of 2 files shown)


wooorm:
Can you add tests for this behavior?

vlad-zhukov:
I already did, it's the same as in my PR message and it's the smallest one I could think of.

But now I think a more real example would be the one with render props. Maybe something like this?

```jsx
<a.b render={(c) => <c.d />} />
```

wooorm:
Oh huh I missed it sorry.

It’s better to not test the serialized output. Your change fixes an exception. So can you add a “normal” test, like all the other ones outside of “jsx”?

- In your existing test, please try and remove everything that’s not needed for the error to occur
- In your newly proposed test, the `a.b` is also not needed?
- What about say the following, or other possible tests?:
  ```js
  export const a = {}

  <a.b />

Lastly, the code is a bit difficult, but I believe your proposed change change the place that tracks references. These components and objects are still referenced? Could we instead improve where and how `_missingMdxReference` is injected?

vlad-zhukov:
> In your newly proposed test, the `a.b` is also not needed?

Correct, it doesn't

> What about say the following, or other possible tests?:

It would work. But I found another issue that if there are no markdown components, `_missingMdxReference` is not injected 🤷 Thats why I added `Hello, world!` in my example

> Lastly, the code is a bit difficult, but I believe your proposed change change the place that tracks references. These components and objects are still referenced? Could we instead improve where and how _missingMdxReference is injected?

As far as I understand, these references are only used for injecting `_missingMdxReference`. And in the next `if` branch this check is there:

https://github.com/mdx-js/mdx/blob/427bcad073b36637b250ba7dffe1b33f75c003ec/packages/mdx/lib/plugin/recma-jsx-rewrite.js#L155:L169

> Also what was your practical MDX (so not the minimal test, but the original) that firs resulted in this problem?

The original code with which I discovered this problem is not far from the minimal test. Something along the lines:

```jsx
<CodePreview preview={() => {
  const store = createStore();
  return <store.Provider>...</store.Provider>;
}} />
```

vlad-zhukov:
See #1988. I will update this PR after that one.

wooorm:
is this ready for review

vlad-zhukov:
It's ready. I found a way to make the first test case remain as it was. What do you think about this approach?

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -136,11 +136,25 @@ export function recmaJsxRewrite(options = {}) {
             const fullId = ids.join('.')
             const id = name.name
 
+            const isInScope = inScope(currentScope, id)
+
             if (!own.call(fnScope.references, fullId)) {
-              fnScope.references[fullId] = {node, component: true}
+              const parentScope = /** @type {Scope|null} */ (
```
wooorm, line 142: is this cast needed? What else can it be?

packages/mdx/test/compile.js
```diff
@@ -521,23 +520,32 @@ test('compile', async () => {
...
+        await run(compileSync('<a render={() => <x.y />} />'))
       )
     )
     assert.unreachable()
   } catch (/** @type {unknown} */ error) {
     const exception = /** @type {Error} */ (error)
     assert.match(
       exception.message,
-      /x is not defined/,
+      /Expected object `x` to be defined/,
```
wooorm, line 534: awesome!

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -136,11 +136,25 @@ export function recmaJsxRewrite(options = {}) {
...
+                currentScope.parent
+              )
+              const parentNode = parentScope && parentScope.node
+              if (
+                !isInScope ||
+                // If the parent is a _createMdxContent function, it's a top-level component in MDX.
```
wooorm, line 148:
```suggestion
                // If the parent scope is `_createMdxContent`, then this
                // references a component we can add a check statement for.
```

I think this explains what’s going on a bit better?

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -136,11 +136,25 @@ export function recmaJsxRewrite(options = {}) {
...
             const id = name.name
 
+            const isInScope = inScope(currentScope, id)
+
             if (!own.call(fnScope.references, fullId)) {
-              fnScope.references[fullId] = {node, component: true}
+              const parentScope = /** @type {Scope|null} */ (
+                currentScope.parent
+              )
+              const parentNode = parentScope && parentScope.node
```
wooorm, line 145:
```suggestion

```

I think the `node` is always there, assuming `parentScope` is non-nully, right? If so, we can move this check to the condition below?

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -136,11 +136,25 @@ export function recmaJsxRewrite(options = {}) {
...
+              if (
+                !isInScope ||
+                // If the parent is a _createMdxContent function, it's a top-level component in MDX.
+                (parentNode &&
+                  parentNode.type === 'FunctionDeclaration' &&
+                  isNamedFunction(parentNode, '_createMdxContent'))
```
wooorm, line 151:
```suggestion
                (parentScope &&
                  parentScope.node.type === 'FunctionDeclaration' &&
                  isNamedFunction(parentScope.node, '_createMdxContent'))
```

wooorm (commented):
Sweet, some style notes, otherwise great!

packages/mdx/lib/plugin/recma-jsx-rewrite.js
vlad-zhukov: It's needed because this `Scope` type is defined locally

https://github.com/mdx-js/mdx/blob/0df684bc343a4b89ecf0a849fe8fefe52233386b/packages/mdx/lib/plugin/recma-jsx-rewrite.js#L19

Without this cast the type is `periscopic.Scope` and `node` is not there.

wooorm (approved):

wooorm:
Thanks! :)

PR #1988, opened 2022-03-26
Fix to add `_missingMdxReference` in some more cases

vlad-zhukov:
While working on #1986 I found that `_missingMdxReference` checks are not injected sometimes. For example:

```mdx
export const a = {}

<a.b />
```

compiles to

```jsx
/*@jsxRuntime automatic @jsxImportSource react*/
const {jsx: _jsx} = arguments[0];
const a = {};
function MDXContent(props = {}) {
  const {wrapper: MDXLayout} = props.components || ({});
  return MDXLayout ? _jsx(MDXLayout, Object.assign({}, props, {
    children: _jsx(_createMdxContent, {})
  })) : _createMdxContent();
  function _createMdxContent() {
    return _jsx(a.b, {});
  }
}
return {
  a,
  default: MDXContent
};
```

but if I add some markdown content

```mdx
Hello, world!

export const a = {}

<a.b />
```

compiles to

```jsx
/*@jsxRuntime automatic @jsxImportSource react*/
const {Fragment: _Fragment, jsx: _jsx, jsxs: _jsxs} = arguments[0];
const a = {};
function MDXContent(props = {}) {
  const {wrapper: MDXLayout} = props.components || ({});
  return MDXLayout ? _jsx(MDXLayout, Object.assign({}, props, {
    children: _jsx(_createMdxContent, {})
  })) : _createMdxContent();
  function _createMdxContent() {
    const _components = Object.assign({
      p: "p"
    }, props.components);
    if (!a) _missingMdxReference("a", false);
    if (!a.b) _missingMdxReference("a.b", true);
    return _jsxs(_Fragment, {
      children: [_jsx(_components.p, {
        children: "Hello, world!"
      }), "\n", "\n", _jsx(a.b, {})]
    });
  }
}
return {
  a,
  default: MDXContent
};
function _missingMdxReference(id, component) {
  throw new Error("Expected " + (component ? "component" : "object") + " `" + id + "` to be defined: you likely forgot to import, pass, or provide it.");
}
```

This happens because all transforms are inside [this `if` statement](https://github.com/mdx-js/mdx/blob/427bcad073b36637b250ba7dffe1b33f75c003ec/packages/mdx/lib/plugin/recma-jsx-rewrite.js#L266).

```
if (defaults.length > 0 || actual.length > 0) {
```

I fixed it by moving code that injects `_missingMdxReference` statements outside of this `if`.

### DIFF
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -516,7 +516,7 @@ test('compile', async () => {
     assert.match(
       exception.message,
       /Expected component `a.b` to be defined/,
-      'should throw if a required member is not passed'
+      'should throw if a used member is not defined locally'
     )
   }
 
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -504,6 +504,7 @@ test('compile', async () => {
     )
   }
 
+  // TODO: this is incorrect behavior, will be fixed in GH-1986
   try {
     renderToStaticMarkup(
       React.createElement(
@@ -520,6 +521,7 @@ test('compile', async () => {
     )
   }
 
+  // TODO: this is incorrect behavior, will be fixed in GH-1986
   try {
     renderToStaticMarkup(
       React.createElement(
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -532,7 +532,7 @@ test('compile', async () => {
     assert.match(
       exception.message,
       /x is not defined/,
-      'should throw if a required member is not passed'
+      'should throw if a used member is not defined locally (JSX in a function)'
     )
   }
 
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -506,6 +506,38 @@ test('compile', async () => {
 
   try {
     renderToStaticMarkup(
+      React.createElement(
+        await run(compileSync('export const a = {}\n\n<a.b />'))
+      )
+    )
+    assert.unreachable()
+  } catch (/** @type {unknown} */ error) {
+    const exception = /** @type {Error} */ (error)
+    assert.match(
+      exception.message,
+      /Expected component `a.b` to be defined/,
+      'should throw if a required member is not passed'
+    )
+  }
+
+  try {
+    renderToStaticMarkup(
+      React.createElement(
+        await run(compileSync('<a render={(x) => <x.y />} />'))
+      )
+    )
+    assert.unreachable()
+  } catch (/** @type {unknown} */ error) {
+    const exception = /** @type {Error} */ (error)
+    assert.match(
+      exception.message,
+      /x is not defined/,
+      'should throw if a required member is not passed'
+    )
+  }
+
+  try {
+    renderToStaticMarkup(
       React.createElement(await run(compileSync('<X />', {development: true})))
     )
     assert.unreachable()
--- a/packages/mdx/lib/plugin/recma-jsx-rewrite.js
+++ b/packages/mdx/lib/plugin/recma-jsx-rewrite.js
@@ -243,25 +243,8 @@ export function recmaJsxRewrite(options = {}) {
             }
           }
 
-          /** @type {string} */
-          let key
-
-          // Add partials (so for `x.y.z` it’d generate `x` and `x.y` too).
-          for (key in scope.references) {
-            if (own.call(scope.references, key)) {
-              const parts = key.split('.')
-              let index = 0
-              while (++index < parts.length) {
-                const partial = parts.slice(0, index).join('.')
-                if (!own.call(scope.references, partial)) {
-                  scope.references[partial] = {
-                    node: scope.references[key].node,
-                    component: false
-                  }
-                }
-              }
-            }
-          }
+          /** @type {Array<Statement>} */
+          const statements = []
 
           if (defaults.length > 0 || actual.length > 0) {
             if (providerImportSource) {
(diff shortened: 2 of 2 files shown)


packages/mdx/test/compile.js
```diff
@@ -791,6 +791,55 @@ test('jsx', async () => {
...
+    String(compileSync('<a render={(x) => <x.y />} />', {jsx: true})),
+    [
+      '/*@jsxRuntime automatic @jsxImportSource react*/',
+      'function MDXContent(props = {}) {',
+      '  const {wrapper: MDXLayout} = props.components || ({});',
+      '  return MDXLayout ? <MDXLayout {...props}><_createMdxContent /></MDXLayout> : _createMdxContent();',
+      '  function _createMdxContent() {',
+      '    if (!x) _missingMdxReference("x", false);',
+      '    if (!x.y) _missingMdxReference("x.y", true);',
+      '    return <a render={x => <x.y />} />;',
```
vlad-zhukov, line 831: I've also added a test case that produces faulty code

wooorm (commented):
Thanks, the code seems good, but can you change the tests to instead add tests that run MDX, which should/should not throw errors and warnings, instead of a certain format of generated code?

Some existing tests for the behavior you’re changing are here: https://github.com/mdx-js/mdx/blob/cd9b32fa84062c10769398ad7e055394d4e69e5b/packages/mdx/test/compile.js#L483-L540.

packages/mdx/test/compile.js
```diff
@@ -504,6 +504,38 @@ test('compile', async () => {
...
+        await run(compileSync('export const a = {}\n\n<a.b />'))
+      )
+    )
+    assert.unreachable()
+  } catch (/** @type {unknown} */ error) {
+    const exception = /** @type {Error} */ (error)
+    assert.match(
+      exception.message,
+      /Expected component `a.b` to be defined/,
+      'should throw if a required member is not passed'
```
wooorm, line 519:
```suggestion
      'should throw if a used member is not defined locally'
```

packages/mdx/test/compile.js
```diff
@@ -504,6 +504,38 @@ test('compile', async () => {
...
+        await run(compileSync('<a render={(x) => <x.y />} />'))
+      )
+    )
+    assert.unreachable()
+  } catch (/** @type {unknown} */ error) {
+    const exception = /** @type {Error} */ (error)
+    assert.match(
+      exception.message,
+      /x is not defined/,
+      'should throw if a required member is not passed'
```
wooorm, line 535:
```suggestion
      'should throw if a used member is not defined locally (JSX in a function)'
```

packages/mdx/test/compile.js
```diff
@@ -504,6 +504,38 @@ test('compile', async () => {
...
+      React.createElement(
+        await run(compileSync('<a render={(x) => <x.y />} />'))
+      )
+    )
+    assert.unreachable()
+  } catch (/** @type {unknown} */ error) {
+    const exception = /** @type {Error} */ (error)
+    assert.match(
+      exception.message,
+      /x is not defined/,
```
wooorm, line 534: Wait, why is this error thrown here? Is `render` being called?

wooorm (commented):
Thanks!
- One big question
- Two tips on improved test messages!

packages/mdx/test/compile.js
vlad-zhukov: It does. That's exactly the issue I fix in #1986. This line is added:

```js
if (!x) _missingMdxReference("x", false);',
```

But `x` is not there, it's defined inside an arrow function.

vlad-zhukov:
Sorry I didn't communicate it properly, but both these errors will be gone in #1986.

wooorm:
Ahh, okay.


If these two tests test *broken* behavior that will be fixed in the next PR, then:
- Can you add a small comment above them for now 😅 `// To do this is wrong, will be fixed in GH-123123`.
- Can you add *actual tests* for what is broken *right now*, and will be fixed by your changes to `lib/`?

Furthermore, for the next PR, I believe the first error should remain an error? Because `b` is not defined in `a`!

vlad-zhukov:
> Can you add actual tests for what is broken right now

I'm not sure how to do it. This PR fixes an edge case so two code paths produce the same code now.

> Furthermore, for the next PR, I believe the first error should remain an error? Because `b` is not defined in `a`!

Oh, right! But it will be a React error because rendering `undefined` is illegal

wooorm:
> This PR fixes an edge case so two code paths produce the same code now.

But there was an error in one of those two though, right?

> Oh, right! But it will be a React error because rendering undefined is illegal

Taking from the code in this issue:

```js
if (!a.b) _missingMdxReference("a.b", true);
```

If possible, *that* should *eventually* be the error.

vlad-zhukov:
The first test covers it, isn't it?

wooorm:
> The first test covers it, isn't it?

I thought so but then you said it didn’t?

> but both these errors will be gone in https://github.com/mdx-js/mdx/pull/1986.

---

To recap:
- If there are tests that are about to change, please add a little comment next to those tests saying that they’re incorrect and will be fixed.
- If there are only incorrect tests, please also add an actual, working, correct test

wooorm (approved):
OK I’ll merge this but I think the first test is the way it should be if possible. Thanks!

vlad-zhukov:
Thanks!

wooorm:
Released in [`2.1.1`](https://github.com/mdx-js/mdx/releases/tag/2.1.1)!

vlad-zhukov:
Thank you for the quick responses and your work on this great project!

Issue #2008, opened 2022-04-09
Windows-compatibility of package script: docs-prerender


userzimmermann:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

```
$ npm run docs-prerender

> docs-prerender
> cross-env NODE_LOADER_CONFIG=website/loader.js node --unhandled-rejections=strict --no-warnings --experimental-loader @node-loader/core website/prerender.js

internal/process/promises.js:213
        triggerUncaughtException(err, true /* fromPromise */);
        ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (internal/errors.js:322:7)
    at defaultResolve (internal/modules/esm/resolve.js:859:11)
    at file:///C:/Users/userz/Projects/mdx/node_modules/@node-loader/core/lib/node-loader-core.js:165:26
    at file:///C:/Users/userz/Projects/mdx/node_modules/@node-loader/core/lib/node-loader-core.js:154:12
    at file:///C:/Users/userz/Projects/mdx/node_modules/@node-loader/core/lib/node-loader-core.js:117:14
    at async Loader.resolve (internal/modules/esm/loader.js:89:29)
    at async Loader.getModuleJob (internal/modules/esm/loader.js:242:17)
    at async Loader.import (internal/modules/esm/loader.js:177:17) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
```

### Solution

Working on it …

### Alternatives

Resignation ;)

PR #2009, opened 2022-04-09
Fix site build on Windows

userzimmermann:
It's redundant & causes Windows-incompatibility for absolute paths, because drive letters are misinterpreted as URL protocols

Solves https://github.com/mdx-js/mdx/issues/2008

### DIFF
--- a/website/prerender.js
+++ b/website/prerender.js
@@ -43,7 +43,7 @@ async function main() {
     Object.keys(manifest)
       .filter((d) => path.basename(d) !== 'index.client.js')
       .map(async (d) => {
-        cache[manifest[d]['*'].id] = await import(fileURLToPath(d))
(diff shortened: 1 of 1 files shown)


wooorm:
Hmm if I recall, these functions are needed for something, though I've indeed seen the problem you get on Windows on CI. I'll look into this some more when I'm behind a computer again on Monday to investigate, though a green CI is a good sign, thanks!

userzimmermann:
> though a green CI is a good sign

It actually doesn't affect current CI—since docs building is not tested; which #2013 would change. Yet, I'm still pondering if the latter is worth the additional CI running time …

wooorm:
> It actually doesn't affect current CI

The full build still runs in CI tho, meaning that unix works!

wooorm (approved):

Issue #2004, opened 2022-04-08
Describe current Site integration issues with CRA 5 in documentation


userzimmermann:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

https://mdxjs.com/docs/getting-started/#create-react-app-cra states that: "There is no need to configure it."

Unfortunately, that's currently not the case with CRA 5 anymore; and many people (including myself) have already spent a lot of time with figuring out reasons & finding workarounds …

### Solution

Please add a statement to https://mdxjs.com/docs/getting-started/#create-react-app-cra —briefly describing the situation & maybe directing people to possible workarounds like those discussed in https://github.com/mdx-js/mdx/discussions/1870 

### Alternatives

Just accepting the situation and waiting for https://github.com/facebook/create-react-app/issues/12166 to get fixed …

userzimmermann:
Ok, @wooorm – If you direct me to the corresponding documentation sources, then I'll add it :)

wooorm:
At the bottom of each page on the website is a link “Edit this page on GitHub”, that should point you there. More on how to contribute is explained in the contribution guide: https://mdxjs.com/community/contribute/.

userzimmermann:
Good morning, @wooorm :) & first: Thanks for quickly merging https://github.com/mdx-js/mdx/pull/2007 —yet I still can't build the documentation locally:

```
$ npm run docs-generate

> docs-generate
> cross-env NODE_LOADER_CONFIG=website/loader-react-server.js node --unhandled-rejections=strict --no-warnings --experimental-loader @node-loader/core --conditions=react-server website/generate.server.js

file:///C:/Users/userz/Projects/mdx/docs/_component/layout.server.js:56
      <div>
      ^

SyntaxError: Unexpected token '<'
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:139:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:236:14)
```

[...]

PR #2010, opened 2022-04-09
docs: Describe current issue w/CRA 5 integration

userzimmermann:
Solves #2004


wooorm:
Thanks! :)

> ESLint-unfriendly

I don’t believe ESLint has opinions about webpack-loader syntax. CRA uses a strict preset that doesn’t allow them. It’s *their* decision.

There is also another way to allow this syntax: `/* eslint-disable import/no-webpack-loader-syntax */`, as shown in the earlier example, or that rule but globally in `eslintConfig` (in `package.json`)?

Is the reason for CRACO really that the syntax is ESLint-unfriendly? Or is it about “normal imports” being more ergonomical, which can only be allowed when rewiring?

I guess that CRACO is currently, with CRA 5, the best solution for most new users? Then it should perhaps come first?

userzimmermann:
> ESLint-unfriendly

Indeed misleading & irrelevant :) —not mentioned anymore …

> I guess that CRACO is currently, with CRA 5, the best solution for most new users? Then it should perhaps come first?

Perhaps … Hmm … It's nonetheless a hack—warning about `incorrect peer dependency "react-scripts@^4.0.0"`—until CRACO gets itself an update for _officially_ supporting CRA 5 …

docs/docs/getting-started.server.mdx
```diff
@@ -506,7 +506,47 @@ Which in turn lets us choose whether to use the `@mdx-js/mdx` or
 through webpack loader syntax in imports.
 
 Install the webpack loader [`@mdx-js/loader`][mdx-loader].
-There is no need to configure it.
+
+To enable direct MDX imports w/o `!@mdx-js/loader!` prefix, the loader can
```
wooorm, line 510:
```suggestion
To enable MDX imports without `!@mdx-js/loader!` prefix, the loader can
```

docs/docs/getting-started.server.mdx
```diff
@@ -506,7 +506,47 @@ Which in turn lets us choose whether to use the `@mdx-js/mdx` or
...
+further be added to webpack’s config, using a
```
wooorm, line 511:
```suggestion
be added to the webpack config, using a
```

docs/docs/getting-started.server.mdx
```diff
@@ -506,7 +506,47 @@ Which in turn lets us choose whether to use the `@mdx-js/mdx` or
...
+[`react-scripts`](http://github.com/facebook/create-react-app/tree/main/packages/react-scripts)
```
wooorm, line 512:
```suggestion
`react-scripts`
```

I don’t think this has to be linked, as it does not add something useful to users that are trying to solve this problem, to open it in a tab?

docs/docs/getting-started.server.mdx
```diff
@@ -506,7 +506,47 @@ Which in turn lets us choose whether to use the `@mdx-js/mdx` or
...
+rewiring tool, e.g. [CRACO](http://github.com/gsoft-inc/craco).
```
wooorm, line 513:
```suggestion
rewiring tool, such as [CRACO](http://github.com/gsoft-inc/craco).
```

`such as`, compared to a latin abbreviation, is often easier to understand.

Are the popular and viable others?

docs/docs/getting-started.server.mdx
```diff
@@ -506,7 +506,47 @@ Which in turn lets us choose whether to use the `@mdx-js/mdx` or
...
+      configure: (webpackConfig) => {
+        addAfterLoader(webpackConfig, loaderByName('babel-loader'), {
+          test: /\.(md|mdx)$/,
+          loader: require.resolve('@mdx-js/loader')
+        })
+
+        return webpackConfig
+      }
+    }
+  }
```
wooorm, line 532:
```suggestion
  const {addAfterLoader, loaderByName} = require('@craco/craco')

  module.exports = {
    webpack: {
      configure(webpackConfig) {
        addAfterLoader(webpackConfig, loaderByName('babel-loader'), {
          test: /\.mdx?$/,
          loader: require.resolve('@mdx-js/loader')
        })
        return webpackConfig
      }
    }
  }
```

- match project code style
- use normal function, there’s no reason to use an arrow function
- use the same regex used in other places

docs/docs/getting-started.server.mdx
```diff
@@ -506,7 +506,47 @@ Which in turn lets us choose whether to use the `@mdx-js/mdx` or
...
+    return <Content />
+  }
+  ```
+</details>
+
+<Note type="info">
+  **Note**: due to [broken MDX import forwarding](http://github.com/facebook/create-react-app/issues/12166)
+  in `react-scripts` 5.x, rewiring is currently necessary when using CRA 5.
+  There’s an [ongoing discussion](http://github.com/mdx-js/mdx/discussions/1870)
+  about this issue.
```
wooorm, line 548:
```suggestion
  **Note**: rewiring is currently needed for CRA 5,
  due to a bug in `react-scripts`
  ([`facebook/create-react-app#12166`](http://github.com/facebook/create-react-app/issues/12166)),
  which is also tracked at
  [`mdx-js/mdx#1870`](http://github.com/mdx-js/mdx/discussions/1870).
```

How about reworded like this? Puts what it means for users (have to rewire) at the front. Then explains and links to in-depth stuff.

wooorm (commented):
Nice! Here’s some wording feedback.

Lastly, I think it would help readers to shuffle the structure: the note 1st, then the normal example 2nd, then the “Expand rewiring example (CRACO)” example 3rd, then the rest of the text?

wooorm:
perhaps good to mention the `incorrect peer dependency` warning?

docs/docs/getting-started.server.mdx
userzimmermann: > Are the popular and viable others?

Popular, yes: https://github.com/timarney/react-app-rewired – the standard for years! – only not compatible yet w/CRA 5; which might not change, given its poor recent maintenance … So actually: CRACO is the only viable one. I'll further rephrase that section …

wooorm (approved):

PR #2013, opened 2022-04-11
.github/workflows: Add jobs for docs

userzimmermann:
Especially to find OS-specific docs building issues, like https://github.com/mdx-js/mdx/issues/2008 —making life easier for docs contributors :)

All jobs should pass after rebasing on top of #2009 once merged …

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,8 +6,6 @@ jobs:
   small:
     name: test / ${{matrix.os}} / ${{matrix.node}}
     runs-on: ${{matrix.os}}
-    env:
-      PUPPETEER_SKIP_DOWNLOAD: 1
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


wooorm:
hmm, while useful, this would make *every* PR much slower to accept. It’s already a giant repo with a large clone size, a big site, and packages for much of the ecosystem. We’ve spend a lot of time making CI work reasonably fast around v2, that I don’t think I want to make each PR slower again just for this.

userzimmermann:
As an alternative, I've added `npm run docs` to the `small` jobs—using Node.js 14 for `macos` & `windows`. Yet, it still approximately doubles their running time …

wooorm:
You could try and look at ignoring the generation of og images (and skipping puppeteer again). That might shave off say 20%?

But, I’m guessing that folks working on the site are already fine contributing anyway, that most contributors will be on unix, and that the custom code for the site won’t change much, so given that it’s working now, I tend to give the most weight to the performance of PRs, so I’m leaning towards not accepting this PR?

userzimmermann:
> I’m leaning towards not accepting this PR?

Np ;) —I've learned a lot from working on it …

But anyway: Could the input easily be limited to just a single page for docs build testing?

wooorm:
> Could the input easily be limited to just a single page for docs build testing?

Nope, it’s all a rather bespoke system to support the undocumented upcoming react server components. It does not have knowledge of which file changed, or what assets are needed to build/show one particular page, unfortunately.

> I've learned a lot from working on the topic

Nice :+1:

---

Alright, closing, thanks!

PR #2021, opened 2022-04-23
Fix support for `baseUrl` rewriting `import.meta.url`

wooorm:
MDX can compile *and* run code, at the same time (called “evaluation”). To evaluate imports and exports in that code, it has to know *from where* to do that, which is a location on the users computer that is not the folder that contains `@mdx-js/mdx` in `node_modules`. That’s what the `baseUrl` option is used for.

Previously though, the URL given as `baseUrl` was not used to replace `import.meta.url` in MDX code. As they are equivalent, this commit introduces that behavior.


wooorm:
[Released in 2.1.2](https://github.com/mdx-js/mdx/releases/tag/2.1.2)!

PR #2033, opened 2022-05-11
Add `recma-nextjs-static-props` to list of plugins

remcohaszing:
I believe this is the first external recma plugin :tada:

### DIFF
--- a/docs/docs/extending-mdx.server.mdx
+++ b/docs/docs/extending-mdx.server.mdx
@@ -68,6 +68,8 @@ You can use this template:
 See also the [list of remark plugins][remark-plugins] and
 [list of rehype plugins][rehype-plugins].
 
+*   [`remcohaszing/recma-nextjs-static-props`](https://github.com/remcohaszing/recma-nextjs-static-props)
+    — expose top-level identifiers in Next.js `_app.js`
 *   [`remcohaszing/rehype-mdx-title`](https://github.com/remcohaszing/rehype-mdx-title)
     — expose the page title as a string
 *   [`remcohaszing/remark-mdx-code-meta`](https://github.com/remcohaszing/remark-mdx-code-meta)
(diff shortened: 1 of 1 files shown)


docs/docs/extending-mdx.server.mdx
```diff
@@ -68,6 +68,8 @@ You can use this template:
 See also the [list of remark plugins][remark-plugins] and
 [list of rehype plugins][rehype-plugins].
 
+*   [`remcohaszing/recma-nextjs-static-props`](https://github.com/remcohaszing/recma-nextjs-static-props)
+    — expose top-level identifiers in Next.js `_app.js`
```
wooorm, line 72: For here:

- Is this really about `app.js`/`_app.js` 🤔

For your readme:

- Should your readme mention that `include` can include things which are not defined without a problem? E.g., `whateverKey`, when not exported from an MDX file, whether that’s fine?
- Maybe a more practical example in your readme with an example MDX file that exports something, and how it can be used on the receiving end?

remcohaszing: Exposing it in `_app.js` is the goal, using `getStaticProps` is the mean. I think it's clear what the plugin does, but I struggle a bit to explain it in a one-line description. I'm open to suggestions.

Thanks for the readme suggestions! I'll add both 👍

wooorm: Oh, so `app.js` is how one could pass components to MDX without a provider! Other users before didn’t know that and instead used the provider: https://github.com/mdx-js/mdx/discussions/2032 and many similar.

You might want to link to the docs for `app.js`.

For a description, I’m thinking along the lines of:

```
automatically export things from MDX pages to Next.js (like `getStaticProps`)
```

But, agree that it’s tough. And this is still a bit long.

Maybe:

```
automatically generate `getStaticProps` (and similar) for Next.js
```

wooorm: friendly ping! :)

remcohaszing: I settled on:

> generate [`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching/get-static-props) exposing top level identifiers in Next.js

I think it’s also important to include what the generated `getStaticProps` does, as it’s not self-explanatory.

ChristianMurphy (approved):

wooorm (approved):

wooorm:
Thanks, I like this new byline!

PR #2040, opened 2022-05-15
Add `@next/mdx` to Next.js getting started guide

remcohaszing:
Next.js maintain their own package, `@next/mdx`, which can be used. It’s also useful to document how to configure the `MDXProvider`.

### DIFF
--- a/docs/docs/getting-started.server.mdx
+++ b/docs/docs/getting-started.server.mdx
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
   <summary>Expand example</summary>
 
   ```js path="next.config.js"
-  module.exports = {
-    // Prefer loading of ES Modules over CommonJS
-    experimental: {esmExternals: true},
+  import nextMDX from '@next/mdx'
+
+  const withMDX = nextMDX({
+    // By default only the .mdx extension is supported.
+    extension: /\.mdx?$/,
+    options: {/* providerImportSource: …, otherOptions… */}
+  })
+
+  export default withMDX({
     // Support MDX files as pages:
     pageExtensions: ['md', 'mdx', 'tsx', 'ts', 'jsx', 'js'],
-    // Support loading `.md`, `.mdx`:
-    webpack(config, options) {
-      config.module.rules.push({
-        test: /\.mdx?$/,
-        use: [
-          // The default `babel-loader` used by Next:
-          options.defaultLoaders.babel,
-          {
-            loader: '@mdx-js/loader',
-            /** @type {import('@mdx-js/loader').Options} */
-            options: {/* jsxImportSource: …, otherOptions… */}
-          }
-        ]
-      })
+  })
+  ```
+</details>
 
-      return config
-    }
+[Next.js](https://nextjs.org) have their own package for supporting MDX.
+
+Install and configure [`@next/mdx`][@next/mdx].
+There is no need to configure your JSX runtime as React is already set up.
+
+The React provider can be configured in [`pages/_app.js`](https://nextjs.org/docs/advanced-features/custom-app).
+In order to use it, you will have to configure the `providerImportSource` as
+well.
+
+<details>
+  <summary>Expand example</summary>
+
+  ```js path="next.config.js"
+  import nextMDX from '@next/mdx'
+
+  const withMDX = nextMDX({
+    // By default only the .mdx extension is supported.
+    extension: /\.mdx?$/,
+    options: { providerImportSource: '@mdx-js/react',  /* otherOptions… */}
+  })
+
+  export default withMDX({
+    // Support MDX files as pages:
+    pageExtensions: ['md', 'mdx', 'tsx', 'ts', 'jsx', 'js'],
+  })
+  ```
+
+  ```jsx path="pages/_app.js"
+  import {MDXProvider} from '@mdx-js/react'
+
+  import {Header} from '../components/Header'
+
+  const components = {
+    h1: Header;
+  }
+
+  export default function App({ Component, pageProps }) {
+    return (
+      <MDXProvider components={components}>
+        <Component {...pageProps} />
+      </MDXProvider>
+    )
   }
   ```
 </details>
 
-[Next](https://nextjs.org) supports webpack loaders by overwriting the webpack
-config in `next.config.js`.
-
-Install and configure the webpack loader [`@mdx-js/loader`][webpack].
-There is no need to configure your JSX runtime as React is already set up.
-
-See also [¶ Webpack][webpack] and [¶ React][react], which you’re using, for
-more on those tools.
+See [Using MDX with Next.js][next-mdx] for more details.
 
 #### Parcel
 
@@ -842,6 +870,8 @@ See their readmes on how to configure them.
 *   If you’re getting errors integrating MDX, see
     [§ Troubleshooting MDX][trouble] or [§ Support][support]
 
+[@next/mdx]: https://github.com/vercel/next.js/tree/canary/packages/next-mdx
+
 [svelte-jsx]: https://github.com/kenoxa/svelte-jsx
 
 [jsx]: #jsx
@@ -873,6 +903,8 @@ See their readmes on how to configure them.
 [mdx-rollup]: /packages/rollup/
 
 [mdx-vue]: /packages/vue/
+
+[next-mdx]: https://nextjs.org/docs/advanced-features/using-mdx
 
 [evaluate]: /packages/mdx/#evaluatefile-options
 
--- a/docs/docs/getting-started.server.mdx
+++ b/docs/docs/getting-started.server.mdx
@@ -549,7 +549,7 @@ the loader to the webpack config, by rewiring `react-scripts` using
 
 <Note type="info">
   **Note**: warnings about CRACO having `incorrect peer dependency
-      "react-scripts@^4.0.0"` can be ignored for the above to work.
+        "react-scripts@^4.0.0"` can be ignored for the above to work.
 </Note>
 
 See also [¶ Webpack][webpack], which is used in CRA, and see [¶ React][react],
@@ -593,7 +593,7 @@ on how to use MDX with Gatsby.
 Install and configure [`@next/mdx`][@next/mdx].
 There is no need to configure your JSX runtime as React is already set up.
 
-The React provider can be configured in [`pages/_app.js`][next-app].
+The MDX provider can be configured in [`pages/_app.js`][next-app].
 In order to use it, you need to configure the `providerImportSource` as
 well.
 
--- a/docs/docs/getting-started.server.mdx
+++ b/docs/docs/getting-started.server.mdx
@@ -549,7 +549,7 @@ the loader to the webpack config, by rewiring `react-scripts` using
 
 <Note type="info">
   **Note**: warnings about CRACO having `incorrect peer dependency
-    "react-scripts@^4.0.0"` can be ignored for the above to work.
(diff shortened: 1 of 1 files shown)


docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
-    experimental: {esmExternals: true},
+  import nextMDX from '@next/mdx'
+
+  const withMDX = nextMDX({
+    // By default only the .mdx extension is supported.
+    extension: /\.mdx?$/,
+    options: {/* providerImportSource: …, otherOptions… */}
+  })
+
+  export default withMDX({
```
wooorm, line 584: I believe Next itself, on pages, asks to first assign to a const, and the export default, so that there’s a name attached.
Should we do the same, or is it fine?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
-      return config
-    }
-  }
+  })
   ```
 </details>
 
-[Next](https://nextjs.org) supports webpack loaders by overwriting the webpack
-config in `next.config.js`.
+[Next.js](https://nextjs.org) have their own package for supporting MDX.
```
wooorm, line 591:
```suggestion
[Next.js](https://nextjs.org) has its own package to support MDX.
```

How about this?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
 
-Install and configure the webpack loader [`@mdx-js/loader`][webpack].
+Install and configure [`@next/mdx`][@next/mdx].
```
wooorm, line 593: I was under the impression that you still need `@mdx-js/loader` as well?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
 There is no need to configure your JSX runtime as React is already set up.
 
-See also [¶ Webpack][webpack] and [¶ React][react], which you’re using, for
-more on those tools.
+The React provider can be configured in [`pages/_app.js`](https://nextjs.org/docs/advanced-features/custom-app).
```
wooorm, line 596: Reference link w/ definition at bottom is the preference for most links

wooorm: I don’t think Next can work with Preact or something else, right?
Should `@mdx-js/react` be installed?
Should it be called React provider or MDX provider? MDX provider for React?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+In order to use it, you will have to configure the `providerImportSource` as
```
wooorm, line 597:
```suggestion
In order to use it, you need to configure the `providerImportSource` as
```

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+    // By default only the .mdx extension is supported.
+    extension: /\.mdx?$/,
+    options: { providerImportSource: '@mdx-js/react',  /* otherOptions… */}
+  })
+
+  export default withMDX({
+    // Support MDX files as pages:
+    pageExtensions: ['md', 'mdx', 'tsx', 'ts', 'jsx', 'js'],
+  })
+  ```
```
wooorm, line 616: Is it needed to repeat this? Maybe ``With the same `next.config.js` as above``?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+
+  ```jsx path="pages/_app.js"
+  import {MDXProvider} from '@mdx-js/react'
+
+  import {Header} from '../components/Header'
+
+  const components = {
+    h1: Header;
```
wooorm, line 624:
```suggestion
    h1: Header
```

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+  export default withMDX({
+    // Support MDX files as pages:
+    pageExtensions: ['md', 'mdx', 'tsx', 'ts', 'jsx', 'js'],
+  })
+  ```
+
+  ```jsx path="pages/_app.js"
+  import {MDXProvider} from '@mdx-js/react'
+
+  import {Header} from '../components/Header'
```
wooorm, line 621:
```suggestion
  import {Header} from '../components/Header.js'
```

Given that node moves towards with-extension, maybe include it here too?
I also don’t think the extra blank line is needed?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+
+  const components = {
+    h1: Header;
+  }
+
+  export default function App({ Component, pageProps }) {
```
wooorm, line 627:
```suggestion
  export default function App({Component, pageProps}) {
```

docs/docs/getting-started.server.mdx
```diff
@@ -842,6 +870,8 @@ See their readmes on how to configure them.
 *   If you’re getting errors integrating MDX, see
     [§ Troubleshooting MDX][trouble] or [§ Support][support]
 
+[@next/mdx]: https://github.com/vercel/next.js/tree/canary/packages/next-mdx
```
wooorm, line 873: Should we link to the Canary branch? Is that the default branch?

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+    return (
+      <MDXProvider components={components}>
+        <Component {...pageProps} />
+      </MDXProvider>
+    )
+  }
+  ```
+</details>
+
+See [Using MDX with Next.js][next-mdx] for more details.
```
wooorm, line 637: Should there also be a link for App in Next.js?

docs/docs/getting-started.server.mdx
remcohaszing: I don’t really understand what you mean, but this syntax is very similar to [this paragraph on using MDX](https://nextjs.org/docs/advanced-features/using-mdx#setup-nextmdx-in-nextjs).

docs/docs/getting-started.server.mdx
remcohaszing: It depends on your package manager. It’s a peer dependency. I [created an issue to propose properly migrating Next.js to MDX 2](https://github.com/vercel/next.js/issues/36888). This includes `@mdx-js/loader` to dependencies, but it was [moved to a discussion](https://github.com/vercel/next.js/discussions/36890) and has already moved to page 2, so I don’t think they want to prioritize this.

Since it’s a peer dependency, it’s installed implicitly when using newer npm versions, but not when using older npm versions or yarn.

docs/docs/getting-started.server.mdx
remcohaszing: Afaik Next only works with React. I think naming it `MDX provider` makes sense.

docs/docs/getting-started.server.mdx
remcohaszing: It’s different. It includes the `providerImportSource` option.

docs/docs/getting-started.server.mdx
wooorm: I believe there’s a warning in Next.js for:

```js
export default () => {}
```

It wants you to assign to a name:

```js
const SomePage = () => {}
export default SomePage
```

That is to attach a name. I think that’s good here too.

docs/docs/getting-started.server.mdx
remcohaszing: Next.js examples don’t use extensions, so I think it’s best to mimic that.

I usually keep a blank line between builtin, third party, and local module imports. In this particular case removing it is fine with me.

docs/docs/getting-started.server.mdx
wooorm: I believe newer npm versions require top-level users to install peer deps? 🤔 I believe in the past it didn’t, but now you do?

docs/docs/getting-started.server.mdx
remcohaszing: Yep, it’s their default branch.

docs/docs/getting-started.server.mdx
wooorm: Perhaps a diff would be clearer, to focus on only adding the provider import source?
Not adamant about this, go with what you want.

docs/docs/getting-started.server.mdx
remcohaszing: This isn’t a page, it’s the configuration file (`next.config.js`).

docs/docs/getting-started.server.mdx
remcohaszing: No, it’s the other way around. Maybe you’re confused because MDX [.npmrc](https://github.com/mdx-js/mdx/blob/main/.npmrc) uses legacy peer deps?

docs/docs/getting-started.server.mdx
remcohaszing: It’s linked right above the details. I don’t think it needs to be linked here again.

docs/docs/getting-started.server.mdx
wooorm: Weird. Not sure why I thought that.

Legacy peer deps are needed to show warnings but prevent errors. Many packages define incorrect peer ranges, and beta React version don’t work with peer ranges. npm 7 prevents them from being used.

docs/docs/getting-started.server.mdx
wooorm: We require ESM and they don’t, maybe that’s why?
I quite strongly prefer sticking with ESM/Node.js here and not document Webpack-specific forgiving features.

docs/docs/getting-started.server.mdx
wooorm: I understand it is not a page.
I use Next.js recommendation as an example.
It is still good to name things.
My assumption is that errors will occur and a name will improve stack traces.

docs/docs/getting-started.server.mdx
remcohaszing: Should we mention on MDX 1 here though? I believe this is mostly a guide for users who are either configuring Next.js with MDX in a new project, or adding MDX support to an existing Next.js project. In both cases I think the focus should be MDX 2.

docs/docs/getting-started.server.mdx
```diff
@@ -573,40 +573,68 @@ on how to use MDX with Gatsby.
...
+<details>
+  <summary>Expand example</summary>
+
+  ```js path="next.config.js"
+  import nextMDX from '@next/mdx'
+
+  const withMDX = nextMDX({
+    // By default only the .mdx extension is supported.
+    extension: /\.mdx?$/,
+    options: { providerImportSource: '@mdx-js/react',  /* otherOptions… */}
```
wooorm, line 609:
```suggestion
    options: {providerImportSource: '@mdx-js/react',  /* otherOptions… */}
```

docs/docs/getting-started.server.mdx
wooorm: nah, this is good!

wooorm (approved):

PR #2050, opened 2022-05-28
Override jsx built-in components

jlaramie:
I've been investigating the differences between `mdx-bundler` and `next-remote-mdx@3.0.5` and discovered that no matter what I did `mdx-bundler` would not replace any `jsx` that had a native version if explicitly used inside of the `mdx` file.

Here is an example which shows 3 different scenarios and the result using `latest`:
```html
<div>
  does not override: <img src="test.png" />
  <br />
  does override: ![Image](test.png)
  <br />
  does override: <Img src="test.png" />
</div>
```
- Scenario 1: `does not override: ` Since this is a built-in component, no override check is performed and `jsxRuntime.jsx.img` is used
- Scenario 2: `does override: Image` All md conversions automatically go through the `components` override check
- Scenario 3: `` Custom components automatically go through the `components` check

Eventually I tracked the difference between handling in `mdx-js/mdx` between version `1.*.*` and `2.*.*`.

Code Removed
```
// @ts-expect-error Allow fields passed through from mdast through hast to
// esast.
else if (node.data && node.data._mdxExplicitJsx) {
// Do not turn explicit JSX into components from `_components`.
// As in, a given `h1` component is used for `# heading` (next case),
// but not for `<h1>heading</h1>`.
```

The only reasoning I could come up with is that maybe it compiles slightly smaller and runs slightly faster but I don't think it'll have any real performance differences.

https://github.com/kentcdodds/mdx-bundler/issues/160


ChristianMurphy:
This behavior is intentional see the discussion in https://github.com/mdx-js/mdx/issues/821

wooorm:
This behavior is intentional

jlaramie:
@ChristianMurphy @wooorm should `disableParentContext` get passed into `rcma-jsx-rewrite.js`? What I'm trying to figure out is how to get `@mdx-js/esbuild` to be able to work the same way `@mdx-js/react` works.

Something like this perhaps?

```
diff --git a/node_modules/@mdx-js/mdx/lib/core.js b/node_modules/@mdx-js/mdx/lib/core.js
index 96e34d7..19d0d6d 100644
--- a/node_modules/@mdx-js/mdx/lib/core.js
+++ b/node_modules/@mdx-js/mdx/lib/core.js
@@ -75,6 +75,7 @@ export function createProcessor(options = {}) {
     remarkPlugins,
     remarkRehypeOptions = {},
     SourceMapGenerator,
+    disableParentContext,
     ...rest
   } = options
   let index = -1
@@ -122,7 +123,7 @@ export function createProcessor(options = {}) {
   pipeline
     .use(rehypeRecma)
     .use(recmaDocument, {...rest, outputFormat})
-    .use(recmaJsxRewrite, {development, providerImportSource, outputFormat})
+    .use(recmaJsxRewrite, {development, providerImportSource, outputFormat, disableParentContext})
 
   if (!jsx) {
     pipeline.use(recmaJsxBuild, {outputFormat})
diff --git a/node_modules/@mdx-js/mdx/lib/plugin/recma-jsx-rewrite.js b/node_modules/@mdx-js/mdx/lib/plugin/recma-jsx-rewrite.js
index e1d8dbf..5a6c7a8 100644
--- a/node_modules/@mdx-js/mdx/lib/plugin/recma-jsx-rewrite.js
+++ b/node_modules/@mdx-js/mdx/lib/plugin/recma-jsx-rewrite.js
@@ -59,7 +59,8 @@ const own = {}.hasOwnProperty
  * @type {import('unified').Plugin<[RecmaJsxRewriteOptions]|[], Program>}
  */
 export function recmaJsxRewrite(options = {}) {
-  const {development, providerImportSource, outputFormat} = options
+  const {development, providerImportSource, outputFormat, disableParentContext} = options
+  console.log('recmaJsxRewrite', options)
 
   return (tree, file) => {
     // Find everything that’s defined in the top-level scope.
@@ -183,7 +184,7 @@ export function recmaJsxRewrite(options = {}) {
           }
           // @ts-expect-error Allow fields passed through from mdast through hast to
           // esast.
-          else if (node.data && node.data._mdxExplicitJsx) {
+          else if (!disableParentContext && node.data && node.data._mdxExplicitJsx) {
             // Do not turn explicit JSX into components from `_components`.
             // As in, a given `h1` component is used for `# heading` (next case),
             // but not for `<h1>heading</h1>`.
```

jlaramie:
Also while messing with the tests it's unclear to me that `disableParentContext` as it is currently implemented fixes this issue. 

`disableParentContext` seems to only control whether markdown tags use the provider or not but have no control over whether native elements can be overridden.

wooorm:
I do not believe `disableParentContext` has anything to do with what you want: it has no relation to your problem.

talatkuyuk:
Those who want to override the jsx built-in components like `` or `<video >` or other lowercased html elements like `<custom />` or `<custom-component />` can use **[`recma-mdx-html-override`](https://github.com/ipikuka/recma-mdx-html-override)** in `recmaPlugins` in mdx options. It is super powerful.

Issue #2029, opened 2022-05-05
Generated JSX should not define component on rerender


remcohaszing:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

^2.0.0-rc.2 (#1810)

### Steps to reproduce

Run the following script:

```js
import { compile } from '@mdx-js/mdx';

const { value } = await compile('hello', { jsx:true })

console.log(value)
```

The `jsx` option is there for readability of the output. The issue applies if it’s either true or false.

### Expected behavior

The output **doesn’t** define a component inside a component. This is a React anti-pattern, because it causes unnecessary rerenders

The following Google search yields some articles that explain why this is a bad idea in React: https://www.google.com/search?q=react+create+component+inside+component. Example article: https://dev.to/borasvm/react-create-component-inside-a-component-456b. The same principles apply to Preact.

Earlier the output was ok:

```jsx
/*@jsxRuntime automatic @jsxImportSource react*/
function MDXContent(props = {}) {
  const _components = Object.assign({
    p: "p"
  }, props.components), {wrapper: MDXLayout} = _components;
  const _content = <><_components.p>{"hello"}</_components.p></>;
  return MDXLayout ? <MDXLayout {...props}>{_content}</MDXLayout> : _content;
}
export default MDXContent;

```

Alternative output that’s also ok for React/Preact, although I recall it’s troublesome for some frameworks:

```jsx
/*@jsxRuntime automatic @jsxImportSource react*/
import { Fragment as _Fragment } from 'react/jsx-runtime.js'
function MDXContent(props = {}) {
  const _components = Object.assign({
    p: "p"
  }, props.components), {wrapper: MDXLayout = _Fragment} = _components;
  return <MDXLayout><_components.p>{"hello"}</_components.p></MDXLayout>;
}
export default MDXContent;
```

### Actual behavior

The output **does** define a component inside a component, which causes unnecessary rerenders.

```jsx
/*@jsxRuntime automatic @jsxImportSource react*/
function MDXContent(props = {}) {
  const {wrapper: MDXLayout} = props.components || ({});
[...]

PR #2062, opened 2022-06-14
Split `_createMdxContent`

0phoff:
This PR Closes GH-2029.

It splits the `_createMdxContent` component function from the `MDXContent` component, because inner components cause a lot of unnecessary rerenders (see GH-2029 for more information).  
Additionally, I added a small "improvement", where we do not check for the existence of `MDXLayout` when a `Layout` is defined inside of the mdx file, as this is unnecessary in that case.


packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -76,43 +76,43 @@ export function recmaJsxRewrite(options = {}) {
...
-        }
 
-        let fnScope = fnStack[0]
+          // MDXContent only ever contains MDXLayout
+          if (
+            isNamedFunction(node, 'MDXContent') &&
+            newScope &&
+            !inScope(newScope, 'MDXLayout')
+          ) {
+            fnStack[stackLength - 1].components.push('MDXLayout')
```
wooorm, line 103: Any reason this is no longer `0`?
Even if in this case there’s only ever `0` on the stack, it’s probably better to be explicit about that and use `0`?

packages/mdx/lib/plugin/recma-document.js
```diff
@@ -518,7 +524,21 @@ export function recmaDocument(options = {}) {
     }
 
     // @ts-expect-error: JSXElements are expressions.
-    const consequent = /** @type {Expression} */ (element)
+    let consequent = /** @type {Expression} */ (element)
```
wooorm, line 527: Maybe a different name, now that this no longer has to be a consequent. `result`?

wooorm (commented):
Really cool that you worked on this!
Also smart to omit the layout check if it isn’t needed!

wooorm (approved):

packages/mdx/lib/plugin/recma-jsx-rewrite.js
wooorm: Ping, just making sure if you saw this!

wooorm (approved):

wooorm:
I’ll try to get a release done this weekend!

ChristianMurphy (approved):

wooorm:
[Released in 2.1.2](https://github.com/mdx-js/mdx/releases/tag/2.1.2)!

PR #2078, opened 2022-07-10
Update loader, rename ` module` to `mod` to avoid confusing with global `module` object

B2o5T:
Update loader, rename ` module` to `mod` to avoid confusing with global `module` object

### DIFF
--- a/packages/loader/index.cjs
+++ b/packages/loader/index.cjs
@@ -10,7 +10,7 @@
 module.exports = function (code) {
   const callback = this.async()
   // Note that `import()` caches, so this should be fast enough.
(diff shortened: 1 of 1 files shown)


wooorm:
Why? Who is confused?

B2o5T:
the first time I saw this file I was thinking about `loader` field in global `module` object

wooorm:
There are a lot of globals, shadowing some will happen.
This is a tiny function, where it’s not hard to loose focus of the one parameter that is passed in the one line it contains.
The value of `module` is a `Module` object, using that name make sense.
Abbreviations have other problems.

Thanks, but no thanks, I think this is fine!

PR #2101, opened 2022-08-11
Fix rewriting of components for custom elements

### LINKED ISSUE #2100: `jsx: true` output does not handle custom elements correctly
### Initial checklist

^2.1.2"

### Link to runnable example

https://stackblitz.com/github/bholmesdev/mdx-shiki-compilation?file=test.mjs

### Steps to reproduce

1. Run the MDX compiler with the `rehypeRaw` plugin applied + some other plugin that inserts a custom element ([Shiki twoslash](https://shikijs.github.io/twoslash/) in this case).
2. Set `jsx: true` in the compiler config
3. Note the value contains `<_components.custom-element-name>` in the output, but dashes (`-`) are not valid JSX!

### Expected behavior

Ideally, the `_components` statement would serialize dashes to camelCase or some other JSX-compliant string like so:

```js
const _components = Object.assign({
    ...
    "customElementName": "custom-element-name"
  }, props.components);
```

### Actual behavior

Instead, `_components` uses the element name untouched:

```js
const _components = Object.assign({
   ...
    "custom-element-name": "custom-element-name"
  }, props.components);
```

### Runtime

Node v16

### Package manager

pnpm

### OS

macOS

### Build and bundle tools

Other (please specify in steps to reproduce)

bholmesdev:
Resolves #2100 
- Writes invalid component identifiers to separate variables, with naming `_component[#]`
- Use separate variables instead of `_components` object when `jsx: true`


packages/mdx/test/compile.js
```diff
@@ -859,9 +859,9 @@ test('jsx', async () => {
       '/*@jsxRuntime automatic @jsxImportSource react*/',
       'function _createMdxContent(props) {',
       '  const _components = Object.assign({',
-      '    "a-b": "a-b"',
+      '    a_b: "a-b"',
       '  }, props.components);',
-      '  return <>{<_components.a-b></_components.a-b>}</>;',
```
bholmesdev, line 864: This test asserts that the output is _invalid_ JSX! Converting dashes `-` to underscores `_` should resolve this issue for custom elements 👍

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -193,16 +195,21 @@ export function recmaJsxRewrite(options = {}) {
...
+            }
+
+            node.openingElement.name = componentName
 
             if (node.closingElement) {
-              node.closingElement.name = toJsxIdOrMemberExpression([
-                '_components',
-                id
-              ])
+              node.closingElement.name = componentName
```
wooorm, line 212: This needs to be cloned. It’s a weird edge case, but we shouldn’t have the same object multiple times in the tree.

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -344,6 +355,17 @@ export function recmaJsxRewrite(options = {}) {
               componentsInit = {type: 'Identifier', name: '_components'}
             }
 
+            for (const [
+              invalidComponentName,
+              variable
+            ] of invalidComponentNameToVariable) {
```
wooorm, line 361: I’d prefer for this to use a bit of a simpler loop. While, a `for .. in`, like much of the other code.
Or at least I find the destructuring a confusing to read, so assigning it to on variable, and use that?

packages/mdx/test/compile.js
```diff
@@ -858,13 +858,11 @@ test('jsx', async () => {
     [
       '/*@jsxRuntime automatic @jsxImportSource react*/',
       'function _createMdxContent(props) {',
-      '  const _components = Object.assign({',
-      '    "a-b": "a-b"',
-      '  }, props.components);',
-      '  return <>{<_components.a-b></_components.a-b>}</>;',
+      '  const _components = props.components || ({}), _component0 = "a-b";',
```
wooorm, line 861: We should still take this from `components['a-b']` though, so people can pass them in? 🤔 Or am I missing something and is it intentional that you’re not doing that?

packages/mdx/test/compile.js
```diff
@@ -858,13 +858,11 @@ test('jsx', async () => {
...
+      '  return <>{<_component0></_component0>}</>;',
       '}',
       'function MDXContent(props = {}) {',
-      '  const {wrapper: MDXLayout} = props.components || ({});',
+      '  const _component0 = "a-b", {wrapper: MDXLayout} = props.components || ({});',
```
wooorm, line 865: Same here as above?

wooorm (commented):
Looking good. One important thing. One small thing. One stylistic thing!

packages/mdx/test/compile.js
bholmesdev: Ah, I think I understand this prop a bit better now. Looks like I didn't implement this right! I'll rework 👍

wooorm: Sweet, thanks!
You could also likely set it on that destructuring object, alongside `wrapper: MDXLayout`? You may need to set a computed field in the AST or so though 🤔: https://astexplorer.net/#/gist/702f0315bf76ddbaf45a0928b0b06e96/784539f882835483b4541f9eff8d81ec4a327733

packages/mdx/lib/plugin/recma-jsx-rewrite.js
bholmesdev: I understand, and I think the variables could have better names! However, I don't know an easier way to iterate over a JS `Map` with type safety (note this is a `Map` and not an object). I tried this refactor using a `for ... in` over the keys as you suggested:

```js
for (const literalValue in invalidComponentNameToVariable) {
  const identifierName = invalidComponentNameToVariable.get(literalValue)
  declarations.push({
    type: 'VariableDeclarator',
    id: {type: 'Identifier', name: identifierName},
    init: {type: 'Literal', value: literalValue}
  })
}
```
...but `identifierName` is of type `string | undefined`, since we can't guarantee a result from a `.get` call. This throws a type error on `name`!

Respectfully, I think this `for ... of` solution is the most readable way to iterate over a `Map` in this case. What do you think about keeping the `for ... of` with this variable naming?

```js
for (const [identifierName, literalValue] of invalidComponentNameToVariable) {
  declarations.push({
    type: 'VariableDeclarator',
    id: {type: 'Identifier', name: identifierName},
    init: {type: 'Literal', value: literalValue}
  })
}
```

wooorm: Alright, sure!

You could use `name, value` as identifiers, and then have them be shortcuts in the declaration. It’s a small loop body, those names are very clear as well?

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -195,21 +195,24 @@ export function recmaJsxRewrite(options = {}) {
               fnScope.tags.push(id)
             }
 
-            let componentName = toJsxIdOrMemberExpression(['_components', id])
+            /** @type {(string | number)[]} */
```
wooorm, line 198:
```suggestion
            /** @type {Array<string | number>} */
```

Rest of the code uses `Array<_>`, so for consistency.
I prefer it because a) it’s more in line with other types over something else, b) it’s the way I’d read something, c) `[]` is also used for tuples

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -356,13 +359,23 @@ export function recmaJsxRewrite(options = {}) {
             }
 
             for (const [
-              invalidComponentName,
-              variable
-            ] of invalidComponentNameToVariable) {
+              _componentsId,
+              _componentName
+            ] of idToInvalidComponentName) {
```
wooorm, line 364: I still kinda like `name`, `value` better, but I’m not blocking on this.
I feel a bit more strongly about not using `_` to start variable names, that is often used with variables (parameters) to mark them as explicitly not used

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -356,13 +359,23 @@ export function recmaJsxRewrite(options = {}) {
...
+              // For component IDs that render invalid JSX (ex. <_components.custom-element>),
+              // Generate a separate variable to index into `_components`
+              // i.e. `const _component0 = _components['custom-element']`
+              //      `return <_component0>...`
```
wooorm, line 368:
```suggestion
              // For JSX IDs that can’t be represented as JavaScript IDs (as in,
              // those with dashes, such as `custom-element`), generate a
              // separate variable that is a valid JS ID (such as `_component0`),
              // and takes it from components:
              // `const _component0 = _components['custom-element']`
```

Taking your phrasing, I feel this suggestion is a bit more correct in how it is worded.
It also replaces Latin abbreviations with English words, which I believe makes them better readable to more people

packages/mdx/test/compile.js
```diff
@@ -858,11 +858,13 @@ test('jsx', async () => {
...
       'function _createMdxContent(props) {',
-      '  const _components = props.components || ({}), _component0 = "a-b";',
+      '  const _components = Object.assign({',
+      '    "a-b": "a-b"',
+      '  }, props.components), _component0 = _components["a-b"];',
       '  return <>{<_component0></_component0>}</>;',
       '}',
       'function MDXContent(props = {}) {',
-      '  const _component0 = "a-b", {wrapper: MDXLayout} = props.components || ({});',
+      '  const _component0 = _components["a-b"], {wrapper: MDXLayout} = props.components || ({});',
```
wooorm, line 867: Something seems to go wrong here in `MDXContent`, which only needs the wrapper. `_component0` does not need to be retrieved? I don’t believe anything other than the wrapper needs to be retrieved.

It also looks like this would throw: `_components` isn’t defined yet! Perhaps it is a good addition to test “invalid identifiers” can run too (here: https://github.com/mdx-js/mdx/blob/69a15b76f15136a33f2f9fe947a605e97d28fc6c/packages/mdx/test/evaluate.js#L328)?

wooorm (commented):
Getting very close, thanks for your continued work! One big thing, rest are nits

packages/mdx/test/compile.js
wooorm: Looks like it indeed throws, here: https://github.com/mdx-js/mdx/runs/7842620509?check_suite_focus=true#step:5:31.

bholmesdev: @wooorm Yep, I noticed that! Trying to find the code path that's injecting this variable inside layouts. Really want it to _not_ to that 😆

bholmesdev: Update: figured it out! No longer injects `const _component0` in layouts 👍

packages/mdx/lib/plugin/recma-jsx-rewrite.js
bholmesdev: Understood, removed those `_`s! I also tried using `name` and `value`, but noticed it leads to `{value: name}` and `{name: value}` in the sections where they're used. It... kinda made my head spin, so I chose `[id, componentName]` instead 😆

bholmesdev:
@wooorm Alright, I think we finally nailed it. Ready for review 😁

wooorm (approved):

wooorm:
Released in [2.1.3](https://github.com/mdx-js/mdx/releases/tag/2.1.3)!

bholmesdev:
@wooorm Woo! Thanks for the thorough review and feedback here. Already learned a ton about the MDX project ❤️

PR #2138, opened 2022-09-28
Use remark instead of rehype for processing meta field

rishi-raj-jain:
As the suggested approach for using meta field for syntax highlighting involves [remark](https://mdxjs.com/guides/syntax-highlighting/#syntax-highlighting-at-compile-time:~:text=But%20what%20if%20you%20want%20to%20access%20meta%3F%20The%20short%20answer%20is%3A%20use%20remark%2Dmdx%2Dcode%2Dmeta.%20It%20lets%20you%20type%20JSX%20attributes%20in%20the%20meta%20part%20which%20you%20can%20access%20by%20with%20a%20component%20for%20pre), I wanted to make the plugin recommended [here](https://mdxjs.com/guides/syntax-highlighting/#syntax-highlighting-at-compile-time:~:text=The%20long%20answer%20is%3A%20do%20it%20yourself%2C%20however%20you%20want%2C%20by%20writing%20a%20custom%20plugin%20to%20interpret%20the%20meta%20field.%20For%20example%2C%20it%E2%80%99s%20possible%20to%20pass%20that%20string%20as%20a%20prop%20with%20a%20rehype%20plugin%3A) consistent with marked as a `remark` plugin instead of a `rehype` plugin. [remark](https://github.com/remarkjs/remark) is the markdown processor (where processing meta field would be correct imo) while [rehype is the html processor](https://github.com/rehypejs/rehype). Hope this makes sense.


wooorm:
This doesn’t work. remark does not have elements. remark does not have attributes.
rehype has attributes (properties), so they can be set there, which is why this is a rehype plugin

rishi-raj-jain:
@wooorm

Then why is [remark-mdx-code-meta](https://github.com/remcohaszing/remark-mdx-code-meta/blob/main/index.ts) is used as a plugin for creating meta field inside code tags?

rishi-raj-jain:
Basically, why not rehype-mdx-code-meta then?

wooorm:
> Then why is [remark-mdx-code-meta](https://github.com/remcohaszing/remark-mdx-code-meta/blob/main/index.ts?rgh-link-date=2022-09-28T07%3A39%3A33Z) is used as a plugin for creating meta field inside code tags?

I don’t understand this question.

> Basically, why not rehype-mdx-code-meta then?

It doesn’t work with HTML elements or attributes. So there is no reason for it to be in the rehype space.

It generates a JSX element in a JS expression. It does not relate to “markdown” or “HTML” things in that sense, only to JavaScript. 

It can do that in the remark or in the rehype space.

wooorm:
> I'd say remark has elements

No.

> rehype doesn't have them

No. `rehype` definitely has element. HTML has elements.

rishi-raj-jain:
Ah HTML elements, yes rehype would have agreed.

rishi-raj-jain:
Wait so I'm super confused now between when to use remark / rehype?

What I'm on right now is that:
- remark: takes in markdown -> plugins alter the markdown -> form the html sent down to rehype
- rehype: takes in HTML (doesn't know what was in the markdown originally) -> plugins alter the html (received in Step 1) 

Is this right?

If so, what I'm saying is:

- on the page (attached screenshot below) recommends usage of remark-code-mdx-meta for preserving the meta field. now that works fine and is sync with what I understand.
Screenshot 2022-09-28 at 1 34 03 PM

- the line: "The long answer is: do it yourself, however you want, by writing a custom plugin to interpret the meta field. For example, it’s possible to pass that string as a prop with a rehype plugin:",
[is possibly creating a confusion](https://github.com/withastro/astro/issues/4616#issuecomment-1260173051) that **instead** of remark-code-mdx-meta one can do this (the rehype plugin) to achieve the same, while I think using remark-code-mdx-meta is a necessary condition to process meta tag as however with rehype.

So now I think you're right, it's rehype only. Just need a NOTE to be there that says for processing meta field with rehype, you need to have meta sent down in the html, for example via remark-mdx-code-meta.

rishi-raj-jain:
Appreciate your patience!

Lmk if this makes sense, and I'll modify the PR to include a NOTE instead

wooorm:
> Wait so I'm super confused now between when to use remark / rehype?

More here: https://github.com/remarkjs/remark-rehype#what-is-this.
And also here: https://mdxjs.com/packages/mdx/#architecture.

> What I'm on right now is that:

Yup, that’s a good description.

> > "The long answer is: […] For example, it’s possible to pass that string as a prop with a rehype plugin:",

It’s one of several possible solutions.
`meta` exists in markdown, so you could handle it there, but you’d like want to use it somewhere in HTML, right? In an attribute of a normal HTML element, or a prop to a JSX elements. If HTML element, then that’s HTML/rehype/hast space. If JSX, it might not matter.

> while I think using remark-code-mdx-meta is a necessary condition to process meta tag as however with rehype.

I think there are several solutions. The meta attribute is not standardized. It could have any format your authors decide on.
You have it contain HTML-like attributes (which is what the inline example does), that suites the HTML/rehype/hast space.
You could treat it like JSX attributes, which only makes sense in MDX, not in normal markdown. That‘s unrelated to the markdown or HTML spaces, it just has to be somewhere in MDX.

> you need to have meta sent down in the html

And this might be the missing piece in your thought process (which totally makes sense).
*Because* `meta` is completely ignored in markdown, and *because* people would likely want to use it to change HTML, it makes sense to pass the `meta` field through on nodes. And that’s what we do.
`meta`, when transforming between the two, is set as a “hidden” field on the HTML element: https://github.com/syntax-tree/mdast-util-to-hast/blob/7d20e86fbb0d8517497a0651dfcb229a6091a0c7/lib/handlers/code.js#L27-L29. It’s not output. But you can handle that HTML element if you want and use it.

wooorm:
Ping! Any follow up Qs? Things you want to work on?

rishi-raj-jain:
@wooorm Hey, will be back on this tomorrow

rishi-raj-jain:
@wooorm Does adding a NOTE that "For processing meta field with rehype, you need to have meta sent down in the html, for example via remark-mdx-code-meta." sounds good?

rishi-raj-jain:
@wooorm Ah I see. I think we can close this then.

PR #2141, opened 2022-10-01
fix: mark `webpack` as optional peer dependency

balazsorban44:
Next.js comes with pre-bundled `webpack` and it shows a warning when installing `@mdx-js/loader`.

This PR also addresses the comment here: https://github.com/vercel/next.js/pull/39167/files#r939687561

### DIFF
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -49,6 +49,11 @@
   "peerDependencies": {
     "webpack": ">=4"
   },
+  "peerDependenciesMeta": {
+    "webpack": {
+       "optional": true
+     }
+  },
   "devDependencies": {
     "@types/loader-utils": "^2.0.0",
     "@types/react": "^18.0.0",


remcohaszing (approved):
I'm not necessarily for or against this change. The code doesn't break if the `webpack` package isn't available, so it should be fine.

wooorm:
Not strongly against this either, but it feels a bit weird though. You’ll run into this problem with *any* webpack plugin.
Similarly, what if Next.js starts bundling React, then it’ll result in every React component library ever to start emitting a warning.

remcohaszing:
The difference is that React code imports `'react'`, whereas Webpack loaders just adhere to an interface defined by Webpack. I would be against this change if it would import `'webpack'`.

wooorm:
> React code imports `'react'`

If that’s the case (package `a` imports package `b`), I’d generally say it should be an explicit dependency. Peer dependencies are for package `a` is responsible for loading and using package `b`. Maybe React here is a complex example though (similar to unified for the AST types).
Although, this loader also imports `webpack`’s types: https://github.com/mdx-js/mdx/blob/9d2aa80ba00e68130763235a4d12902a5cfe017b/packages/loader/index.d.ts#L3, https://github.com/mdx-js/mdx/blob/9d2aa80ba00e68130763235a4d12902a5cfe017b/packages/loader/lib/index.js#L8-L9.

remcohaszing:
> > React code imports `'react'`
> 
> If that’s the case (package `a` imports package `b`), I’d generally say it should be an explicit dependency. Peer dependencies are for package `a` is responsible for loading and using package `b` (so `b` specifies a peer dependency on which min version of package `a` is supported). Maybe React here is a complex example though (similar to unified for the AST types).

Peer dependencies are ideal for packages that need to make sure the same instance of a package is loaded and reused. This is often framework code, such as `react` or `angular`, or any plugin based library that requires a plugin to import the library.

> Although, this loader also imports `webpack`’s types:

Yes, this package imports Webpack’s types, which would also make it a (required) peer dependency. However, the types of this package are unused. It’s consumed by referencing the package using a string, not by importing it.

wooorm:
> However, the types of this package are unused

People import the types of this plugins options so as to get type checking for their webpack config.

ChristianMurphy:
> [it] shows a warning when installing @mdx-js/loader.

What is "it" in this context?
npm/yarn, next itself, or something else?

And to confirm I understand correctly, it is a warning message only, it doesn't break next's ability to use mdx?

---

The code looks fine.

But my initial take would be mildly against doing this.
It seems a decision by next to bundle cwebpack rather than leverage it as a dependency has put them at odds with most other webpack projects.

And if it comes to a choice between mildly inconvenience for people using webpack in a standard way, vs mild inconvenience for a project which chooses to use webpack in an a highly custom and unusual way. (as next appears to).
I'd rather the standard usage be more seemless.

remcohaszing:
> > However, the types of this package are unused
> 
> People import the types of this plugins options so as to get type checking for their webpack config.

I haven’t really seen this much in practice, but you’re right, they could. I do think these people probably already have `webpack` installed anyway, so it shouldn’t be a problem.

---

> > [it] shows a warning when installing @mdx-js/loader.
> 
> What is "it" in this context? npm/yarn, next itself, or something else?
>
> And to confirm I understand correctly, it is a warning message only, it doesn't break next's ability to use mdx?

Yarn and npm with `--legacy-peer-deps` warn on this. The current version of npm installs peer dependencies, makes sure a resolution is found that satisfies all peer dependency constraints, and errors if there isn’t.

> The code looks fine.
> 
> But my initial take would be mildly against doing this. It seems a decision by next to bundle webpack rather than leverage it as a dependency has put them at odds with most other webpack projects.
> 
> And if it comes to a choice between mildly inconvenience for people using webpack in a standard way, vs mild inconvenience for a project which chooses to use webpack in an a highly custom and unusual way. (as next appears to). I'd rather the standard usage be more seemless.

I agree with this rationale 100%, but it doesn’t really change that I’m not strongly against this change.

wooorm:
> I haven’t really seen this much in practice

I’ve seen in tweeted and it’s in our examples: https://mdxjs.com/docs/getting-started/#webpack.

>  I do think these people probably already have webpack installed anyway

Why would you think that?

wooorm:
@balazsorban44 What do you think of the above comments? Wouldn’t any other webpack plugin have the same problem in Next?

helmturner:
👋 Hey there. Long time listener; first time caller.

Not strongly for or against this, but I might recommend adding a prominent note about this in the docs.

I'm developing a recma plugin specifically for use with MDX-loader in Next, and I've run into a cryptic 'webpack build error' that's absent during development. The 'unmet peer dependency' warning had me chasing a wild goose for a bit.

wooorm:
> specifically for use with MDX-loader in Next

What is so special about the custom webpack loader in Next.js that it needs special handling compared to “normal” webpack?
Or are you making something specific to Next.js (e.g., [`recma-nextjs-static-props`](https://github.com/remcohaszing/recma-nextjs-static-props))?

> but I might recommend adding a prominent note about this in the docs.

Where would that note be placed and what would it contain, to help a future reader like you?

wooorm:
I think this can be closed? This package does use `webpack`, it’s not really optional (there’s no handling for it missing). So as far as I understand this is correct for most people?

PR #2148, opened 2022-10-07
Fix casing of babel option in `getting-started.server.mdx`

kochie:
Fixes a small typo in the babel config where the `sourceFilename` is actually `sourceFileName` in the latest babel version.

### DIFF
--- a/docs/docs/getting-started.server.mdx
+++ b/docs/docs/getting-started.server.mdx
@@ -434,12 +434,12 @@ if you’re using that, for more info.
   // other things through to the normal Babel parser.
   function babelParserWithMdx(value, options) {
     if (
-      options.sourceFilename &&
-      /\.mdx?$/.test(path.extname(options.sourceFilename))
+      options.sourceFileName &&
+      /\.mdx?$/.test(path.extname(options.sourceFileName))
     ) {
       // Babel does not support async parsers, unfortunately.
       return compileSync(
-        {value, path: options.sourceFilename},
+        {value, path: options.sourceFileName},
         // Tell `@mdx-js/mdx` to return a Babel tree instead of serialized JS.
         {recmaPlugins: [recmaBabel], /* jsxImportSource: …, otherOptions… */}
       ).result


wooorm:
Can you confirm that this gave a warning, or didn’t work at all?

I remember running into babel using two casings of I believe this word. I think I raised an issue with them and that it’s been solved.
But wanted to verify!

kochie:
Yeah I just wrote a transformer for jest and needed to change the case to get it working.

```js
// import { compile, compileSync } from '@mdx-js/mdx'
import babel from '@babel/core'

import path from 'path'
import parser from '@babel/parser'
import estreeToBabel from 'estree-to-babel'
import { compileSync } from '@mdx-js/mdx'

export function babelPluginSyntaxMdx() {
  // Tell Babel to use a different parser.
  return { parserOverride: babelParserWithMdx }
}

// A Babel parser that parses MDX files with `@mdx-js/mdx` and passes any
// other things through to the normal Babel parser.
function babelParserWithMdx(value, options) {
  console.log(
    options
    // /\.mdx?$/.test(path.extname(options.sourceFilename))
  )
  if (
    options.sourceFileName &&
    /\.mdx?$/.test(path.extname(options.sourceFileName))
  ) {
    // Babel does not support async parsers, unfortunately.
    return compileSync(
      { value, path: options.sourceFileName },
      // Tell `@mdx-js/mdx` to return a Babel tree instead of serialized JS.
      {
        recmaPlugins: [recmaBabel] /* jsxImportSource: …, otherOptions… */,
        // outputFormat: 'function-body',
      }
    ).result
  }

  return parser.parse(value, options)
}

// A “recma” plugin is a unified plugin that runs on the estree (used by
// `@mdx-js/mdx` and much of the JS ecosystem but not Babel).
// This plugin defines `'estree-to-babel'` as the compiler, which means that
// the resulting Babel tree is given back by `compileSync`.
function recmaBabel() {
  Object.assign(this, { Compiler: estreeToBabel })
}

export default {
  process(sourceText, sourcePath, options) {

    const babelRes = babel.transform(sourceText, {
      filename: sourcePath,

      presets: ['@babel/preset-env'],
      plugins: [babelPluginSyntaxMdx],
    })

    console.log(babelRes.code)

    return babelRes
  },
}

```

I should point out I'm using jest with esm modules enabled, not sure if that would cause any difference.
`jest@29.1.2`
`@mdx-js/mdx@2.1.4`
`@babel/core@7.19.3`

wooorm (approved):

wooorm:
Cool, thanks!

PR #2169, opened 2022-11-07
v1/loader: Upgrade loader-utils to ^2.0.4

72636c:
The pinned version on 1.x is flagged with security vulnerabilities.

1. https://security.snyk.io/vuln/SNYK-JS-LOADERUTILS-3043105
2. https://security.snyk.io/vuln/SNYK-JS-LOADERUTILS-3105943

### DIFF
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -34,6 +34,6 @@
   "dependencies": {
     "@mdx-js/mdx": "1.6.22",
     "@mdx-js/react": "1.6.22",
-    "loader-utils": "^2.0.3"
+    "loader-utils": "^2.0.4"
   }
 }
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -34,6 +34,6 @@
   "dependencies": {
     "@mdx-js/mdx": "1.6.22",
     "@mdx-js/react": "1.6.22",
-    "loader-utils": "2.0.0"
+    "loader-utils": "^2.0.3"
   }
 }


ChristianMurphy:
Appreciate the time and consideration @72636c!

Taking a step back.
Please take a moment to read https://mdxjs.com/docs/getting-started/#security and https://overreacted.io/npm-audit-broken-by-design (which also applies to Snyk and many other dependency auditing tools)

MDX is a full programming language, as with any programming language, _you need to trust the authors creating code_.
For this to be a security risk, untrusted and maliciously authored code would be running.
Which is inherently unsafe both with and without this change.

My initial take would be it's not worth changing.
MDX 1 was somewhat notorious for breaking on patch/minor versions of dependencies (all the versions are locked for that reason).
I'd rather not risk breaking downstream packages, for something which isn't really a vulnerability in the context of MDX.

Also note, that MDX 2 resolved the dependency locking issue, uses version ranges, has many upgraded dependencies, and many performance/stability improvements overall.
I'd highly recommend upgrading to MDX version 2, and if MDX is being provided through another tool, encouraging them to upgrade to MDX version 2.

Open to further discussion if folks strongly disagree on applying this change to MDX 1 and would like to share their reasoning.
Closing this out for now.

72636c:
Thanks for taking the time to review the change @ChristianMurphy. I definitely understand the MDX stance here, but I would note that this feels like a pragmatic change that would benefit developers in organisations that subscribe to such supply chain security solutions, even if they aren't a silver bullet.

I'd love to adopt MDX V2 but the ESM requirement is a similarly challenging hurdle to overcome.

ChristianMurphy:
> I definitely understand the MDX stance here, but I would note that this feels like a pragmatic change that would benefit developers in organisations that subscribe to such supply chain security solutions, even if they aren't a silver bullet.

I understand and appreciate that.
A few considerations worth noting here.

First, there are some dependencies in MDX 1 which will be flagged by dependency chain tools which cannot feasibly be fixed, notably `trim` https://github.com/mdx-js/mdx/discussions/2150#discussioncomment-3842113

Second, this may, and likely does, have unintended consequences which makes what feels like a quick and insignificant change, more onerous.
When upgrading a dependency, the entire dependency tree shifts, which will likely cause CI and/or downstream packages to fail.
CI hasn't flagged anything yet, because it hasn't actually tested the change.
CI checks the exact versions in the lockfile https://github.com/mdx-js/mdx/blob/c91b00c673bcf3e7c28b861fd692b69016026c45/.github/workflows/ci.yml#L38
This PR does not currently upgrade the lockfile, meaning the change isn't tested.

For additional context, last few dependency upgrades applied to version 1 took a long time to re-stabilize the dependency tree (in the order of weeks IIRC), and on release resulted in follow up issues which required further dependency changes which added even more time.
Which adds to my caution/concern with applying this change, and why I believe moving to MDX 2 is generally more time/cost effective for adopters.

> I'd love to adopt MDX V2 but the ESM requirement is a similarly challenging hurdle to overcome.

Many toolchains are now fairly good at supporting ESM.
And 12.8% of the top NPM packages are now ESM https://nitter.net/robpalmer2/status/1592232202577874944#m
For tools which don't support ESM by default, https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c offers some great pointers on how to configure them to support ESM.
Also feel free to reach out in [discussions](https://github.com/mdx-js/mdx/discussions) if you run into a specific roadblock you've run into, if you'd like to talk it through with the community.

Issue #2105, opened 2022-08-15
[TypeScript] @mdx-js/rollup doesn't match Vite 3's plugin types


guoyunhe:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

@mdx-js/rollup 2.1.2

### Steps to reproduce

Create a vite.config.ts with:

```ts
import mdx from '@mdx-js/rollup';
import react from '@vitejs/plugin-react';
import { UserConfig } from 'vite';

const config: UserConfig = {
    plugins: [
      react({ jsxRuntime: 'classic' }),
      mdx({
        jsxRuntime: 'classic',
        providerImportSource: '@mdx-js/react',
      }) as any,
    ],
}

export default config;
```

### Expected behavior

TypeScript should not show errors

### Actual behavior

TypeScript shows errors:

```
error TS2322: Type 'Plugin' is not assignable to type 'PluginOption'.
  Type 'Plugin' is not assignable to type 'Plugin_2'.
    Types of property 'resolveId' are incompatible.
      Type 'ObjectHook<(this: PluginContext, source: string, importer: string | undefined, options: { custom?: CustomPluginOptions | undefined; isEntry: boolean; }) => ResolveIdResult | Promise<...>, {}> | undefined' is not assignable to type '((this: PluginContext, source: string, importer: string | undefined, options: { custom?: CustomPluginOptions | undefined; ssr?: boolean | undefined; isEntry: boolean; }) => ResolveIdResult | Promise<...>) | undefined'.
        Type '{ handler: (this: PluginContext, source: string, importer: string | undefined, options: { custom?: CustomPluginOptions | undefined; isEntry: boolean; }) => ResolveIdResult | Promise<...>; order?: "pre" | ... 2 more ... | undefined; }' is not assignable to type '(this: PluginContext, source: string, importer: string | undefined, options: { custom?: CustomPluginOptions | undefined; ssr?: boolean | undefined; isEntry: boolean; }) => ResolveIdResult | Promise<...>'.
[...]

PR #2180, opened 2022-11-23
[Vite] Add  note `info: peerDependency requirement`

ryuujo1573:
Type conflicts can be caused by installing `rollup` with version `>=3` as peer dependency unconsciously.

Possible prevention of #2105

### DIFF
--- a/docs/docs/getting-started.server.mdx
+++ b/docs/docs/getting-started.server.mdx
@@ -346,14 +346,10 @@ If you use more modern JavaScript features than what your users support,
 [configure Vite’s
 `build.target`](https://vitejs.dev/guide/build.html#browser-compatibility).
 
-<details>
-  <summary>Possible Compatibility Issue</summary>
-  As for [vite](https://github.com/vitejs/vite) 3, `rollup@^2.79.1` is applied.
-  You might notice that there was a warning of peer dependency problem. Please
-  ignore it or add `rollup` manually to your devDependencies with 
-  **version `^2.79.1`**. This avoids possible incompatibility, for when the 
-  version of `rollup` is not specified, `rollup@>=3` will be install by default.
-</details>
+<Note type="info">
+  **note**: Vite 3 requires the older `rollup@<3`. Make sure `rollup@3` is not
+  installed.
+</Note>
 
 See also [¶ Rollup][rollup], which is used in Vite, and see [¶ Vue][vue], if
 you’re using that, for more info.
--- a/docs/docs/getting-started.server.mdx
+++ b/docs/docs/getting-started.server.mdx
@@ -347,8 +347,8 @@ If you use more modern JavaScript features than what your users support,
 `build.target`](https://vitejs.dev/guide/build.html#browser-compatibility).
 
 <Note type="info">
-  **note**: Vite 3 requires the older `rollup@<3`. Make sure `rollup@3` is not
-  installed.
+  **Note**: Vite 3 requires the older `rollup@2`.
+  Make sure `rollup@3` is not installed.
 </Note>
 
 See also [¶ Rollup][rollup], which is used in Vite, and see [¶ Vue][vue], if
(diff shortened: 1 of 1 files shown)


wooorm:
I think this is about pnpm + vite.
If pnpm doesn’t install peers for you, that’s fine, but we allow v2 and v3 of rollup as peers.
If vite wants a specific version, that’s on them.

I don’t think it’s worth it to maintain a list of which versions of Rollup that Vite supports here?

ryuujo1573:
I'm aware now that this is about `pnpm` + `vite`. 

> I don’t think it’s worth it to maintain a list of which versions of Rollup that Vite supports here?

Yes, bing is my friend, nevertheless it would be nice for newbees who might use this directly with `vite` + `pnpm`, which reduces some problem. 

Since `vite` was mentioned here, I might say that this tip provided as "possible conflicts" is of help? And this might not change unless `vite` to use `rollup@3`?
Just some ameteur opinions, please feel free to close this as won't-fix~

Thank you 🎵

wooorm:
> nevertheless it would be nice for newbees who might use this directly with vite + pnpm, which reduces some problem.

I think it’s inevitable that beginners will run into problems when trying to combine:

a) a pretty new rapidly developing package manager that does thing differently from how everything else works (pnpm)
b) a pretty new rapidly developing build tool that does thing different from how everything else works (Vite)
c) a best-effort type system (typescript)
d) anything else (mdx)

This is a problem with you not yet knowing how versions of packages work together (semver). How to install a previous version that some package wants instead of the latest. How to read pnpms error messages. How to silence a TS error. You will run into similar problems with any other tool.

That’s fine, totally understandable, and you’ll learn all that! But I think our docs will get worse for a lot of people if we try to answer all your questions in them.

> I might say that this tip provided as "possible conflicts" is of help

This is a problem for a small number of users.
Such a section is also hard to maintain: it means that I need to check Vite’s docs every month or so to see if they support Rollup 3 or not.
There is also the problem that we already have a lot of docs. It can get overwhelming. It needs to be succinct.

---

Now. The root of this, is that Vite (currently at v3) wants you to install Rollup 2. And Vite v4 will support Rollup 3. We can say that succinctly:

```mdx
<Note type="info">
  **Note**: Vite 3 (current) requires the older `rollup@2`.
  Vite 4 (beta) supports `rollup@3`.
</Note>
```

ryuujo1573:
Thanks a lot!
I'm now getting used to the abcs of docs maintainence. 🤔
hmmmm, I was wondering if it would be fine to open the PR like this.

I will learn to contribute smartly, and please let me express my appreciation once again for your clear explanation and instruction.
Yeeee.

docs/docs/getting-started.server.mdx
```diff
@@ -346,14 +346,10 @@ If you use more modern JavaScript features than what your users support,
...
-  <summary>Possible Compatibility Issue</summary>
-  As for [vite](https://github.com/vitejs/vite) 3, `rollup@^2.79.1` is applied.
-  You might notice that there was a warning of peer dependency problem. Please
-  ignore it or add `rollup` manually to your devDependencies with 
-  **version `^2.79.1`**. This avoids possible incompatibility, for when the 
-  version of `rollup` is not specified, `rollup@>=3` will be install by default.
-</details>
+<Note type="info">
+  **note**: Vite 3 requires the older `rollup@<3`. Make sure `rollup@3` is not
+  installed.
```
wooorm, line 351:
```suggestion
  **note**: Vite 3 requires the older `rollup@2`.
  Make sure `rollup@3` is not installed.
```

a) it specifically needs v2
b) Fix formatting

Otherwise good!

ryuujo1573: Sorry, I've just turned to vscode for linting possible format issues. And sadly a force push to erase typo 😂

ryuujo1573: All is well 😜

wooorm (approved):

wooorm:
Thank you, @ryuujo1573!!

PR #2201, opened 2022-12-23
Add inferring of `development` options from webpack loader mode

### LINKED ISSUE #2198: Automatic development mode in Webpack loader
### Initial checklist

### Problem

Webpack supports a [`mode`](https://webpack.js.org/configuration/mode/) setting, which can be set to `production`, `development`, or `none`. MDX supports a boolean `development`, but does not use webpack’s `mode`.

It would make sense to set the MDX `development` option based on Webpack’s `mode`.

### Solution

Default the MDX `development` option to Webpack’s `mode === 'development'`. If `development` is specified explicitly in the loader options, that should take precedence.

### Alternatives

We could also use `mode !== 'production'`, which would change the behaviour of the `'none'` mode, but I think it’s best if `'none'` would use the default value provided by `@mdx-js/mdx`.

remcohaszing:
Closes #2198

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -35,7 +35,10 @@ const cache = new WeakMap()
 export function loader(value, callback) {
   /** @type {Defaults} */
   const defaults = this.sourceMap ? {SourceMapGenerator} : {}
-  const options = /** @type {CompileOptions} */ (this.getOptions())
+  const options = {
+    development: this.mode === 'development',
+    .../** @type {CompileOptions} */ (this.getOptions())
+  }
   const config = {...defaults, ...options}
   const hash = getOptionsHash(options)
   // Some loaders set `undefined` (see `TypeStrong/ts-loader`).
--- a/packages/loader/test/index.test.js
+++ b/packages/loader/test/index.test.js
@@ -138,6 +138,12 @@ webpack.mdx:1:22: Unexpected end of file in expression, expected a corresponding
     'should compile (preact)'
   )
 
+  assert.ok(
+    // @ts-expect-error Type definitions don’t define __source.
+    modPreact.default.default({}).__source,
+    'should infer the development option from webpack’s mode'
+  )
+
   assert.match(
     String(await fs.readFile(new URL('preact.cjs', base))),
     /\/\/# sourceMappingURL/,
--- a/packages/loader/test/index.test.js
+++ b/packages/loader/test/index.test.js
@@ -95,6 +95,13 @@ webpack.mdx:1:22: Unexpected end of file in expression, expected a corresponding
     'should compile (react)'
   )
 
+  const reactOutput = await fs.readFile(new URL('react.cjs', base), 'utf8')
+  assert.not.match(
+    reactOutput,
+    /react_jsx_dev_runtime__WEBPACK_IMPORTED_MODULE_\d+__\.jsxDEV/,
+    'should infer the development option from webpack’s production mode'
+  )
+
   await fs.unlink(new URL('react.cjs', base))
 
   // Preact and source maps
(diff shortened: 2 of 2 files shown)


packages/loader/test/index.test.js
```diff
@@ -138,6 +138,12 @@ webpack.mdx:1:22: Unexpected end of file in expression, expected a corresponding
     'should compile (preact)'
   )
 
+  assert.ok(
+    // @ts-expect-error Type definitions don’t define __source.
+    modPreact.default.default({}).__source,
```
ChristianMurphy, line 143: I'm not sure I follow this test case.
Does this property only exist in development mode?
If so, would we need a test to show it is missing in production mode?

ChristianMurphy (approved):
Loader change looks good 👍 
Have a minor question on the test case.

packages/loader/test/index.test.js
remcohaszing: Yes, this property only exists in development mode. Currently a webpack build with mode `'none'` is created for React, and one with mode `'development'` for Preact. Preact attaches the `__source` property to the node, React does not (not even in development). I guess React tracks this somewhere internally.

ChristianMurphy: Interesting 🤔 
Would it make sense to also have a test in production mode, and verify the property is missing? Or is that assumed/implied by a test elsewhere?

remcohaszing: It’s also possible to read the compiled output and look for certain patterns. It’s not very elegant IMO, but it could work.

wooorm: I think there should be an explanation next to this test: I agree with Christian that reading this test code is confusing.

The `reactBuild` has `mode: 'none'`, so there it should use a “production” automatic jsx import, and here we should have a development import, perhaps we can `t.match` for those?

packages/loader/lib/index.js
```diff
@@ -35,7 +35,10 @@ const cache = new WeakMap()
 export function loader(value, callback) {
   /** @type {Defaults} */
   const defaults = this.sourceMap ? {SourceMapGenerator} : {}
-  const options = /** @type {CompileOptions} */ (this.getOptions())
+  const options = {
+    development: this.mode === 'development',
+    .../** @type {CompileOptions} */ (this.getOptions())
+  }
```
wooorm, line 41: From reading [the docs](https://webpack.js.org/configuration/mode/), it seems like `mode: development` sets `process.env.NODE_ENV`? If that’s the case, shouldn’t we already be picking up on this?

remcohaszing: No, it uses `DefinePlugin` to define `process.env.NODE_ENV` in the bundle. It doesn’t set the actual environment variable, so the loader doesn’t pick it up.

I juse tried it using the following setup:

```js
// webpack.config.js
export default {
  module: {
    rules: [
      { use: './loader.cjs' }
    ]
  }
}
```

```js
// loader.cjs
module.exports = (content) => {
  console.log(process.env.NODE_ENV)
  return 'console.log("loader");' + content
}
```

```js
// index.js
console.log(process.env.NODE_ENV)
```

Now building yields:

```
$ webpack build --mode production
process.env.NODE_ENV undefined
asset main.js 49 bytes [emitted] [minimized] (name: main)
./src/index.js 33 bytes [built] [code generated]
webpack 5.75.0 compiled successfully in 153 ms
```

And this is the build output:

```js
(()=>{"use strict";console.log("production")})();
```

packages/loader/test/index.test.js
remcohaszing: I changed it to use `assert.match` now. It’s a bit annoying to find a propert string to match, as output changes significantly with the mode, but I found an ok regex to test against.

wooorm (approved):

wooorm:
Thanks Remco!

wooorm:
Thanks, released in [`2.3.0`](https://github.com/mdx-js/mdx/releases/tag/2.3.0)

PR #2204, opened 2022-12-25
[docs] Update on-demand docs to avoid DEV JSX runtime

joostdecock:
The example code above will (since version 2.2) throw an error because `_jsxDEV is not a function`.

To avoid that, we need to pass the `development: false` option, so I've updated these docs to do so.

See: https://github.com/mdx-js/mdx/discussions/2203

### DIFF
--- a/docs/guides/mdx-on-demand.server.mdx
+++ b/docs/guides/mdx-on-demand.server.mdx
@@ -28,7 +28,11 @@ On the server:
 ```js path="server.js"
 import {compile} from '@mdx-js/mdx'
 
-const code = String(await compile('# hi', {outputFormat: 'function-body' /* …otherOptions */ }))
+const code = String(await compile('# hi', {
+  outputFormat: 'function-body',
+  development: false, // Avoid use of the jsxDEV runtime
+  /* …otherOptions */ 
+}))
 // To do: send `code` to the client somehow.
 ```
 
@@ -82,7 +86,11 @@ export default function Page({code}) {
 
 export async function getStaticProps() {
   const code = String(
-    await compile('# hi', {outputFormat: 'function-body' /* …otherOptions */})
+    await compile('# hi', {
+      outputFormat: 'function-body',
+      development: false, // Avoid use of the jsxDEV runtime
+      /* …otherOptions */
+    })
   )
   return {props: {code}}
 }


docs/guides/mdx-on-demand.server.mdx
```diff
@@ -28,7 +28,11 @@ On the server:
 ```js path="server.js"
 
-const code = String(await compile('# hi', {outputFormat: 'function-body' /* …otherOptions */ }))
+const code = String(await compile('# hi', {
+  outputFormat: 'function-body',
+  development: false, // Avoid use of the jsxDEV runtime
```
remcohaszing, line 33: I’m thinking maybe something like this? It’s a bit annoying, because it’s all very situational.

```suggestion
  // development: false, // Use this to avoid use of jsx-dev-runtime during development
```

Maybe for the client we can use:

```js
import {run} from '@mdx-js/mdx'
import * as devRuntime from 'react/jsx-dev-runtime'

const code = '' // To do: get `code` from server somehow.
const deveopment = process.env.NODE_ENV // To do: determine development mode for your setup.

const {default: Content} = await run(code, development ? devRuntime : runtime)
```

wooorm, line 33:
```suggestion
  development: false, // Generate code for production.
```

I’d suggest this: better to document what this does, instead of double negatives.

Alternatively, you can show both dev and prod versions

```suggestion
  development: false, // Generate code for production.
  // developmment: true, // Pass this to generate development code.
```

And then use the following below:

```
import * as runtime from 'react/jsx-runtime' // Production.
// import * as runtime from 'react/jsx-dev-runtime' // Development.
```

docs/guides/mdx-on-demand.server.mdx
```diff
@@ -82,7 +86,11 @@ export default function Page({code}) {
 
 export async function getStaticProps() {
   const code = String(
-    await compile('# hi', {outputFormat: 'function-body' /* …otherOptions */})
+    await compile('# hi', {
+      outputFormat: 'function-body',
+      development: false, // Avoid use of the jsxDEV runtime
```
wooorm, line 91: Same as above

wooorm (commented):
There are many more places where `run` is shown. Perhaps good to extend what you want to change here, to those places too, as it applies to them too?

PR #2211, opened 2022-12-31
Add js extension to mdx/types import

Gerrit0:
Because `@mdx-js/mdx`'s package.json includes `"type": "module"`, imports within it which reference a file must include file extensions, including type only imports. This breaks downstream TypeScript users who have `"module": "Node16"` in their `tsconfig.json` file.

```
node_modules/@mdx-js/mdx/lib/evaluate.d.ts:23:32 - error TS2307: Cannot find module 'mdx/types' 
  or its corresponding type declarations.

23 export type ExportMap = import('mdx/types').MDXModule;
                                  ~~~~~~~~~~~
```

Yes, the `.js` looks odd since we're importing a declaration file from `@types/mdx` here, but it is correct, since in a TS-native setup, it would be describing a `.js` file alongside the declaration file.

cc: @gaurishh

### DIFF
--- a/packages/esbuild/test/index.test.js
+++ b/packages/esbuild/test/index.test.js
@@ -3,7 +3,7 @@
  * @typedef {import('esbuild').Message} Message
  * @typedef {import('hast').Root} Root
  * @typedef {import('vfile').VFile} VFile
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
  *
  * @typedef {import('remark-mdx')}
  */
--- a/packages/loader/test/index.test.js
+++ b/packages/loader/test/index.test.js
@@ -1,5 +1,5 @@
 /**
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
  * @typedef {import('preact').FunctionComponent<unknown>} PreactComponent
  * @typedef {import('vue').Component} VueComponent
  * @typedef {import('vue').SetupContext} SetupContext
--- a/packages/mdx/lib/evaluate.js
+++ b/packages/mdx/lib/evaluate.js
@@ -2,7 +2,7 @@
  * @typedef {import('vfile').VFileCompatible} VFileCompatible
  * @typedef {import('./util/resolve-evaluate-options.js').EvaluateOptions} EvaluateOptions
  *
- * @typedef {import('mdx/types').MDXModule} ExportMap
+ * @typedef {import('mdx/types.js').MDXModule} ExportMap
  */
 
 import {compile, compileSync} from './compile.js'
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -1,6 +1,6 @@
 /**
- * @typedef {import('mdx/types').MDXModule} MDXModule
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXModule} MDXModule
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
  * @typedef {import('hast').Root} Root
  * @typedef {import('../lib/compile.js').VFileCompatible} VFileCompatible
  */
--- a/packages/node-loader/test/index.test.js
+++ b/packages/node-loader/test/index.test.js
@@ -1,5 +1,5 @@
 /**
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
  */
 
 import {promises as fs} from 'fs'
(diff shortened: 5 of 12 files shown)


remcohaszing (approved):
Thanks! Nice catch!

ChristianMurphy (approved):
LGTM, thanks!
Would it also make sense to set the module type in the root configuration, so this is tested going forward?https://github.com/mdx-js/mdx/blob/600b12ade3c1cd5bcbca9afc160b8456aa33a82e/tsconfig.json#L2-L6

remcohaszing:
Adding on @ChristianMurphy’s comment, the correct configuration would be to replace the `module` value with `node16` and remove the `moduleResolution` and `allowSyntheticDefaultImports` options.

I also wouldn’t mind removing the redundant `allowJs` option and adding `"forceConsistentCasingInFileNames": true` and `"newLine": "lf"` while we’re fixing up `tsconfig.json`.

wooorm:
I agree that `module: node16` is a good idea.

Other refactoring of `tsconfig` should be a separate PR tho

Gerrit0:
... yeah, thought that might happen. Main reason I spot fixed the one piece breaking external consumers.

ChristianMurphy:
At least 15 of the errors are in the `lib/` folder of their respective project, the project implementation, and likely do impact downstream users.

Gerrit0:
... okay, so that wasn't some weird local issue... thought I just didn't know how to use workspaces

ChristianMurphy:
The `estree-walker` error is likely related to or the same as https://github.com/Rich-Harris/estree-walker/pull/31

Gerrit0:
I think I'll leave this here - don't have the time to spend a ton of time troubleshooting in an unfamiliar repo I don't use ;)

ChristianMurphy:
> unfamiliar repo I don't use

The errors are coming out of `@mdx-js/mdx` the one and only part of the repo you explicitly said you do use and originally edited. 🙂 

That said, with `estree-walker` and `periscopic` being blockers, it likely makes sense to walk `node16` back for a bit here, and nudge @Rich-Harris with some PRs to round out and fix ESM support upstream.

Gerrit0:
This PR is coming out of a question on the TypeScript Discord - I don't actually use `@mdx-js/mdx`, just helped someone else get around these errors then figured it should be easy to fix them upstream.

I agree - the Node16 changes would be nice, but until ESM works properly for those dependencies...

remcohaszing:
> That said, with `estree-walker` and `periscopic` being blockers, it likely makes sense to walk `node16` back for a bit here,

Yes, but this shouldn’t block this PR. Even if this doesn’t test it, it does fix the issue.

An alternative solution could be to add `exports` in `@types/mdx`.

wooorm:
What is blocking? I don’t see CI breaking?

remcohaszing:
Potentially the fact that we can't use and therefor can't test using `module: node16`. If I misinterpreted that, then never mind. 🙂

Regardless, it's still worth considering the alternative. I currently don't have a strong opinion about it.

Gerrit0:
>  the fact that we can't use and therefor can't test using module: node16

Yep, exactly

wooorm:
Oh wait, I missed that, I thought it was added here.
If `estree-walker` is failing, why not add `@ts-expect-error` for it? One of the benefits of only using TS for types, not for compiling. I did that here: https://github.com/syntax-tree/estree-util-build-jsx/commit/3c78c801e9ef7712c9b6b86bdfadcc5e7a39093e (L26)

Gerrit0:
Unfortunately doing that would result in two modules which currently have intellisense being stuck with `any`, which I suspect would be an overall worse developer experience here.

https://github.com/mdx-js/mdx/actions/runs/3823843079/jobs/6505439634#step:5:25

wooorm:
Internal types is fine, we have tests. They don’t reach users of this project

remcohaszing:
We could also use `patch-package` to patch `estree-walker`, since that's already used anyway.

Rich-Harris:
I merged https://github.com/Rich-Harris/estree-walker/pull/31 and released `estree-walker@3.0.2`, is there anything else I need to fix?

wooorm:
Nice, I don’t think so! Thanks Rich!

(while you’re here, there was another bug with export maps in periscopic/estree-walker/is-reference, it’s now fixed in Next.js/SWC, but might pop up in other compilers: https://github.com/mdx-js/mdx/discussions/2168#discussioncomment-4068254). TLDR: Node recommends always adding a `default` condition, which isn’t (wasn’t?) in these packages!

ChristianMurphy:
> is there anything else I need to fix?

Thanks @Rich-Harris! Much appreciated!
Potentially a few more tweaks could be merged for the typings to support the new `node16` resolution.

https://github.com/Rich-Harris/is-reference/pull/14
https://github.com/Rich-Harris/periscopic/pull/19

and one incoming for `extree-walker` as well
https://github.com/Rich-Harris/periscopic/pull/19

wooorm:
Done! :)

PR #2221, opened 2023-01-13
Add `remark-mdx-chartjs` to list of plugins

pangelani:


docs/docs/extending-mdx.server.mdx
```diff
@@ -81,6 +81,8 @@ See also the [list of remark plugins][remark-plugins] and
     — change frontmatter (YAML) metadata to exports
 *   [`goodproblems/remark-mdx-math-enhanced`](https://github.com/goodproblems/remark-mdx-math-enhanced)
     — enhance math with JavaScript inside it
+*   [`pangelani/remark-mdx-chartjs`](https://github.com/pangelani/remark-mdx-chartjs)
+    — replace fenced code blocks with charts using `react-chartjs-2`.
```
remcohaszing, line 85: I think it would be nice to add a link to `react-chartjs-2`.

```suggestion
    — replace fenced code blocks with charts using [`react-chartjs-2`](https://react-chartjs-2.js.org/).
```

remcohaszing (commented):
Thank you for creating this plugin / adding it here!

wooorm:
Instead of serializing as JSON and then parsing with acorn (which is slow and prone to breaking), it would be better to use something like https://github.com/remcohaszing/estree-util-value-to-estree!

remcohaszing:
I agree :100: with @wooorm’s tip. Also you can omit the `raw` properties and it’s ok to use empty strings as the value of `mdxjsEsm` nodes as those are unused.

As far as this PR is concerned though, looks good to me :+1:

remcohaszing (approved):

pangelani:
Thanks @wooorm and @remcohaszing! I added the suggested improvement to the pluginl

wooorm (approved):

PR #2228, opened 2023-01-16
[WIP] types: use node16 module resolution in typescript

ChristianMurphy:
Mostly ready, this will depend on some upstream changes being finished.
https://github.com/Rich-Harris/periscopic/pull/19 and https://github.com/Rich-Harris/estree-walker/pull/34

### DIFF
--- a/packages/esbuild/test/index.test.js
+++ b/packages/esbuild/test/index.test.js
@@ -3,7 +3,7 @@
  * @typedef {import('esbuild').Message} Message
  * @typedef {import('hast').Root} Root
  * @typedef {import('vfile').VFile} VFile
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
  *
  * @typedef {import('remark-mdx')}
  */
--- a/packages/loader/test/index.test.js
+++ b/packages/loader/test/index.test.js
@@ -1,5 +1,5 @@
 /**
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
  * @typedef {import('preact').FunctionComponent<unknown>} PreactComponent
  * @typedef {import('vue').Component} VueComponent
  * @typedef {import('vue').SetupContext} SetupContext
--- a/packages/mdx/lib/evaluate.js
+++ b/packages/mdx/lib/evaluate.js
@@ -2,7 +2,7 @@
  * @typedef {import('vfile').VFileCompatible} VFileCompatible
  * @typedef {import('./util/resolve-evaluate-options.js').EvaluateOptions} EvaluateOptions
  *
- * @typedef {import('mdx/types').MDXModule} ExportMap
+ * @typedef {import('mdx/types.js').MDXModule} ExportMap
  */
 
 import {compile, compileSync} from './compile.js'
--- a/packages/mdx/test/compile.js
+++ b/packages/mdx/test/compile.js
@@ -1,6 +1,6 @@
 /**
- * @typedef {import('mdx/types').MDXModule} MDXModule
- * @typedef {import('mdx/types').MDXContent} MDXContent
+ * @typedef {import('mdx/types.js').MDXModule} MDXModule
+ * @typedef {import('mdx/types.js').MDXContent} MDXContent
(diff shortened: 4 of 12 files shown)


ChristianMurphy:
Upstream issues resolved, this is ready for review @mdx-js/maintainers

packages/mdx/lib/plugin/rehype-remove-raw.js
```diff
@@ -1,5 +1,6 @@
 /**
  * @typedef {import('hast').Root} Root
+ * @typedef {import('mdast-util-to-hast').Raw} DoNotRemoveUsedToAddRawToNodeType
```
ChristianMurphy, line 3: @wooorm @remcohaszing do you have a trick for getting the raw node type to register itself?
I tried this, it doesn't seem to work.
The code below needs to be able to filter for the `raw` node type (a custom node added by this plugin) from a `hast` tree

wooorm: Maybe it registers without the `.Raw`? (https://github.com/syntax-tree/mdast-util-to-hast#types)

But I thought this does register it?

remcohaszing: `mdast-util-to-hast` needs to be loaded. THis can be done by either:

1. Using a TypeScript tripple slash directive (not sure if it works in JSDoc too)
   ```ts
   /// <reference types="mdast-util-to-hast" />
   ```
   This means the these types will also be loaded for people who use this file
2. Adding it to `"types"` in `tsconfig.json` compiler options.
   ```jsonc
   {
     // …
     "compilerOptions": {
       // …
       "types": ["mdast-util-to-hast"]
     }
   }
   ```

I recommend using the latter.

Regardless of this issue, I recommend using the `"types"` compiler option in a mono repo like this. It prevents TypeScript from loading unwanted type definitions from `@types` dependencies. E.g. without it, the TypeScript loads `@types/react` while checking `@mdx-js/preact` and `@mdx-js/vue`.

ChristianMurphy: Thanks for the suggestions!

🤔 I've tried adding

```json
  "compilerOptions": {
    "types": ["mdast-util-to-hast", "@types/node"]
  }
```
to _tsconfig.json_

as well as

```js
/**
 * @typedef {import('mdast-util-to-hast')} DoNotRemoveUsedToAddRawToNodeType
 */
```

both still show `raw` type as `never` 🤔 

---

Will keep trying different options.

ChristianMurphy: committed in bd4c814fe7b9ad756e29795aad4fce48f540ed53 for reference

wooorm: The problem is that `mdast-util-to-hast` was compiled before TS 4.9.
It uses `export type Raw = import('./complex-types').Raw` in there, which is broken, because there is no extension.
`mdast-util-to-hast` has now been published with types that were compiled with TS 4.9.
If you update the package lock (and rebase), this PR will work!

packages/mdx/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["lib/**/*.js", "test/**/*.js", "*.js"]
+  "include": ["lib/**/*.js", "test/**/*.js", "*.js"],
+  "compilerOptions": {
+    "types": ["mdast-util-to-hast", "@types/node"]
+  }
```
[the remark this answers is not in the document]
ChristianMurphy: Updated in [746afe3](https://github.com/mdx-js/mdx/pull/2228/commits/746afe3d07199af0d28b761da5bc572808096c83)

packages/mdx/lib/plugin/rehype-remove-raw.js
ChristianMurphy: Updated in [746afe3](https://github.com/mdx-js/mdx/pull/2228/commits/746afe3d07199af0d28b761da5bc572808096c83)

wooorm (approved):

wooorm:
Sweet, thanks!

wooorm:
Thanks, released in [`2.3.0`](https://github.com/mdx-js/mdx/releases/tag/2.3.0)

Issue #2247, opened 2023-02-08
Compatibility with Solid and other JSX flavors


DaniGuardiola:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

Currently this package is only compatible with React-like JSX flavors, for example by outputting `className` for CSS classes, and camel cased CSS properties in the style object..

### Solution

The underlying utility used in the recma plugin to create the jsx is `hast-util-to-estree`. This util has recently updated to support options to customize how it get rendered (for example, by using `class` instead of `className`, which is necessary for Solid).

These new options should be made available and passed through by this package, so that users can take advantage of them.

Context: https://github.com/syntax-tree/hast-util-to-estree/issues/7#issuecomment-1421953456

### Alternatives

None :)

PR #2255, opened 2023-02-09
Add improved support for non-React frameworks

wooorm:
React uses their own style of attributes (similar to the DOM), such as `className` instead of `class`.
Some ther frameworks don’t support that: they want you to pass `class`. Similarly, React wants `WebkitBoxShadow`, while other frameworks want `-webkit-box-shadow`.

This particularly becomes an issue around more complex things such as SVG.
With this commit, you can now configure the exact style that your particular framework accepts.

Closes GH-2247.

### DIFF
--- a/packages/mdx/lib/core.js
+++ b/packages/mdx/lib/core.js
@@ -2,6 +2,7 @@
  * @typedef {import('remark-rehype').Options} RemarkRehypeOptions
  * @typedef {import('unified').PluggableList} PluggableList
  * @typedef {import('unified').Processor} Processor
+ * @typedef {import('./plugin/rehype-recma.js').Options} RehypeRecmaOptions
  * @typedef {import('./plugin/recma-document.js').RecmaDocumentOptions} RecmaDocumentOptions
  * @typedef {import('./plugin/recma-stringify.js').RecmaStringifyOptions} RecmaStringifyOptions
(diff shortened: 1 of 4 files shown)


ChristianMurphy (approved):

DaniGuardiola:
@wooorm after upgrading to 2.3.0 in my project, this didn't work, because `hast-util-to-estree` was still on the 2.2.1 version. Should the `package.json` entry be updated to `"^2.3.0"` maybe?

DaniGuardiola:
Of course, running `pnpm update --latest` did bump it, but doesn't `"^2.3.0"` make it technically incompatible?

ChristianMurphy:
@DaniGuardiola update your lockfile.
`mdx` allows and encourages the latest of the current major version line.
https://github.com/mdx-js/mdx/blob/5f6833068368041f63dcf020b2fd578a4d6987a7/packages/mdx/package.json#L55

We intentionally do **NOT** bump to versions like `^2.3.0` to allow `npm` and `yarn` to de-duplicate more if an older version is used elsewhere.

That said, `pnpm` and `yarn` are known to try to unnecessarily lock older versions, prioritizing minimizing change above de-duplication.
In that case running the lockfile update command for the given package manager is the best approach to get the upgrade.

PR #2295, opened 2023-05-04
Add more options to playground, directives, format, etc

slorber:
This PR adds new useful options to have in the playground:

- format md: to easily test the output in CommonMark mode
- directive: to parse syntax such as `:::warning`
- rehype-raw: useful to reparse raw html string nodes to proper hast tree
- position: to easily strip noise from output AST

Note: I think it's better to not have positions in the AST by default, this creates useless noise and I doubt people use positions in the playground anyway.

CleanShot 2023-05-11 at 16 17 11

### DIFF
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -22,10 +22,23 @@ import {lowlight} from 'lowlight/lib/core.js'
 import javascript from 'highlight.js/lib/languages/javascript'
 import json from 'highlight.js/lib/languages/json'
 import markdown from 'highlight.js/lib/languages/markdown'
+import {visit as visitEstree} from 'estree-util-visit'
 
 lowlight.registerLanguage('js', javascript)
 lowlight.registerLanguage('json', json)
 lowlight.registerLanguage('md', markdown)
+
+export function removePositionEsast(tree) {
+  visitEstree(tree, remove)
+  return tree
+
+  function remove(node) {
+    delete node.loc
+    delete node.start
+    delete node.end
+    delete node.range
+  }
+}
 
 function useMdx(defaults) {
   const [state, setState] = useState({...defaults, file: null})
@@ -61,7 +74,7 @@ function useMdx(defaults) {
         if (!config.position) {
           removePosition(file.data.mdast, {force: true})
           removePosition(file.data.hast, {force: true})
-          removePosition(file.data.esast, {force: true})
+          removePositionEsast(file.data.esast)
         }
       } catch (error) {
         const message =
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -10,6 +10,7 @@ import remarkGfm from 'remark-gfm'
 import remarkFrontmatter from 'remark-frontmatter'
 import remarkDirective from 'remark-directive'
 import remarkMath from 'remark-math'
+import {removePosition} from 'unist-util-remove-position'
 import CodeMirror from 'rodemirror'
 import {basicSetup} from 'codemirror'
 import {markdown as langMarkdown} from '@codemirror/lang-markdown'
@@ -56,6 +57,12 @@ function useMdx(defaults) {
             recmaPlugins: [capture('esast')]
           })
         ).default
+
+        if (!config.position) {
+          removePosition(file.data.mdast, {force: true})
+          removePosition(file.data.hast, {force: true})
+          removePosition(file.data.esast, {force: true})
+        }
       } catch (error) {
         const message =
           error instanceof VFileMessage ? error : new VFileMessage(error)
@@ -108,6 +115,7 @@ export function Editor({children}) {
   const extensions = useMemo(() => [basicSetup, oneDark, langMarkdown()], [])
   const [state, setConfig] = useMdx({
     formatMd: false,
+    position: false,
     gfm: false,
     frontmatter: false,
     directive: false,
@@ -228,6 +236,19 @@ export function Editor({children}) {
               Use{' '}
               <a href="https://mdxjs.com/packages/mdx/#optionsformat">
                 <code>format: &apos;md&apos;</code>
+              </a>
+            </label>
+            <label>
+              <input
+                checked={state.position}
+                type="checkbox"
+                onChange={() =>
+                  setConfig({...state, position: !state.position})
+                }
+              />{' '}
+              Use{' '}
+              <a href="https://github.com/syntax-tree/unist#position">
+                <code>position</code>
               </a>
             </label>
           </form>
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -10,6 +10,7 @@ import remarkGfm from 'remark-gfm'
 import remarkFrontmatter from 'remark-frontmatter'
 import remarkDirective from 'remark-directive'
 import remarkMath from 'remark-math'
+import {visit as visitEstree} from 'estree-util-visit'
 import {removePosition} from 'unist-util-remove-position'
 import CodeMirror from 'rodemirror'
 import {basicSetup} from 'codemirror'
@@ -22,7 +23,6 @@ import {lowlight} from 'lowlight/lib/core.js'
 import javascript from 'highlight.js/lib/languages/javascript'
 import json from 'highlight.js/lib/languages/json'
 import markdown from 'highlight.js/lib/languages/markdown'
-import {visit as visitEstree} from 'estree-util-visit'
 
 lowlight.registerLanguage('js', javascript)
 lowlight.registerLanguage('json', json)
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -29,18 +29,6 @@ lowlight.registerLanguage('js', javascript)
 lowlight.registerLanguage('json', json)
 lowlight.registerLanguage('md', markdown)
 
-function removePositionEsast(tree) {
-  visitEstree(tree, remove)
-  return tree
-
-  function remove(node) {
-    delete node.loc
-    delete node.start
-    delete node.end
-    delete node.range
-  }
-}
-
 function useMdx(defaults) {
   const [state, setState] = useState({...defaults, file: null})
   const {run: setConfig} = useDebounceFn(
@@ -415,4 +403,16 @@ export function Editor({children}) {
       </Tabs>
     </div>
   )
+}
+
+function removePositionEsast(tree) {
+  visitEstree(tree, remove)
+  return tree
+
+  function remove(node) {
+    delete node.loc
+    delete node.start
+    delete node.end
+    delete node.range
+  }
 }
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -29,7 +29,7 @@ lowlight.registerLanguage('js', javascript)
 lowlight.registerLanguage('json', json)
 lowlight.registerLanguage('md', markdown)
 
-export function removePositionEsast(tree) {
+function removePositionEsast(tree) {
   visitEstree(tree, remove)
   return tree
 
--- a/docs/_component/editor.client.js
+++ b/docs/_component/editor.client.js
@@ -5,7 +5,7 @@ import {VFile} from 'vfile'
 import {VFileMessage} from 'vfile-message'
 import {statistics} from 'vfile-statistics'
 import {reporter} from 'vfile-reporter'
-import {evaluate} from '@mdx-js/mdx'
+import {evaluate, nodeTypes} from '@mdx-js/mdx'
 import remarkGfm from 'remark-gfm'
 import rehypeRaw from 'rehype-raw'
 import remarkFrontmatter from 'remark-frontmatter'
@@ -60,7 +60,8 @@ function useMdx(defaults) {
       remarkPlugins.push(capture('mdast'))
 
       const rehypePlugins = []
-      if (config.rehypeRaw) rehypePlugins.push(rehypeRaw)
+      if (config.rehypeRaw)
+        rehypePlugins.push([rehypeRaw, {passThrough: nodeTypes}])
       rehypePlugins.push(capture('hast'))
 
       try {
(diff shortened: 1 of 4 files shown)


slorber:
BTW @wooorm adding this md option is useful for me to better understand how the md parser works.

Currently I'm facing a little Docusaurus issue where html tags are removed (or sanitized) from the output.

For example: 

```md

## HTML

### Styling

<span style="color: blue;">blue span</span>

<p style="color: green;">green p</p>

<button style="color: red;">red button</button>

<div style="border: solid; background-color: lime; padding: 10px">
  lime <span style="color: red; margin: 10px;">red</span>
</div>

### Embeds

### Images

#### Closed image tag:

<img src="/img/docusaurus.png"/>

#### Unclosed image tag:

<img src="/img/docusaurus.png">

### Iframe

<iframe src="/"/>

### Security

```md
<p>
  When pressing this button, no alert should be printed
  <button onClick="alert('unsafe');">Click me</button>
</p>
```

<p>
  When pressing this button, no alert should be printed
  <button onClick="alert('unsafe');">Click me</button>
</p>
```

<img width="899" alt="CleanShot 2023-05-04 at 19 30 12@2x" src="https://user-images.githubusercontent.com/749374/236282852-06bf5dba-b75f-496d-9f22-b724d13b30f9.png">


https://deploy-preview-8960--docusaurus-2.netlify.app/tests/pages/markdown-tests-md

---

I tried to use some options like 

```jsx
        remarkRehypeOptions: {
          allowDangerousHtml: true,
          passThrough: ['div', 'button', 'img', 'iframe', 'html'],
        },
```

But it doesn't seem to work, see https://github.com/facebook/docusaurus/pull/8960

Any idea how to allow arbitrary html tags? So far it seems only `<span>` is in the final output, but without the style attribute.

---

Also wondering about the default behavior: considering MDX is a language and already unsafe, doesn't it make sense to allow arbitrary tags in a similar unsafe way when compiling `.md` files through the MDX package? 

I mean, it makes sense to me that we get the same safety level for `.md` and `.mdx` files considering we pass the exact same plugin options. It's also easier to explain to users that by default it's not safe in both cases. Eventually, provide opt-in for the safety of `.md` files.

docs/_component/editor.client.js
```diff
@@ -198,6 +211,46 @@ export function Editor({children}) {
...
             </label>
+            <label>
+              <input
+                checked={state.directive}
+                type="checkbox"
+                onChange={() =>
+                  setConfig({...state, directive: !state.directive})
+                }
+              />{' '}
+              Use{' '}
```
remcohaszing, line 222: Nit: For such constructs I think it’s nicer to use a single text node:

```suggestion
              />
              {' Use '}
```

docs/_component/editor.client.js
```diff
@@ -198,6 +211,46 @@ export function Editor({children}) {
...
+                type="checkbox"
+                onChange={() =>
+                  setConfig({...state, formatMd: !state.formatMd})
+                }
+              />{' '}
+              Use{' '}
+              <a href="https://mdxjs.com/packages/mdx/#optionsformat">
+                <code>format: &apos;md&apos;</code>
+              </a>
+            </label>
```
remcohaszing, line 240: Maybe we should add a tooltip here to explain MDX infers this based on the input file extension by default. I think this option makes sense for the playground, but it’s best to discourage using this option.

Actually, maybe a radio button is more appropriate here than a checkbox, although the checkbox currently fits nice in the UI.

docs/_component/editor.client.js
slorber: kind of agree yes, but I just copied the constructs above and didn't want to refactor too much things at once 🤪 

CleanShot 2023-05-05 at 17 38 58@2x

docs/_component/editor.client.js
slorber: yes a radio or select can also work, open to change a bit the UI, not sure how to organize now that it's not just a list of remark plugins.

> but it’s best to discourage using this option.

Why? Docusaurus uses MDX and I think supporting a commonmark mode can help existing docs sites to adopt it more easily, ie to have something less strict running asap without a ton of compilation errors to fix

docs/_component/editor.client.js
```diff
@@ -53,6 +57,12 @@ function useMdx(defaults) {
             recmaPlugins: [capture('esast')]
           })
         ).default
+
+        if (!config.position) {
+          removePosition(file.data.mdast, {force: true})
+          removePosition(file.data.hast, {force: true})
+          removePosition(file.data.esast, {force: true})
```
slorber, line 64: BTW not sure this works for esast, is there a tool to remove esast positions to make exploring the AST more convenient?

I mean these nodes:

```js
"start": 68,
                                            "end": 75,
                                            "loc": {
                                              "start": {
                                                "line": 4,
                                                "column": 13,
                                                "offset": 68
                                              },
                                              "end": {
                                                "line": 4,
                                                "column": 20,
                                                "offset": 75
                                              }
                                            },
```

remcohaszing: You can try [`unist-util-position-from-estree`](https://github.com/syntax-tree/unist-util-position-from-estree).

docs/_component/editor.client.js
remcohaszing: Ah right, I didn’t notice. I think consistency is more important then.

docs/_component/editor.client.js
remcohaszing: > yes a radio or select can also work, open to change a bit the UI, not sure how to organize now that it's not just a list of remark plugins.

Let’s wait for @wooorm’s feedback, as I think he designed the website. I think it looks great already. :)

> > but it’s best to discourage using this option.
> 
> Why? Docusaurus uses MDX and I think supporting a commonmark mode can help existing docs sites to adopt it more easily, ie to have something less strict running asap without a ton of compilation errors to fix

This option changes whether files are treated as commonmark or MDX. Leaving it off, means this is inferred from the file extension. It’s not about strictness (although basically anything is technically valid markdown, but MDX has syntax errors). This option is useful when the file extension is unknown (such as in this playground), but I wouldn’t recommend it otherwise. I don’t know how Docusaurus uses it exactly though.

slorber: On my side, I find it helpful to understand the MDX compiler behavior when handling .md files. 

I'm not really saying the `format` option is useful or recommended, and would be fine if instead we had the ability to customize the file name for example, as long as I can easily get the AST for an input parsed with CommonMark compat.

---

Although it's not recommended, Docusaurus historically has many sites that use MDX with a `.md` extension. One of the first unexpected things people stumble upon while upgrading to MDX 2 is that some tags (images, videos... ) do not appear anymore. See feedback here: https://github.com/facebook/docusaurus/discussions/8469#discussioncomment-5804253

For this reason, it's likely I'll add an option `format: 'mdx'`, to allow our users to upgrade their MDX version without having to rename all their files at the same time.

docs/_component/editor.client.js
wooorm: @remcohaszing that only gets a position, it doesn’t clear start/end/loc

wooorm:
> I tried to use some options like […]

`passThrough` is for node types. There is no `button` node. That value is a tag name of an `element` node instead. There are `raw` nodes. Or `mdxJsEsm` nodes.

> Any idea how to allow arbitrary html tags? So far it seems only `<span>` is in the final output, but without the style attribute.

If `span` is there, then your question is answered: they are allowed. No options needed.
However, I do not see `span`. I don’t think there are any tags.

> Also wondering about the default behavior: considering MDX is a language and already unsafe, doesn't it make sense to allow arbitrary tags in a similar unsafe way when compiling .md files through the MDX package?

Tags are allowed. In MDX. But then they are JSX. You are not using the MDX language. You are using the markdown language. Which supports raw strings of HTML inside it.

No framework (react/preact/vue/whatever) understands strings of HTML, they don’t know what to do with them, to illustrate with a common example: `"<details><summary>"`, `"</summary>"`, and `"</details>"`.

The point of MDX is to *actually* parses JSX and markdown combined.
If you turn off MDX, and still want those strings, you need an HTML parser.
And you could just as well not use MDX. So we don’t by default.

If you want to support HTML in plain markdown, are happy to include more code, and to do more work, then you can use `rehype-raw`: https://github.com/mdx-js/mdx/blob/e77be3a0028761d238bef1ed6b0e66b794f9c320/packages/mdx/test/compile.js#L733-L746.

docs/_component/editor.client.js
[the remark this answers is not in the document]
slorber: >  Don’t “connect” it to an option. But allow people to choose whether the input is plain markdown or MDX

I'm not sure to understand what you mean here 😅 

Updated the PR to this UI

CleanShot 2023-05-10 at 18 05 59@2x

docs/_component/editor.client.js
```diff
@@ -198,6 +211,46 @@ export function Editor({children}) {
...
+                onChange={() =>
+                  setConfig({...state, directive: !state.directive})
+                }
+              />{' '}
+              Use{' '}
+              <a href="https://github.com/remarkjs/remark-directive">
+                <code>remark-directive</code>
+              </a>
+            </label>
+            <hr />
```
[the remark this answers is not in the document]
slorber: hr removed

docs/_component/editor.client.js
slorber: Replaced with:

```js
import {visit as visitEstree} from 'estree-util-visit'

export function removePositionEsast(tree) {
  visitEstree(tree, remove)
  return tree

  function remove(node) {
    delete node.loc
    delete node.start
    delete node.end
    delete node.range
  }
}
```

docs/_component/editor.client.js
remcohaszing: You can pass the filename. [`compile()`](https://mdxjs.com/packages/mdx/#compilefile-options) takes a VFile compatible. In fact, I recommend always passing a filename if possible. Some plugins may depend on it. For example the JSX dev runtime depends on it.

```js
await compile({
  value: '# Hello <strong>world</strong>!',
  path: '/home/me/hello.mdx'
})
```

I would also be cautious treating `.md` files as MDX. MDX treats certain constructs differently.

I.e. given the following content:

```markdown
<div>
  foo

  bar
</div>
```

In markdown, this is equivalent to the following HTML:

```html
<div> foo bar </div>
```

In MDX, this is equivalen to the following JSX:

```jsx
<div>
  <p>foo</p>
  <p>bar</p>
</div>
```

I understand it’s useful to add the option for migration purposes in Docusaurus.

---

> it’s best to discourage using this option.

I think this may have been misinterpreted. I means it’s best to discourage using this option in general, but there are definitely legitimate use cases. The playground, where there are no actual file names, is a good use case for such an option.

slorber:
Thanks @wooorm, was able to make it work in Docusaurus with rehype-raw + passThrough of MDX nodes (some remark plugins like admonition are producing those)

I added remark-raw as a playground option too because I think it can help understand better how it all works:

CleanShot 2023-05-11 at 16 22 38


---

Note: the latest version of rehype-raw still uses hast-util-raw@7 while latest version is @8 so I didn't get your enhanced MDX error message (https://github.com/syntax-tree/hast-util-raw/commit/8e7f703e4c5622c7ef7456bd77dcde3a432bd048)

wooorm:
> so I didn't get your enhanced MDX error message

Yep, soon! Always takes time to go through the entire ecosystem ;)

docs/_component/editor.client.js
```diff
@@ -163,6 +193,41 @@ export function Editor({children}) {
...
+                  type="radio"
+                  name="language"
+                  checked={!state.formatMd}
+                  onChange={() => {
+                    setConfig({...state, formatMd: false})
+                  }}
+                />{' '}
+                Use <code>MDX</code>
+              </label>
+              <span style={{margin: '1em'}}>{' | '}</span>
```
wooorm, line 208:
```suggestion
              <span style={{margin: '1em'}}>{' '}</span>
```

hmm I think the `|` looks a bit weird, but I don’t want to spend too much time on the design now and don’t care too much.
So this is optional. I want to improve the site some more soon anyway!

package.json
```diff
@@ -89,6 +89,7 @@
     "rehype-slug": "^5.0.0",
     "rehype-stringify": "^9.0.0",
     "remark-cli": "^11.0.0",
+    "remark-directive": "^2.0.1",
```
wooorm, line 92:
```suggestion
    "remark-directive": "^2.0.0",
```

Let’s stick with the style of using clear ranges (except if we know stuff breaks)!

wooorm (commented):
Some small nits, then ready to go! Appreciate your continued work!

docs/_component/editor.client.js
```diff
@@ -25,34 +29,56 @@ lowlight.registerLanguage('js', javascript)
...
-
       if (config.gfm) remarkPlugins.push(remarkGfm)
       if (config.frontmatter) remarkPlugins.push(remarkFrontmatter)
       if (config.math) remarkPlugins.push(remarkMath)
-
+      if (config.directive) remarkPlugins.push(remarkDirective)
       remarkPlugins.push(capture('mdast'))
 
+      const rehypePlugins = []
+      if (config.rehypeRaw) rehypePlugins.push(rehypeRaw)
```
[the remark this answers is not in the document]
slorber: Added the passThrough nodeTypes, looks like it works better for format "mdx" indeed 👍

docs/_component/editor.client.js
```diff
@@ -25,34 +29,56 @@ lowlight.registerLanguage('js', javascript)
 lowlight.registerLanguage('json', json)
 lowlight.registerLanguage('md', markdown)
 
+export function removePositionEsast(tree) {
```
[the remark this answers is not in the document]
wooorm: Yep, I mean, place it in the highest scope, at the bottom!

wooorm:
Sweeet, thanks for your continued work!
Commented to clarify that Q!

docs/_component/editor.client.js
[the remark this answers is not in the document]
slorber: Done 👍 

But It was already at the highest scope, and useMdx() is at the top. So we have one helper fn above and under the default exported component 🤪

wooorm (approved):

wooorm:
Awesome, thank you!!

PR #2296, opened 2023-05-04
Add resizable editor/layout to playground

slorber:
When working with the playground and larger MDX content, I find it quite annoying to not being able to increase the size of the editors/AST explorer/preview

Also some CSS looked a bit messy/legacy and hardcoded, and there are some existing bugs like the horizontal scrollbar not appearing immediately:

CleanShot 2023-05-04 at 18 42 19@2x

----

I tried to do some cleanup, fix some bugs, and to make the height resizeable

CleanShot 2023-05-04 at 18 44 26@2x

CleanShot 2023-05-04 at 18 45 27

CleanShot 2023-05-04 at 18 46 54

CleanShot 2023-05-04 at 18 47 28

---

Not sure some of the CSS changes do not have any other unwanted impact on other pages, are they only used in the playground?

---

## Test links

Playground:
- before: https://mdxjs.com/playground/
- after: https://mdx-7kndugcie-mdx.vercel.app/playground/
- local: http://localhost:3000/playground/

### DIFF
--- a/docs/_asset/index.css
+++ b/docs/_asset/index.css
@@ -757,7 +757,9 @@ button.success {
   mask-image: paint(squircle);
   --squircle-radius: 10px;
   border-radius: 10px;
+}
 
+.frame-resizeable {
   display: flex;
   flex-direction: column;
   overflow: auto;
@@ -936,6 +938,10 @@ button.success {
   flex-grow: 1;
   margin: calc(0.125 * (1em + 1ex));
   background-color: var(--fg);
+}
+
+.playground-editor-options-tab-panel label {
+  margin-block: calc(0.5em);
 }
 
 .playground-result-badge::after {
--- a/docs/_asset/index.css
+++ b/docs/_asset/index.css
@@ -941,7 +941,7 @@ button.success {
 }
 
 .playground-editor-options-tab-panel label {
-  margin-block: calc(0.5em);
+  margin-block: 0.5em;
 }
 
 .playground-result-badge::after {
--- a/docs/_asset/index.css
+++ b/docs/_asset/index.css
@@ -301,11 +301,6 @@ pre code,
   background-color: #fafafa !important; /* Color from one-light */
 }
 
-.frame-body-box-fixed-height {
-  height: 20rem;
-  overflow-y: auto;
-}
-
 .frame-body-box {
   padding: calc(1em + 1ex);
 }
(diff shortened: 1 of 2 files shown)


docs/_asset/index.css
```diff
@@ -762,10 +757,21 @@ button.success {
   mask-image: paint(squircle);
   --squircle-radius: 10px;
   border-radius: 10px;
+
+  display: flex;
+  flex-direction: column;
+  overflow: auto;
+
+  min-height: 10rem;
+  height: 24rem;
```
slorber, line 766: these 2 can now be more easily adjusted to define the initial height of the panels

docs/_asset/index.css
```diff
@@ -789,11 +791,54 @@ button.success {
...
 .frame-body {
   border-bottom-left-radius: 3px;
   border-bottom-right-radius: 3px;
 }
 
+.react-tabs__tab-panel--selected > .frame-body {
+  flex-grow: 1;
+  display: flex;
+  flex-direction: column;
+}
```
wooorm, line 818: This selector applies to the options, making them rather big

docs/_asset/index.css
```diff
@@ -789,11 +791,54 @@ button.success {
...
+
+.react-tabs__tab-panel--selected:not(.tab-panel-scrollable) > .frame-body {
+  overflow: hidden;
+}
+
+.frame-body > div,
```
wooorm, line 824: This selector applies to the default “Box” (`padding: 1rem; background-color: violet;`) in the example, making it fill the available width. Is that intentional?

wooorm:
This also seems to affect all other code examples, such as on https://mdx-7kndugcie-mdx.vercel.app/table-of-components/. Perhaps use a new class specific to “resizable” code examples?

docs/_asset/index.css
slorber: Oh yes, this is because margin collapsing only applies to block and not flex

slorber: Fixed, also made them slightly smaller, will probably be better for introducing new options

CleanShot 2023-05-10 at 19 14 44@2x

docs/_asset/index.css
slorber: You mean the available height? 

Not intentional, fixed with better CSS scoping: `.frame-body > .playground-editor-code-mirror`

slorber:
> This also seems to affect all other code examples, such as on [mdx-7kndugcie-mdx.vercel.app/table-of-components](https://mdx-7kndugcie-mdx.vercel.app/table-of-components/). Perhaps use a new class specific to “resizable” code examples?

Should be fixed, added a class `.frame-resizeable`

slorber:
Note these reviews are a bit painful and it's easy to miss an unwanted side-effect.

On Docusaurus I started using a visual diff tool called Argos: I screenshot all sitemap pages before/after and generate a diff report from that. If you want to take a look the setup is not super complicated: https://github.com/facebook/docusaurus/pull/8947

wooorm (approved):

wooorm:
Thanks!

PR #2297, opened 2023-05-04
docs: instructions to run the docs locally

slorber:
I tried to infer the process from what I see being built on Vercel: https://vercel.com/mdx/mdx/A4DCozU1kbYNtMTc95JbAiNqt4m2

As far as I understand there's no dev mode with hot reload right?


wooorm:
Thank you!

PR #2318, opened 2023-06-28
Sort default components

remcohaszing:
This makes the output a bit more predictable.


wooorm:
- Why only `tags`?
- a direct `sort`, because it mutates *and* returns, always seems a bit finicky to me, maybe a `const tags = [...scope.tags].sort()`?

remcohaszing:
- What more than `tags`?
- It’s mutating a variable that only exists in the local scope, so I figured it’s ok. I can create a shallow copy if you still think that’s better.

wooorm:
There’s also `objects`, `components`, and `idToInvalidComponentName` in [`scope`](https://github.com/mdx-js/mdx/blob/cb427e256848dd118ed47b26d14029e7ca138c0f/packages/mdx/lib/plugin/recma-jsx-rewrite.js#L35-L39). `references` is already sorted it looks like!

> I can create a shallow copy if you still think that’s better.

Yeah I think it‘s slightly better, but it’s mostly theoretical yeah

remcohaszing:
I decided to sort the `actual` array, which is where `objects`, and `components` end up.

I also noticed some of these arrays contain unique values, so they would ideally be sets instead of arrays.

packages/mdx/lib/plugin/recma-jsx-rewrite.js
```diff
@@ -363,18 +365,22 @@ export function recmaJsxRewrite(options) {
...
                 // `const _component0 = _components['custom-element']`
                 declarations.push({
                   type: 'VariableDeclarator',
-                  id: {type: 'Identifier', name: componentName},
+                  id: {
+                    type: 'Identifier',
+                    name: /** @type {string} */ (
+                      scope.idToInvalidComponentName.get(id)
+                    )
+                  },
```
wooorm, line 383: It’s a bit sad this cast is needed, as we could use `.entries` above, though then that would make sorting a bit harder.

packages/mdx/test/compile.js
```diff
@@ -792,8 +792,8 @@ test('jsx', async () => {
       '/*@jsxRuntime automatic @jsxImportSource react*/',
       'function _createMdxContent(props) {',
       '  const _components = Object.assign({',
-      '    p: "p",',
-      '    em: "em"',
+      '    em: "em",',
+      '    p: "p"',
       '  }, props.components);',
```
wooorm, line 797: Is really only this one test case affected? I’d imagine a ton more would change?

packages/mdx/lib/plugin/recma-jsx-rewrite.js
remcohaszing: Actually that’s not too bad. I decided to change it.

packages/mdx/test/compile.js
remcohaszing: Yes. I was surprised too!

wooorm (approved):

PR #2320, opened 2023-07-04
Use node:test

remcohaszing:
This replaces `uvu` with the `node --test` command and `node:test` module.


packages/esbuild/package.json
```diff
@@ -57,7 +57,7 @@
   "scripts": {
     "prepack": "npm run build",
     "build": "tsc --build --clean && tsc --build && type-coverage",
-    "test-api": "uvu test \"\\.js$\"",
+    "test-api": "node test/index.test.js",
```
wooorm, line 60: The extra `.test` is superfluous to as all test files are already in a test folder. And it’s not how the rest of the projects look like.

packages/esbuild/test/index.test.js
```diff
@@ -140,36 +140,30 @@ test('@mdx-js/esbuild', async () => {
...
-    assert.match(
-      String(error),
-      /No loader is configured for "\.md" files/,
-      'should not handle `.md` files w/ `format: mdx`'
-    )
-  }
+    }),
+    new Error(
+      'Build failed with 1 error:\nerror: No loader is configured for ".md" files: test/esbuild.md'
+    ),
```
wooorm, line 151: I have bad experiences with providing strings. I think errors work the same. Different versions of Node throw different errors. Paths use different slashes across OSes.
Can you reuse the regexes that existed before?

packages/loader/package.json
```diff
@@ -67,7 +67,7 @@
   "scripts": {
     "prepack": "npm run build",
     "build": "tsc --build --clean && tsc --build && type-coverage",
-    "test-api": "uvu test \"\\.js$\"",
+    "test-api": "node --test",
```
wooorm, line 70: I don’t get the need for the `--test` flag?

packages/loader/test/index.test.js
```diff
@@ -80,11 +80,11 @@ webpack.mdx:1:22: Unexpected end of file in expression, expected a corresponding
     }
   })
 
-  assert.not.ok(reactBuild?.hasErrors())
+  assert.ok(!reactBuild?.hasErrors())
 
   // One for ESM loading CJS, one for webpack.
   const modReact = /** @type {{default: {default: MDXContent}}} */ (
-    // @ts-expect-error file is dynamically generated
+    // @ts-ignore file is dynamically generated
```
wooorm, line 87: Does this sometimes work now?

remcohaszing: It always sometimes worked. First the file can’t be found. If this test is reached, the file is writen, so the file is resolved. If the assertions succeed, the file is removed, and it’s an error again.

packages/esbuild/package.json
remcohaszing: I agree. I decided not to touch rename the file for now. I could though.

packages/loader/package.json
remcohaszing: `node --test` discovers and runs test files. It’s the most elegant way to run them IMO.

packages/esbuild/package.json
wooorm: Ahh okay missed that, yeah, uvu needed the file structure. I don’t mind renaming now!

packages/loader/test/index.test.js
wooorm: ah okay, perhaps we need better clean up then?

packages/loader/package.json
wooorm: `node index.js` does too? 😅

I feel like that’s much less “magic”. As it’s similar to how I run all other JS code?

remcohaszing: Sure, `node index.js` can run a single test file.

It matters as soon you have multiple test files (such as `@mdx-js/mdx`). At this point you need a tool to run all of these test files, which `node --test` provides. The alternative is to create a script that calls [`run`](https://nodejs.org/api/test.html#runoptions) from `node:test`, which is essentially what `node --test` does for you.

I like using `node --test` for everything, as it provides a single well-known entry point for running tests using the Node.js.

packages/loader/test/index.test.js
remcohaszing: Yes. As a follow-up, I would like to split these big test cases into smaller ones, like https://github.com/syntax-tree/estree-util-build-jsx/pull/7. That would make it simpler to add better cleanup functions.

packages/loader/package.json
wooorm: I also don’t really care to learn another [resolution mechanism](https://nodejs.org/api/test.html#running-tests-from-the-command-line), to learn that oh wait it doesn’t look for `.jsx` files (which might actually be a problem in this PR?)

packages/loader/test/index.test.js
wooorm: Yeah, sounds good

ChristianMurphy (commented):
Thanks @remcohaszing!
All for migrating to node:test.
Once CI is passing and @wooorm's comments are addressed this will LGTM.

packages/loader/package.json
remcohaszing: The difference is that `node --test` runs all test files in a separate process. Each test file is isolated. Also state from one file (e.g. a patched library or cwd) doesn’t leak into other test files. This is not the same as importing all test files.

Let’s say we have these files:

```js
// a.js
import { beforeEach, test } from 'node:test'

beforeEach(() => {
  console.log('beforeEach a')
})

test(() => {})
```

```js
// b.js
import { beforeEach, test } from 'node:test'

beforeEach(() => {
  console.log('beforeEach b')
})

test(() => {})
```

```js
// index.js
import './test/a.js'
import './test/b.js'
```

`node --test` runs `beforeEach` before each test in the file it’s defined in.

```sh
$ node --test   
ℹ beforeEach a
✔ <anonymous> (1.686736ms)
ℹ beforeEach b
✔ <anonymous> (1.722439ms)
ℹ tests 2
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 146.868511
```

`node --test` runs `beforeEach` before each test in all files imported.

```sh
$ node index.js 
beforeEach a
beforeEach b
beforeEach a
beforeEach b
✔ <anonymous> (1.814503ms)
✔ <anonymous> (0.353704ms)
ℹ tests 2
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 7.957022
```

Also the `npm test` script isn’t always used for running just tests. E.g. within the unified ecosystem it’s used to run testing, formatting, and linting.

I do agree it’s unfortunate `node --test` doesn’t support custom file extensions.

wooorm: Well that’s kinda my point, I don’t want stuff like that, that behaves differently when run one way or another 🥲 Again, I like just regular files, that don’t interfere, and if weird `process.cwd` hacks are needed that they clean up themselves. That *can* be run fine with Node normally.

And I don’t want to an API with a giant surface. I want to throw errors if stuff is wrong. And in 4 years, when a new test framework is around, we can switch again easily.

Feels very similar to whether to write in `.ts` or `.js` files. I am not against types or tests, far from it. But it’s so much easier to maintain things for years if files are just plain javascript files.

> within the unified ecosystem it’s used to run testing, formatting, and linting.

`test-api` and `test-coverage` do that, then.

You are sometimes going to need to pass other things to Node, with or without `--test`. So `node --test` doesn’t always run all the expected tests in the expected way. It sometimes needs to be configured. And that’s what npm scripts solve.

ChristianMurphy: > The difference is that node --test runs all test files in a separate process. Each test file is isolated. Also state from one file (e.g. a patched library or cwd) doesn’t leak into other test files. This is not the same as importing all test files.

On projects outside unified, I've tended to use the BDD API to handle this ([`describe`](https://nodejs.org/api/test.html#describename-options-fn) blocks with [`it`](https://nodejs.org/api/test.html#itname-options-fn) tests)
Which allows scoping `before`, `after`, `beforeEach`, and `afterEach` to a single test suite.

remcohaszing: > Well that’s kinda my point, I don’t want stuff like that, that behaves differently when run one way or another :smiling_face_with_tear: Again, I like just regular files, that don’t interfere, and if weird `process.cwd` hacks are needed that they clean up themselves. That _can_ be run fine with Node normally.

This is exactly what I want. I want test files that don’t interfere with each other, so tests can be run in isolation. I.e. when running `npm test` and something goes wrong in `test/compile.js`, I want to be able to run `node test/compile.js` to troubleshoot the issue. This isn’t possible if it’s run via an import, because other tests might interfere.

I also like using `describe` to organize somewhat bigger test suites, but that’s not the same as running each test file in a separate process via `node --test`.

I will change it, because I’d rather move from `uvu` to `node:test` than block it over this, but I really think we should leverage `node --test`.

wooorm (approved):

wooorm:
TLDR didn’t read everything but LGTM ;)

PR #2327, opened 2023-07-14
Use subtests for esbuild, mdx, and remark-mdx

remcohaszing:
These packages have significantly sized test cases. Using subtests helps when troubleshooting these tests.

### DIFF
--- a/packages/esbuild/test/index.js
+++ b/packages/esbuild/test/index.js
@@ -102,6 +102,7 @@ test('@mdx-js/esbuild', async (t) => {
       plugins: [esbuildMdx()]
     })
 
+    /** @type {MDXContent} */
     const Content =
       /* @ts-expect-error file is dynamically generated */
       (await import('./esbuild-md.js')).default // type-coverage:ignore-line
@@ -128,6 +129,7 @@ test('@mdx-js/esbuild', async (t) => {
       plugins: [esbuildMdx({mdExtensions: [], mdxExtensions: ['.md']})]
     })
 
+    /** @type {MDXContent} */
     const Content =
       /* @ts-expect-error file is dynamically generated */
       (await import('./esbuild-md-as-mdx.js')).default // type-coverage:ignore-line
@@ -471,6 +473,7 @@ test('@mdx-js/esbuild', async (t) => {
       bundle: true
     })
 
+    /** @type {MDXContent} */
     const Content =
       /** @ts-expect-error file is dynamically generated */
       (await import('./esbuild-compile-from-memory.js')).default // type-coverage:ignore-line
@@ -499,6 +502,7 @@ test('@mdx-js/esbuild', async (t) => {
         bundle: true
       })
 
+      /** @type {MDXContent} */
       const Content =
         /** @ts-expect-error file is dynamically generated */
         (await import('./esbuild-compile-from-memory-empty.js')).default // type-coverage:ignore-line
@@ -531,6 +535,7 @@ test('@mdx-js/esbuild', async (t) => {
       plugins: [esbuildMdx({allowDangerousRemoteMdx: true})]
     })
 
+    /** @type {MDXContent} */
     const Content =
       /* @ts-expect-error file is dynamically generated */
       (await import('./esbuild-with-remote-md.js')).default // type-coverage:ignore-line
@@ -568,6 +573,7 @@ test('@mdx-js/esbuild', async (t) => {
       plugins: [esbuildMdx({allowDangerousRemoteMdx: true})]
     })
 
+    /** @type {MDXContent} */
     const Content =
       /* @ts-expect-error file is dynamically generated */
       (await import('./esbuild-with-remote-mdx.js')).default // type-coverage:ignore-line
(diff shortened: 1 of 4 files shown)


packages/esbuild/test/index.js
```diff
@@ -16,177 +16,189 @@ import React from 'react'
...
+
+    /** @type {MDXContent} */
+    const Content =
+      /* @ts-expect-error file is dynamically generated */
+      (await import('./esbuild.js')).default // type-coverage:ignore-line
+
+    assert.equal(
+      renderToStaticMarkup(React.createElement(Content)),
+      '<h1>Hello, World!</h1>',
+      'should compile'
```
wooorm, line 43: a) I thought you didn’t want assertion messages?
b) to not have messages there, I have been moving those to `x` in `test(x, ...)`. So ``'should compile `.md`'`` instead of `'MD'`

wooorm (commented):
Looks like a couple cases earlier on have a capitalized, different message, while the assertion message is still in place. I think you can move that message to the test case (so that it prints like “should ...”). Then it’s TLDR but fine!

ChristianMurphy (approved):
Thanks @remcohaszing!
It might be even nicer to use `describe` and `it("should ...")` for suites and tests, but that is nice to have.
This in and of itself is a good improvement. 👍

wooorm:
I’m personally not much of a fan of describe and it. AFAIK they don’t add anything really, and using `test` only seems just simpler?

packages/esbuild/test/index.js
remcohaszing: The esbuild tests were already somewhat categorized using comments. I used those comments as the test names. I changed it to use the previous assertion messages now, as that’s consistent with the other tests indeed.

remcohaszing:
I like using `describe` blocks, but chose subtests because @wooorm has expressed he prefers those in another issue somewhere. It’s mostly a stylistic difference.

ChristianMurphy:
> AFAIK they don’t add anything really

They add clarity, by being closer to natural language.
`describe` clarifies this is what we are testing.
`it` naturally flows into test description beginning with `"should ..."`.

> and using `test` only seems just simpler?

some `test` actually test things, other `test` don't test anything and act only as a label, that feels more confusing to me.

Again, this is not a blocker.

wooorm:
Thanks Remco! Appreciate the work! <3

wooorm:
@ChristianMurphy

> They add clarity, by being closer to natural language.

I don’t really buy into this argument; it sounds nice, but if that’s the case, why don‘t we do that in other places as much as tests?

> some test actually test things, other test don't test anything and act only as a label, that feels more confusing to me.

It used to be that there was a clearer divide between test as the whole wrapper, and assert for the actual testing things. Remco pointed out that he prefers wrapping assertions into separate test cases so that tests keep running if one assertion fails, and that the output is nicer (you see a bunch of green stuff for what’s good, red stuff for what isn’t).
These are differences where `tape` was different from `node:test`. Although I find it noisy to have so many `test` cases, I think they’re fine trade-offs, dropping a dev-dep and switching to the built in `node:test` is nicer.

I feel like that’s more *another* layer of `test`s grouping things; assertions still do the only “actual testing”: all `test`s are just groupings and labels.

ChristianMurphy:
> why don‘t we do that in other places as much as tests?

That's a good question, why not?
I don't really buy into "we have unclear naming already, so let's keep doing it"

> These are differences where `tape` was different from `node:test`. Although I find it noisy to have so many test cases, I think they’re fine trade-offs, dropping a dev-dep and switching to the built in `node:test` is nicer.

agreed

> I feel like that’s more another layer of `test`s grouping things; assertions still do the only “actual testing”: all `test`s are just groupings and labels.

I agree there is another layer of grouping.
I feel like spec style testing (`describe` and `it`) fits better with `unified`'s culture of coding to spec.
And that the main reason it isn't already in place, is that `tape` didn't support this convention for testing.

wooorm:
> That's a good question, why not?

My personal answer would be: because programming is not natural language. Trying to merge them doesn’t work well. But that’s from my experience.

> And that the main reason it isn't already in place, is that tape didn't support this convention for testing.

I used to use it https://github.com/retextjs/retext/blob/1.0.0/test.js, jasmine/mocha, I personally preferred `t` style over describe/it-style!

ChristianMurphy:
> My personal answer would be: because programming is not natural language. Trying to merge them doesn’t work well. But that’s from my experience.

I'm not saying they are precisely the same, nor that we should merge them.
We do aim to name things sensibly in English, with names that follow convention so they are easily understood.
Spec driven development naming is one such convention.

> This is interesting! Why?

It by naming convention and design convention focuses on, the why and what of how a piece of code should work.
Over a standard test driven approach, which aims to hit 100% coverage, without much consideration to why specific features or edge cases are needed.

wooorm:
> We do aim to name things sensibly in English, with names that follow convention so they are easily understood.

Important distinction is of course, sure, we *use* natural language, but trying to make a running English sentence with code is something else.

To me, `test('should parse a self-closing tag', …` reads as clearly as `it('should parse a self-closing tag', …`. But having one name for grouping things, `test`, seems easier.

PR #2358, opened 2023-09-29
Fix site

wooorm:
* new very fast, more spacious, playground
* proper syntax highlighting for MDX
* static (except for playground), no RSC, super fast, can be hosted everywhere


wooorm:
i ain't reading all that; i'm happy for u tho; or sorry that happened

PR #2375, opened 2023-10-19
Replace Object.assign with spread syntax in docs

remcohaszing:
`estree-util-build-jsx` was updated to version 3, which uses spread syntax. However, some documentation was still using `Object.assign` in output examples.


wooorm:
I’m still about to go through all the docs, and check all input/output examples.

Do you think you have all? I can merge but I’ll go through everything anyway!

remcohaszing:
Ah ok, I thought maybe you missed this. These were all occurrences of `Object.assign`.

ChristianMurphy (approved):

PR #2377, opened 2023-10-20
Add GitHub pages

remcohaszing:
This requires additional changes in the repository GitHub Pages settings. _Source_ needs to be set to _GitHub Actions_, and a custom domain needs to be configured.


.github/workflows/website.yml
```diff
@@ -0,0 +1,25 @@
+name: website
+on:
+  push:
+    branches: [main]
```
wooorm, line 4:
```suggestion
    branches:
      - main
```

.github/workflows/website.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    environment:
+      name: Website
+      url: ${{ steps.deployment.outputs.page_url }}
```
wooorm, line 14:
```suggestion
      url: ${{steps.deployment.outputs.page_url}}
```

.github/workflows/website.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with: {node-version: 18}
```
wooorm, line 18:
```suggestion
        with:
          node-version: 18
```

wooorm: Why 18? `node` is the always stable one. Perhaps that?

wooorm (commented):
One Q rest stylistic

.github/workflows/website.yml
remcohaszing: I copy-pasted this and forgot to update.

wooorm (approved):

wooorm:
>  a custom domain needs to be configured.

Do you mean a `CNAME` file in `public/`?

remcohaszing:
No, when using GitHub actions as a source, the custom domain must be configured under the pages settings.

wooorm:
I think that’s the same? when you change the custom domain in the interface, it writes to a CNAME file? Or at least it always used to be that way?

remcohaszing:
I found it in the [docs](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site#configuring-a-subdomain) now:

> Under "Custom domain", type your custom domain, then click **Save**. If you are publishing your site from a branch, this will create a commit that adds a `CNAME` file directly to the root of your source branch. If you are publishing your site with a custom GitHub Actions workflow, no `CNAME` file is created. For more information about your publishing source, see "[Configuring a publishing source for your GitHub Pages site](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)."

PR #2392, opened 2023-11-04
Fix reset of playground render errors

ManasMakde:
The error boundary does not automatically reset on changing the text value, To fix it the [`value`](https://github.com/mdx-js/mdx/blob/efff74e48aec9286c62185bf04f7b8aa72e07bd6/docs/_asset/editor.jsx#L149C45-L149C45) state was added to the `resetKeys` in `ErrorBoundary`

You can see the difference on the [official site](https://mdxjs.com/playground/) vs localhost:

https://github.com/mdx-js/mdx/assets/73928607/ce24eb47-b9f9-41be-8882-a5d1c54ad9be

### DIFF
--- a/docs/_asset/editor.jsx
+++ b/docs/_asset/editor.jsx
@@ -210,7 +210,7 @@ function Playground() {
           })
 
           return (
-            <ErrorBoundary FallbackComponent={ErrorFallback}>
+            <ErrorBoundary FallbackComponent={ErrorFallback} resetKeys={[value]}>
               <div className="playground-result">{mod.default({})}</div>
             </ErrorBoundary>
           )


wooorm:
a) it does seem to work for me? 🤔
b) there’s a button under the error you could click if not?

ManasMakde:
a) I don't understand what you mean?

b) The button works but it feels too tedious & disturbs the typing flow and it would be quite nice feature given the change is only 1 line :)

c) I didn't find it either but I found [this solution on stackoverflow](https://stackoverflow.com/a/71884022)

wooorm:
> a) I don't understand what you mean?

You have a bug. I don’t have that bug

> c) I didn't find it either but I found [this solution on stackoverflow](https://stackoverflow.com/a/71884022)

That seems to be about `key`? Which is a React thing, unrelated to this package, although it indeed resets when it changes

---

Anyway, if this didn’t work for you and does not, sure, we can try it!

wooorm (approved):

wooorm:
Thank you!

PR #2398, opened 2023-11-10
Refactor to handle uncaught webpack errors

remcohaszing:
A dynamic import may fail for various reasons. This was unhandled. This means the loader could wait indefinitely. Now it defers the error to Webpack.

I didn’t encounter this, but it the code just caught my eye. I don’t know how to test it. It doesn’t affect test coverage.

### DIFF
--- a/packages/loader/index.cjs
+++ b/packages/loader/index.cjs
@@ -28,5 +28,5 @@ function loader(code) {
   // Note that `import()` caches, so this should be fast enough.
   import('./lib/index.js').then((module) => {
     return module.loader.call(this, code, callback)
-  })
+  }, callback)
 }


wooorm:
I don’t think `import('./lib/index.js')` could ever fail? We’re importing our ESM wrapper, that shouldn‘t ever throw at the top-level (except when developing maybe, but then TS/xo should likely catch it?)

remcohaszing:
This fails if any dependency breaks while importing. This should never occur, but people make mistakes. A dependency could be missing. People could mess with their `node_modules`. In this case I expect a crash, not an infinite loading state.

wooorm:
right but the issue sounds like this affects users — but it just affects us while maintaining this, right?

remcohaszing:
This affects users too. Any dependency can technically break at any time. And users can also mess with their own `node_modules`.

wooorm:
I mean, I don’t think people manually messing up their node_modules... Then they can break anything. They can break every package that way.

wooorm (approved):

wooorm:
Thanks!

wooorm:
Released in [3.0.1](https://github.com/mdx-js/mdx/releases/tag/3.0.1)! Thanks!

PR #2409, opened 2023-11-29
Fix `mdx-embed` link in docs

torn4dom4n:
The old URL is broken so I updated a new URL, see https://github.com/PaulieScanlon/mdx-embed/pull/253.


wooorm (approved):

wooorm:
Thank you!

PR #2438, opened 2024-02-08
Fix generated JSX pragmas for new babel

remcohaszing:
### Initial checklist

### Description of changes

Each JSX pragma needs to be on its own line for tools like Babel or TypeScript to understand them.

There is a broken pragma in the @mdx-js/preact tests, which was left untouched. Due to a misconfiguration, to fix it, would break type checking.

### DIFF
--- a/docs/_asset/editor.jsx
+++ b/docs/_asset/editor.jsx
@@ -1,4 +1,4 @@
-/**
+/*
  * @jsxRuntime automatic
  * @jsxImportSource react
  */
--- a/docs/_asset/editor.jsx
+++ b/docs/_asset/editor.jsx
@@ -1,5 +1,7 @@
-/* @jsxRuntime automatic
-@jsxImportSource react */
+/**
+ * @jsxRuntime automatic
+ * @jsxImportSource react
+ */
 
 /**
  * @typedef {import('@wooorm/starry-night').Grammar} Grammar
--- a/docs/_asset/editor.jsx
+++ b/docs/_asset/editor.jsx
@@ -1,4 +1,5 @@
-/* @jsxRuntime automatic @jsxImportSource react */
+/* @jsxRuntime automatic
+@jsxImportSource react */
 
 /**
  * @typedef {import('@wooorm/starry-night').Grammar} Grammar
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -41,7 +41,8 @@ That’s *roughly* turned into the following JavaScript.
 The below might help to form a mental model:
 
 ```tsx path="output-outline.jsx"
-/* @jsxRuntime automatic @jsxImportSource react */
+/* @jsxRuntime automatic
+@jsxImportSource react */
 
 export function Thing() {
   return <>World</>
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -41,8 +41,10 @@ That’s *roughly* turned into the following JavaScript.
 The below might help to form a mental model:
 
 ```tsx path="output-outline.jsx"
-/* @jsxRuntime automatic
-@jsxImportSource react */
+/*
+@jsxRuntime automatic
+@jsxImportSource react
+*/
 
 export function Thing() {
   return <>World</>
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -41,8 +41,8 @@ That’s *roughly* turned into the following JavaScript.
 The below might help to form a mental model:
 
 ```tsx path="output-outline.jsx"
-/*@jsxRuntime automatic*/
-/*@jsxImportSource react*/
+/* @jsxRuntime automatic */
+/* @jsxImportSource react */
 
 export function Thing() {
   return <>World</>
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -41,10 +41,8 @@ That’s *roughly* turned into the following JavaScript.
 The below might help to form a mental model:
 
 ```tsx path="output-outline.jsx"
-/*
-@jsxRuntime automatic
-@jsxImportSource react
-*/
+/*@jsxRuntime automatic*/
+/*@jsxImportSource react*/
 
 export function Thing() {
   return <>World</>
--- a/packages/mdx/lib/plugin/recma-document.js
+++ b/packages/mdx/lib/plugin/recma-document.js
@@ -105,7 +105,7 @@ export function recmaDocument(options) {
     if (pragmas.length > 0) {
       tree.comments.unshift({
         type: 'Block',
-        value: pragmas.join('\n'),
+        value: '\n' + pragmas.join('\n') + '\n',
         data: {_mdxIsPragmaComment: true}
       })
     }
--- a/packages/mdx/lib/plugin/recma-document.js
+++ b/packages/mdx/lib/plugin/recma-document.js
@@ -73,8 +73,6 @@ export function recmaDocument(options) {
     const exportedIdentifiers = []
     /** @type {Array<Directive | ModuleDeclaration | Statement>} */
     const replacement = []
-    /** @type {Array<string>} */
-    const pragmas = []
     let exportAllCount = 0
     /** @type {ExportDefaultDeclaration | ExportSpecifier | undefined} */
     let layout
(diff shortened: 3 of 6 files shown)


wooorm:
It looks like Babel tightened this after this was written (in xdm): https://github.com/babel/babel/pull/14012/files vs https://github.com/wooorm/xdm/blame/4713bd50b87dc07cb0e654553e54b599342c3292/lib/plugin/recma-document.js#L68
SWC used to be strict, but loosened it (due to our comments and babel at the time): https://github.com/swc-project/swc/pull/2377

remcohaszing:
I first encountered this in MDX analyzer. The way it’s used there is unrelated to usage here, but that’s where I noticed. Someone from the Volar team mentioned the pragmas need to be on separate lines.

I verified this Babel playground only picked up on these pragmas when they are on separate lines. I also verified this in a [TypeScript playground](https://www.typescriptlang.org/play?#code/PQKgBAAgVgzgHgJQK4DsAuBLAtgUzAQyTQHst9MBjSWOASSwAdiAnNAZWKWYrywxQwBaGmBDAAUOOxNWogjDAIc+CmjAAzZqTAByZstU7JFYihhqcAGxy50YALxgAPABMMANzDAAfEA) (look at the output changes if you add a newline between those pragmas).

I also noticed the Preact tests in the infer JSX types based on the React types, not the Preact types. To fix this requires a bit more work that’s out of scope of this PR, but I would like to fix it in a follow-up PR.

wooorm:
TypeScript has also changed: https://github.com/babel/babel/issues/13896#issuecomment-956718381.

If the Preact problem requires a different fix, how do you know it relates to this? AFAIK TS always has had problems with multiple JSX runtimes at the same time.

I’m fine with using the new format that Babel and TypeScript introduced.
But I’d prefer it to be pretty:

```js
/**
 * @jsxWhatever
 * @jsxStuff asdasd
 */
```

So like line endings and asterisks and aligned

remcohaszing:
Although two asterisks at the beginning works, I think one is better. We don’t want this to be accidentally picked up as JSDoc somehow.

Apparently `astring` dedents block comments. So the comment becomes this:

```js
/*
* @jsxWhatever
* @jsxStuff asdasd
*/
```

IMO this is slightly prettier:

```js
/*
@jsxWhatever
@jsxStuff asdasd
*/
```

remcohaszing:
> If the Preact problem requires a different fix, how do you know it relates to this? AFAIK TS always has had problems with multiple JSX runtimes at the same time.

We should leverage TypeScript references to support mono repos and separate projects. There should be only one JSX runtime per project. This is unrelated to the MDX output, which is handled by this PR.

wooorm:
I dunno, feels weird to have to do that? Why “should” there be one runtime per project? I would think it is pretty reasonable for folks to have different runtimes in one project (with comments in files), and that this is a bug in TS?

wooorm:
> We don’t want this to be accidentally picked up as JSDoc somehow

Why not? Them starting at a line and using `@` makes them look a lot like JSDoc. Which I infer to be because they’re *intended* to look like JSDoc?

> IMO this is slightly prettier:

Also fine by me

wooorm:
I quite like the separate comments but it’s a bit extra work for the `_mdxIsPragmaComment` stuff aka this handling: https://github.com/mdx-js/mdx/commit/f48d038b9667a644b1dc508a7ffd894b2b673ea7.

What do you think, go with separate comments?

remcohaszing:
`string` strips leading / trailing whitespace 1-liner block comments too, but I still like it slightly better.

docs/_asset/editor.jsx
```diff
@@ -1,4 +1,7 @@
-/* @jsxRuntime automatic @jsxImportSource react */
+/**
```
wooorm, line 1:
```suggestion
/*
```

docs/docs/using-mdx.mdx
```diff
@@ -41,7 +41,8 @@ That’s *roughly* turned into the following JavaScript.
 The below might help to form a mental model:
 
 ```tsx path="output-outline.jsx"
-/* @jsxRuntime automatic @jsxImportSource react */
+/*@jsxRuntime automatic*/
+/*@jsxImportSource react*/
```
wooorm, line 45:
```suggestion
/* @jsxRuntime automatic */
/* @jsxImportSource react */
```

This one is not exact output, but a rough idea on how it looks, so can be pretty

packages/mdx/lib/plugin/recma-document.js
```diff
@@ -102,10 +102,10 @@ export function recmaDocument(options) {
...
 
-    if (pragmas.length > 0) {
+    for (let index = pragmas.length; index--; index >= 0) {
       tree.comments.unshift({
         type: 'Block',
-        value: pragmas.join(' '),
+        value: pragmas[index],
         data: {_mdxIsPragmaComment: true}
       })
     }
```
wooorm, line 111: Nitty, but I don‘t really like the reverse stuff.
This works locally (but I can’t suggest it on GH as it overlaps):

```js
    tree.comments.unshift(
      ...pragmas.map(function (d) {
        return /** @type {const} */ ({
          type: 'Block',
          value: d,
          data: {_mdxIsPragmaComment: true}
        })
      })
    )
```

Only don’t really like the `const` but TS needs it apparently

packages/mdx/lib/plugin/recma-jsx-build.js
```diff
@@ -45,13 +45,15 @@ export function recmaJsxBuild(options) {
...
-      tree.comments.shift()
+    if (tree.comments) {
+      let index = 0
+      while (index < tree.comments.length) {
+        if (tree.comments[index].data?._mdxIsPragmaComment) {
+          tree.comments.splice(index, 1)
+        } else {
+          index++
+        }
+      }
```
wooorm, line 56:
```suggestion
    if (tree.comments) {
      tree.comments = tree.comments.filter(function (d) {
        return !d.data || !d.data._mdxIsPragmaComment
      })
```

Simpler?

remcohaszing: This creates a copy instead of modifying the original array. Fine with me, but I want to make sure you’re aware.

Also even simpler:

```suggestion
    if (tree.comments) {
      tree.comments = tree.comments.filter(function (d) {
        return !d.data?._mdxIsPragmaComment
      })
```

Or even:

```suggestion
    if (tree.comments) {
      tree.comments = tree.comments.filter((d) => !d.data?._mdxIsPragmaComment)
```

wooorm: I prefer real functions instead of arrows. `?.` is fine by me.
Both real functions and optional are in the rest of the code base.

New array is fine. This isn‘t a slow path. I’m more worried about iterating and splicing combined, easy to get wrong and hard to get fast in my micromark experience

packages/mdx/lib/plugin/recma-document.js
wooorm: Ahh. I dunno maybe a lot of repetition. Maybe more comments will be added in the future too.
I don‘t mind my const too much? How do you feel about that?

remcohaszing: I already applied this using a function to avoid repetition. I like it, but I can revert if you don’t.

wooorm:
Released in [3.0.1](https://github.com/mdx-js/mdx/releases/tag/3.0.1)! Thanks!

wooorm:
This broke the website. Looks like esbuild doesn’t pick it up

wooorm:
Both

```
/* @jsxRuntime automatic */
/* @jsxImportSource react */
```

And:

```
/* @jsxRuntime automatic @jsxImportSource react */
```

Are fine. So the release is not broken.

docs/_asset/editor.jsx
```diff
@@ -1,4 +1,7 @@
-/* @jsxRuntime automatic @jsxImportSource react */
```
Brunojoes69, line 1: -/*@.jsRuntimex automatic @jsImportSourcex react */

packages/mdx/lib/plugin/recma-document.js
```diff
@@ -73,8 +73,6 @@ export function recmaDocument(options) {
     const exportedIdentifiers = []
     /** @type {Array<Directive | ModuleDeclaration | Statement>} */
     const replacement = []
-    /** @type {Array<string>} */
-    const pragmas = []
```
Brunojoes69, line 77: - const paragrams = ()

PR #2439, opened 2024-02-12
Add `rel=sponsored` to sponsor links

ChristianMurphy:
### Initial checklist

### Description of changes

Matching annotations to the ones used on https://unifiedjs.com/


remcohaszing (approved):
Since this is MDX, and not some markdown rendered on GitHub, perhaps we could leverage a React component. We could even avoid using a table altogether.

I approved, because I’m fine with this change regardless.

wooorm:
This is copy/pasted from the readmes when everything is changed. That’s why it’s nice to have it similar to the rest, minimal changes.

wooorm:
Thanks!

PR #2440, opened 2024-02-13
Change webpack peer dependency to optional in loader

chenjiahan:
### Initial checklist

### Description of changes

[Rspack](https://github.com/web-infra-dev/rspack) is a fast Rust-based web bundler and it is compatible with the architecture and ecosystem of webpack.

The `@mdx-js/loader` can be used with Rspack, but users will get a peer dependency warning, adding peerDependenciesMeta can resolve it. 

The same as: https://github.com/jantimon/html-webpack-plugin/pull/1829

### DIFF
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -47,7 +47,16 @@
     "source-map": "^0.7.0"
   },
   "peerDependencies": {
+    "@rspack/core": "0.x || 1.x",
     "webpack": ">=5"
+  },
+  "peerDependenciesMeta": {
+    "@rspack/core": {
+      "optional": true
+    },
+    "webpack": {
+      "optional": true
+    }
   },
   "devDependencies": {},
   "scripts": {
--- a/packages/loader/package.json
+++ b/packages/loader/package.json
@@ -47,13 +47,9 @@
     "source-map": "^0.7.0"
   },
   "peerDependencies": {
-    "@rspack/core": "0.x || 1.x",
     "webpack": ">=5"
   },
   "peerDependenciesMeta": {
-    "@rspack/core": {
-      "optional": true
-    },
     "webpack": {
       "optional": true
     }
--- a/packages/loader/readme.md
+++ b/packages/loader/readme.md
@@ -7,7 +7,7 @@
 [![Backers][backers-badge]][collective]
 [![Chat][chat-badge]][chat]
 
-webpack / Rspack loader for MDX.
+webpack loader for MDX.
 
 <!-- more -->
 
@@ -34,7 +34,8 @@ This package is a webpack loader to support MDX.
 
 ## When should I use this?
 
-This integration is useful if you’re using webpack or Rspack (or another tool that uses webpack, such as Next.js).
+This integration is useful if you’re using webpack (or another tool that uses
+webpack, such as Next.js or Rspack).
 
 This integration can be combined with the Babel loader to compile modern
 JavaScript features to ones your users support.
--- a/packages/loader/readme.md
+++ b/packages/loader/readme.md
@@ -7,7 +7,7 @@
 [![Backers][backers-badge]][collective]
 [![Chat][chat-badge]][chat]
 
-webpack loader for MDX.
+webpack / Rspack loader for MDX.
 
 <!-- more -->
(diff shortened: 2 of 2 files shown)


packages/loader/package.json
```diff
@@ -47,8 +47,17 @@
     "source-map": "^0.7.0"
   },
   "peerDependencies": {
+    "@rspack/core": "0.x || 1.x",
```
remcohaszing, line 50: We don’t need a peer dependency for every framework that’s compatible with Webpack.

```suggestion
```

packages/loader/package.json
```diff
@@ -47,8 +47,17 @@
...
     "webpack": ">=5"
   },
+  "peerDependenciesMeta": {
+    "@rspack/core": {
+      "optional": true
+    },
```
remcohaszing, line 56: This suggestion is related to the comment above.

```suggestion
```

packages/loader/readme.md
```diff
@@ -7,7 +7,7 @@
 [![Backers][backers-badge]][collective]
 [![Chat][chat-badge]][chat]
 
-webpack loader for MDX.
+webpack / Rspack loader for MDX.
```
remcohaszing, line 10:
```suggestion
webpack loader for MDX.
```

packages/loader/readme.md
```diff
@@ -34,8 +34,7 @@ This package is a webpack loader to support MDX.
 
 ## When should I use this?
 
-This integration is useful if you’re using webpack (or another tool that uses
-webpack, such as Next.js).
+This integration is useful if you’re using webpack or Rspack (or another tool that uses webpack, such as Next.js).
```
remcohaszing, line 37:
```suggestion
This integration is useful if you’re using webpack (or another tool that uses
webpack, such as Next.js or Rspack).
```

remcohaszing (changes requested):
I think all modern package managers install peer dependencies. They don’t install optional peer dependencies. The ones that don’t install peer dependencies merely show a warning, which is annoying, but not an error.

The `webpack` peer dependency exists for 2 reasons:

1. To indicate this is only compatible with Webpack versions 5 or greater.
2. For the type definitions, which users may or may not use. This is really similar to the `@types/react` peer dependency, which was discussed in #2281. We ended up not using optional peer dependencies there.

wooorm:
typescript users that use rspack should likely still have webpack for types to work?
How likely is it that rspack users also use typescript?

If rspack supports webpack loaders, we don’t have to change the docs from webpack loader to webpack or rspack loader. Rspack users know that they can use webpack loaders.

chenjiahan:
Thank you for all the review comments, I have updated the PR content, just treating webpack as an optional peer dependency.
Some Rspack users will use TypeScript, but the types of webpack are not necessary in most scenarios, so it might be more appropriate to not install webpack.

wooorm:
> Some Rspack users will use TypeScript, but the types of webpack are not necessary in most scenarios, so it might be more appropriate to not install webpack.

How do you configure loaders such as this in rspack then?
It is IMO quite useful to have type safety for the options of this loader. Those types are in this package. And they need `webpack`

wooorm:
You can use TS outside of `.ts` files! With `@type` jsdoc. That’s what is done in the Use example here: https://github.com/mdx-js/mdx/tree/main/packages/loader#use. We recommend it to prevent mistakes.

chenjiahan:
This is a great method 👍🏻

As far as I know, `import('@mdx-js/loader').Options` does not depend on the types of webpack, webpack only provides the `LoaderContext` type, which will not affect `Options` type.

remcohaszing:
> As far as I know, `import('@mdx-js/loader').Options` does not depend on the types of webpack, webpack only provides the `LoaderContext` type, which will not affect `Options` type.

This is true. However, the package also exports the type of the loader, which does depend on the webpack types.

I’m indifferent on whether or not the webpack peer dependency should be optional.

I do welcome the documentation change.

remcohaszing (approved):

wooorm:
Thank you :)

chenjiahan:
Thank you for your time ❤️

PR #2443, opened 2024-02-22
add use client directive to react package

souporserious:
### Initial checklist

*   [ ] If applicable, I’ve added docs and tests

### Description of changes

Frameworks like the Next.js MDX plugin default to `@mdx-js/react` if no `mdx-components` file is present. When this happens, an error is thrown due to `useContext` in `useMDXComponents` and it's not apparent what's going on. This fixes this default use case in Next.js when overriding MDX components is not needed.

### DIFF
--- a/packages/react/lib/index.js
+++ b/packages/react/lib/index.js
@@ -1,3 +1,4 @@
+'use client'
 /**
  * @typedef {import('mdx/types.js').MDXComponents} MDXComponents
  * @typedef {import('react').Component<{}, {}, unknown>} Component


ChristianMurphy:
Welcome @souporserious! 👋
Thanks for opening a PR and starting a discussion.

> This fixes this default use case in Next.js when overriding MDX components is not needed.

I'm not opposed to this. Though I am a bit cautious, directives are still a canary feature of React (as of 2024-02-15) and subject to change ahead of the React 19 release.
https://react.dev/blog/2024/02/15/react-labs-what-we-have-been-working-on-february-2024#new-features-in-react-canary
I want to avoid `mdx` having to release multiple major versions trying to offer a stable support (in MDX) for an unstable API (in React and Next).
**If** we add this, I feel like there may need to be some notes about the experimental nature and how this is not covered by SemVer in `mdx`

> When this happens, an error is thrown due to useContext in useMDXComponents and it's not apparent what's going on

I do see some notes about context in the canary docs https://react.dev/reference/react/use-client#using-third-party-libraries
It would be nice if react itself provided better error/warning/info messages to offer guidance, since tracking features which are and are not supported server side will be a react ecosystem wide issue, not just an MDX specific one.

It is also worth noting, while using `context` is supported, it is generally not recommended.
Passing a `components` prop directly is the preferred method of adding custom components.

With that in mind, I wonder if it would be worth having a context-free server-side render-able version as well as a context-full client-only version. 🤔 

/cc @mdx-js/core thoughts?

remcohaszing:
I think this is a good change for RSC in general, not necessarily for Next.js. They are better off accepting https://github.com/vercel/next.js/pull/59693 and aplying the additional suggestion there.

Did you test this change to make sure it works?

souporserious:
@wooorm this is the error in Next.js that is thrown without adding an `mdx-components` file since it defaults to `@mdx-js/react`:

image

This PR was an attempt to fix that, but after making an [example here](https://github.com/souporserious/mdx-use-client) it looks like there's this additional Next.js error so it doesn't seem adding `use client` is a proper fix:

image

> Is our react project really a client file/component?

Good point, probably best to let the author choose where the boundary is.

---

@remcohaszing yeah, this does seem to be more of a Next.js related issue so I'll close this PR in favor the PR you linked. Providing a default `mdx-components` implementation would be more fitting to address the issue here. It seems `@mdx-js/react` shouldn't be loaded at all since it doesn't seem to ever work.

Issue #1655, opened 2021-09-15
mdx-js/runtime: Avoid unstable element type to improve render performance


marcustyphoon:
Heya! We're using the runtime package to live-render a preview window in netlify-cms-app. Naturally, the performance of this use case is never going to be amazing, but I believe I found a way to improve it noticeably.

### Subject of the feature

mdx-js/runtime currently creates and calls a function that defines and returns an `<MDXContent>` component containing the desired content. I believe that, because this is a function that is defined on every render, its type is different on every render, so React's reconciler will always tear down the DOM component tree rather than diffing the DOM elements, if mdx-js/runtime is used live.

https://github.com/mdx-js/mdx/blob/952f6dc5244013346ecf028a4d6d8973053de38f/packages/runtime/src/index.js#L6-L10

To test this, I made this change, and it seemed in a quick test to drop our render times by 4x. (From 4000ms... don't ask :P)

```diff
- React.createElement(MDXContent, props)
+ React.createElement(React.Fragment, null, MDXContent(props))
```
I'm anything but a React DOM expert or familiar with this codebase, so I imagine there's a much, much better way to do this.

### Expected behavior

The MDX runtime creates the same component type tree when given the same input, as far as the React reconciler is concerned.

### Alternatives

I can't think of any, besides "the performance is not quite as good as it could be."

wooorm:
I don’t think `MDXContent({components})` is valid though? Now it’s a) not given all `props`, b) no longer passed to `React.createElement`

marcustyphoon:
It's valid in the sense that the code does "work" and render the MDX. MDXContent is a functional component that just returns the entire MDX body as props, so calling it as a function and using its output as the children prop of a fragment gives the valid component tree. Is it correct code? Quite sure it's not, but it was the first way I thought of to test if the performance gain was real.

I imagine the correct method is something like defining MDXContent as a wrapper component in index.js that just renders its children, then having the dynamically created function invoke it with the MDX as the children prop? (I would test this, but I couldn't actually figure out how to build the repository...)

wooorm:
This repo is currently rather messy.
I’m porting xdm back into it. That repo is workable: https://github.com/wooorm/xdm#evaluatefile-options (a lot of names and options have changed)

marcustyphoon:
Interesting! I saw that repo, but saw the line "No runtime" in the about, and figured it wasn't relevant :D I'll check it out.

[...]

PR #2445, opened 2024-02-25
Always treat _createMdxContent as a JSX component

### LINKED ISSUE #2444: react hmr (fast refresh) breaks when deleting all content of a mdx file
### Initial checklist

3.0.1

### Link to runnable example

https://stackblitz.com/edit/vitejs-vite-sorj9w?file=src%2Ftest.mdx,vite.config.ts&terminal=dev

### Steps to reproduce

1. Go to https://stackblitz.com/edit/vitejs-vite-sorj9w?file=src%2Ftest.mdx,vite.config.ts&terminal=dev  (run `npm run dev` if it is not auto run)
2. Delete all content of `test.mdx`
3. The app is broken after hmr:
image

### Expected behavior

The app should not be broken.

### Actual behavior

The app is broken.

### Runtime

Node v20

### Package manager

npm v10

### OS

macOS

### Build and bundle tools

Vite

remcohaszing:
### Initial checklist

### Description of changes

If the user provides a `wrapper` component, `_createMdxContent` was treated as a JSX component. Otherwise it was invoked as a function. Because `_createMdxContent` uses a hook, this hooks becomes part of the `MDXContent` component conditionally. This breaks React’s rule of hooks.

Closes #2444


wooorm:
Using a function improves performance, that's why it's a function instead of a component. It also allows for diffing; the recommendation is to call the MDXContent yourself if it works even.
The provide function is not always called. And, does not necessarily contain a hook.
Should most users loose performance because of this? Should non-react users, or people who use different or no providers, get a slower experience?

remcohaszing:
I get your point, though I have my doubts about the significance of the performance impact.

An alternative partial solution could be to let the React and Preact providers default the `wrapper` comonent to `Fragment`. This fixes the problem if `wrapper` is toggled via a provider. It doesn’t fix the problem if the user then explicitly passes a `wrapper` component via props and toggles that between `undefined` and a defined value.

remcohaszing:
> > I have my doubts about the significance of the performance impact.
> 
> You didn’t before? #2029, #2062.

That’s a very different issue. I don’t see how that’s related.

---

> The compiler knows when it injects `provideComponents`. I am surprised you don’t want to go about it with that info?

I’m not sure what you mean by this. Perhaps I’m missing something you’re seeing?

wooorm:
Huh, I thought it was added in a PR somewhere. I can’t find the trace quickly in xdm other than https://github.com/wooorm/xdm/commit/f0fde3b5c1f6d2de8c0b40aba63f027b6d7edcfd. Anyway, see https://github.com/mdx-js/mdx/issues/1655 for more on speed.

> I’m not sure what you mean by this. Perhaps I’m missing something you’re seeing?

A hook can exist in the provider.
We can support a hook in the provider, but not decrease performance for people that do not use providers, by switching functionality based on whether the compiler is configured with a provider.
You can make the changes in this PR optional.

remcohaszing:
Ah, I thought you didn’t want this transform in case for example `@mdx-js/vue` or a third party provider is used as `providerImportSource`.

Another alternative I’m thinking of is something along these lines:

```diff
  function _createMdxContent(props) {
-   const _components = {
-     ..._provideComponents(),
-     ...props.components,
-     p: 'p',
-   }
+   const _components = props.components
  }

  export default function MDXContent(props = {}) {
-   const {wrapper: MDXLayout} = {
+   const components = {
      ..._provideComponents(),
      ...props.components,
+     p: 'p',
    }
+   const MDXLayout = components.wrapper
    return MDXLayout
-     ? <MDXLayout {...props}>
-         <_createMdxContent {...props} />
-       </MDXLayout>
-     : _createMdxContent(props)
+     ? <MDXLayout {...props} components={components}>
+         <_createMdxContent {...props} components={components} />
+       </MDXLayout>
+     : _createMdxContent({...props, components: components})
  }
```

The benefit is that `_provideComponents()` is only called once. The downside with this is that the `components` prop constructed does not have a stable identity.

wooorm:
> in case for example @mdx-js/vue

Well, that would be quite nice yeah. That we only have a slow path for providers that might need it. And inline it for those that don’t. But I don’t know how. And generally I’d recommend against providers. So it’s not the common case.

> `_provideComponents()` is only called once

I believe that shouldn’t happen. Component resolution needs to be in the tree at the right place. Otherwise it would break context based APIs: components defined in MDX files (exported from them) also use the provider for “missing” components.

wooorm:
@remcohaszing Coming back to this, where are we standing?
Reading through the code, I guess, the primary thing that is missing is an actual test that used to fail or warn, and works now.
And, can we make this optional: Only do `<_createMdxContent>` where there are providers being passed.

Reading through the thread, what would be nice, is some benchmark: I wrote this code, this way as a function, because it is faster.

Looking at the [OP again](https://github.com/mdx-js/mdx/issues/2444#issuecomment-1962418203), perhaps we can also solve this by generating a `_components = _provideComponents()` in `_createMdxComponents()`, even when it is unused?

wooorm:
@remcohaszing Ping! :)

wooorm:
I would appreciate an answer to my above questions. Including the one on whether “`_components = _provideComponents()` in `_createMdxComponents()`, even when it is unused?” works. Perhaps all this code is not needed. We can only know this if there are actual tests and benchmarks that were broken before and are improved now.

remcohaszing:
> I would appreciate an answer to my above questions.

Sorry for the delay. I’m really busy with other things right now!

---

> Reading through the code, I guess, the primary thing that is missing is an actual test that used to fail or warn, and works now.

The test involves React state updates. I suppose the best way to test it is [React Testing Library](https://testing-library.com/docs/react-testing-library/intro).

> And, can we make this optional: Only do `<_createMdxContent>` where there are providers being passed.
>
> Reading through the thread, what would be nice, is some benchmark: I wrote this code, this way as a function, because it is faster.

I understand that calling `_createMdxComponents(props)` could be faster than `_jsx(_createMdxComponents, { ...props })`. However, I really doubt its significance. We’re talking about avoiding the creation of a single JSX element. It might even be slower, because it could trigger rerenders of the JSX inside `_createMdxComponents` that wouldn’t happen when using JSX. A benchmark could be interesting, but without a benchmark, the current implementation seems like a premature optimization. I can test this using React devtools and a simple React app when I have time.

> Looking at the [OP again](https://github.com/mdx-js/mdx/issues/2444#issuecomment-1962418203), perhaps we can also solve this by generating a `_components = _provideComponents()` in `_createMdxComponents()`, even when it is unused?

Has calling `useContext()` (via `_provideComponents()`) a lower performance impact than calling `_jsx()`.

---

All of my above answers are about React. The exact impact could of course vary per JSX implementation.

wooorm:
> We’re talking about avoiding the creation of a single JSX element.

It’s the top element. The problem is the diffing. See https://github.com/mdx-js/mdx/issues/1655. This is why the generated code is this way.

> Has calling `useContext()` (via `_provideComponents()`) a lower performance impact than calling `_jsx()`.

That is unknown. People can choose what they do in `_provideComponents`.
The current state is optimized to not call if not needed.
React chooses to break when this is optimized.
React chooses if context is slow or fast.

Plus: OP’s problem is about “Delete all content of test.mdx”. That is what produces this problem with React. How often does that happen? Empty MDX document are probably not what we need to make very fast (they likely already are)

PR #2452, opened 2024-03-05
Fix ESM type import in Webpack loader

remcohaszing:
### Initial checklist

### Description of changes

TypeScript 5.3 introduced support for import attributes and `resolution-mode`. This can be used to import types from ESM files into CJS files without a `@ts-ignore` comment.

The `@ts-ignore` comment is stripped from generated type definitions, which leads to a type error for consumers. The module resolution mode is not, which fixes it for users using TypeScript 5.3 or greater.


ChristianMurphy (approved):

wooorm (approved):

wooorm:
Thanks!

PR #2465, opened 2024-04-05
Fix the JSX runtime types in RunOptions

### LINKED ISSUE #2463: Type Error on jsx type with react jsx runtime
### Initial checklist

3.0.1

### Link to runnable example

https://stackblitz.com/edit/vitejs-vite-ahedk8?file=src%2Fmain.tsx

### Steps to reproduce

You can see type error in the sandbox for options parameter of the evaluate function. (in main.tsx file)

### Expected behavior

Doesn't have type error with `evaluate` function when we run the sample code from [documentation](https://mdxjs.com/packages/mdx/#evaluatefile-options):
```
import {evaluate} from '@mdx-js/mdx'

console.log(await evaluate(file, runtime))
```

### Actual behavior

Type error:
```
Argument of type 'typeof import("stackblitz:/node_modules/@types/react/ts5.0/jsx-runtime")' is not assignable to parameter of type 'Readonly<EvaluateOptions>'.
  Types of property 'jsx' are incompatible.
    Type '(type: ElementType<any, keyof IntrinsicElements>, props: unknown, key?: Key | undefined) => ReactElement<any, string | JSXElementConstructor<any>>' is not assignable to type 'Jsx'.
      Types of parameters 'type' and 'type' are incompatible.
        Type 'unknown' is not assignable to type 'ElementType<any, keyof IntrinsicElements>'.
```

### Runtime

_No response_

### Package manager

_No response_

### OS

_No response_

### Build and bundle tools

_No response_

remcohaszing:
### Initial checklist

### Description of changes

`@types/react` now has types for `react/jsx-runtime` and `react/jsx-dev-runtime`. These types are not compatible with the types provided by `hast-util-to-jsx-runtime`, which are used by MDX.

To resolve the issue, all runtime related options have been changed to `unknown`. Since the user is supposed to pass in whatever JSX runtime they imported, this should be sufficient. This removes the dependency on `hast-util-to-jsx-runtime`.

This fixes several type errors. An outdated workaround has been removed from the documentation.

Closes #2463

### DIFF
--- a/packages/mdx/lib/util/resolve-evaluate-options.js
+++ b/packages/mdx/lib/util/resolve-evaluate-options.js
@@ -1,7 +1,27 @@
 /**
- * @import {Fragment, Jsx, JsxDev} from 'hast-util-to-jsx-runtime'
  * @import {MDXComponents} from 'mdx/types.js'
  * @import {CompileOptions} from '../compile.js'
+ */
+
+/**
+ * @typedef {unknown} Fragment
+ */
+
+/**
+ * @callback Jsx
+ * @param {any} type
+ * @param {object} props
+ * @param {any} [key]
+ */
+
+/**
+ * @callback JsxDev
+ * @param {any} type
+ * @param {object} props
+ * @param {any} key
+ * @param {boolean} isStatic
+ * @param {object} [source]
+ * @param {object} [self]
  */
 
 /**
--- a/docs/_asset/editor.jsx
+++ b/docs/_asset/editor.jsx
@@ -204,9 +204,7 @@ function Playground() {
           /** @type {MDXModule} */
           const result = await run(String(file), {
             Fragment,
-            // @ts-expect-error: to do: fix in `hast-util-to-jsx-runtime`.
             jsx,
-            // @ts-expect-error: to do: fix in `hast-util-to-jsx-runtime`.
             jsxs,
             baseUrl: window.location.href
           })
--- a/docs/migrating/v3.mdx
+++ b/docs/migrating/v3.mdx
@@ -52,16 +52,9 @@ You will get a runtime error if these features are used in MDX without
 If you passed the `useDynamicImport` option before, remove it, the behavior
 is now the default.
 
-If you use `react/jsx-runtime`, you might get a TypeScript error (such as
-`Property 'Fragment' is missing in type`), because it is typed incorrectly.
-To remediate this, do:
-
 ```tsx
-import {type Fragment, type Jsx, run} from '@mdx-js/mdx'
-import * as runtime_ from 'react/jsx-runtime'
-
-// @ts-expect-error: the automatic react runtime is untyped.
-const runtime: {Fragment: Fragment; jsx: Jsx; jsxs: Jsx} = runtime_
+import {run} from '@mdx-js/mdx'
+import * as runtime from 'react/jsx-runtime'
 
 const result = await run('# hi', {...runtime, baseUrl: import.meta.url})
 ```
--- a/package-lock.json
+++ b/package-lock.json
@@ -7776,6 +7776,7 @@
       "version": "2.3.0",
       "resolved": "https://registry.npmjs.org/hast-util-to-jsx-runtime/-/hast-util-to-jsx-runtime-2.3.0.tgz",
       "integrity": "sha512-H/y0+IWPdsLLS738P8tDnrQ8Z+dj12zQQ6WC11TIM21C8WFVoIxcqWXf2H3hiTVZjF1AWqoimGwrTWecWrnmRQ==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@types/estree": "^1.0.0",
@@ -7803,12 +7804,14 @@
       "version": "0.2.3",
       "resolved": "https://registry.npmjs.org/inline-style-parser/-/inline-style-parser-0.2.3.tgz",
       "integrity": "sha512-qlD8YNDqyTKTyuITrDOffsl6Tdhv+UC4hcdAVuQsK4IMQ99nSgd1MIA/Q+jQYoh9r3hVUXhYh7urSRmXPkW04g==",
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/hast-util-to-jsx-runtime/node_modules/style-to-object": {
       "version": "1.0.6",
       "resolved": "https://registry.npmjs.org/style-to-object/-/style-to-object-1.0.6.tgz",
       "integrity": "sha512-khxq+Qm3xEyZfKd/y9L3oIWQimxuc4STrQKtQn8aSDRHb8mFgpukgX1hdzfrMEW6JCjyJ8p89x+IUMVnCBI1PA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "inline-style-parser": "0.2.3"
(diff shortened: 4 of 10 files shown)


packages/mdx/lib/util/resolve-evaluate-options.js
```diff
@@ -29,13 +26,13 @@
...
- * @property {Fragment} Fragment
+ * @property {unknown} Fragment
  *   Symbol to use for fragments (**required**).
- * @property {Jsx | null | undefined} [jsx]
+ * @property {unknown} [jsx]
  *   Function to generate an element with static children in production mode.
- * @property {JsxDev | null | undefined} [jsxDEV]
+ * @property {unknown} [jsxDEV]
  *   Function to generate an element in development mode.
- * @property {Jsx | null | undefined} [jsxs]
```
wooorm, line 38: Should we at least use something like `Function`?
We could also improve the `hast-util-to-jsx-runtime` types?
Fixing the type error is nice, but removing all type errors around this, not so sure?

remcohaszing: We could use `Function`, or better: `(type: any, props: any) => any`.

We should improve `hast-util-to-jsx-runtime`, but personally I think this type isn’t worth having an extra dependency regardless.

wooorm: > or better: `(type: any, props: any) => any`.

Right, and even better would be an even more correct type. We do know a lot about props and key. And at that point, it’s nice to have the more complex type defined and tested in a single place?

wooorm (commented):
I guess the breaking change happened in the other PR, but `Fragment` and similar were types that were exposed from MDX. And now aren‘t.

wooorm:
I think we’ve had this conversation partially a couple times already, the gist of my stance is: generally, we have types for things. so why are we now removing types?
I don’t consider hiding type errors by treating everything as `unknown` an improvement.
These things are at least functions. And probably stricter than that. Not `unknown`/`any`?

remcohaszing:
I think we should aim at the following:

1. :heavy_check_mark: Working runtime code.
2. ❌ Types that make user code at least compile with TypeScript.
3. ❌ Types that provide more helpful type checks.

While (3) is great, I strongly believe we need to aim for (2) first. That’s what this PR solves. I don’t think any users are helped with types that prevent the user from using the code without type casting (to / from `unknown` even) or `@ts-ignore` checks.

I doubt it’s useful, because users are typically supposed to pass in a known runtime, not define one. However, it **is** possible to use the following somewhat stricted types:

```js
/**
 * @typedef {unknown} Fragment
 */

/**
 * @callback Jsx
 * @param {any} type
 * @param {object} props
 * @param {any} [key]
 */

/**
 * @callback JsxDev
 * @param {any} type
 * @param {object} props
 * @param {any} key
 * @param {boolean} isStatic
 * @param {object} [source]
 * @param {object} [self]
 */
```

- `Fragment`: was already `unknown`.
- `type`: must be `any`. The real type of `type` is defined by the JSX runtime. It can’t be `unknown`, because the type of `type` must be assignable to what the JSX runtime expects.
- `props`: must be an object.
- `key`: same thing as `type`
- `isStatic`: of this we actually know it must be a boolean.
- `source`: we know the shape, it might be a bit dangerous to assume the type provided by the JSX runtime is compatible.
- `self`: is always an object.

The problem with those types is that `type-coverage` complains about some of the arguments being `any`. So CI fails.

wooorm:
>  I strongly believe we need to aim for (2) first

Why? I don‘t think we’ve done that before: implement an `any` instead of actually solving types. Why do you want to implement a simple workaround instead of solving this bug?

> I doubt it’s useful, because users are typically supposed to pass in a known runtime, not define one. However, it is possible to use the following somewhat stricted types:

I do think that people should be able to type check `jsx`/`jsxs`/`Fragment`.

> The problem with those types is that type-coverage complains about some of the arguments being any. So CI fails.

That sounds like something that can be solved?

remcohaszing:
> > I strongly believe we need to aim for (2) first
> 
> Why? I don‘t think we’ve done that before: implement an `any` instead of actually solving types. Why do you want to implement a simple workaround instead of solving this bug?

TypeScript is supposed to help users. Types that are wrong are significantly worse than types that are loose. What you call a _simple workaround_, I call an imperfect, but pragmatic solution.

We actually do use `any` in a couple of places in this repo. It happens in places where we act as the user. This PR removes those occurrences.

- In [`docs/_asset/editor.jsx`](https://github.com/mdx-js/mdx/pull/2465/files#diff-b3ed649084f45f1bdd398127b3fbc6192ba07df31381e8dbee9254bfada8da31L207-R208) we use `@ts-expect-error` to suppress a type error. This is a somewhat worse version of casting it to any.
- In [`docs/migrating/v3.mdx`](https://github.com/mdx-js/mdx/pull/2465/files#diff-679ca51ad6d21fa798def40edbe6d21f3c7e9e42633e8511cc8b440eed67c92fL63) we tell our users to use `@ts-expect-error`, which again is a worse version of casting to any.
- The same happens in [`packages/mdx/test/evaluate.js`](https://github.com/mdx-js/mdx/pull/2465/files#diff-5d7b573670da8b3ab45636dd6493c29101f2d318cbaca9f9e2c944ff9a6c51beL14-L19)
- In [`packages/react/test/index.jsx`](https://github.com/mdx-js/mdx/pull/2465/files#diff-727827a4e281640e8c765de2f6446b79765ce246ad644562fdf2c9b408c40be7L14-L16) we cast `as unknown as SomethingElse`, which is the same as `as any as SomethingElse`.

> > I doubt it’s useful, because users are typically supposed to pass in a known runtime, not define one. However, it is possible to use the following somewhat stricted types:
> 
> I do think that people should be able to type check `jsx`/`jsxs`/`Fragment`.

Sure, it’s nice, but the user is bound to the types of their JSX framework and MDX. They just import one thing and pass it to a function. They don’t have a say in how any of it is typed or implemented. I’m not saying stricter types are bad, I’m saying the status quo (incorrect types) is significantly worse than loose types.

> Also keep in mind that these types are from another project where this is important. It’s important in several of our projects. We can solve it once, in a good way?

`hast-util-to-jsx-runtime` is only used in MDX for its (incorrect) types. Yes, the problem needs to be solved there as well, but there’s no reason to block a fix in MDX for that.

> > The problem with those types is that type-coverage complains about some of the arguments being any. So CI fails.
> 
> That sounds like something that can be solved?

I don’t see how, except for disabling `type-coverage`, at least for the file that contains the `any`.

wooorm:
> They just import one thing and pass it to a function.

I disagree.
Framework developers are also humans: they might want to check whether their runtime works with MDX.
There are also tools that are not frameworks, but wrap JSX runtimes: think emotion, [restyle](https://github.com/souporserious/restyle). They might want to check whether their runtime works with MDX.
And finally there are humans that for other reasons just wrap JSX runtimes, such as for debugging or to learn.

> They don’t have a say in how any of it is typed or implemented.

I disagree, I think they do.

> I’m saying the status quo (incorrect types) is significantly worse than loose types.

Facebook did not ship types. It was a `@ts-expect-error`.
Now they ship types. It is a `@ts-expect-error`.

> hast-util-to-jsx-runtime is only used in MDX for its (incorrect) types.

This is intentional: if we can get correct types from another place that is fine too.
I argue that most of the bugs we uncovered in JSX runtimes (svelte, solid, vue, preact) would’ve been improved if there were some good types/docs for the automatic runtime.

> Yes, the problem needs to be solved there as well, but there’s no reason to block a fix in MDX for that.

We often implement solutions in the place where the problem originates, instead of patching intermediate packages?

> I don’t see how, except for disabling type-coverage, at least for the file that contains the any.

`type-coverage` supports ignore comments.
`type-coverage` also supports choosing which files to ignore: you can put `any`s in a particular file and ignore only that file.

wooorm:
Externally fixed, thanks!

PR #2498, opened 2024-06-25
Refactor to use TypeScript `@import` JSDoc tags

remcohaszing:
### Initial checklist

### Description of changes

Previously we used `@typedef` tags to simulate type imports. The problem is that are not really imports. They define and export a new type. We now use `@import` introduced by TypeScript 5.5 to actually import types.

The pattern using `@typedef` is still used in index files to export types.

This change also replaces the type casting inside MDX files with definitions of the `Props` types.

This also replaces usage of the global `JSX` namespace with proper types. For TypeScript<5.1 the correct return type of React components is `ReactElement`, for Preact `VNode`. For TypeScript>=5.1, the correct return type for React components is `ReactNode`, for Preact `ComponentChildren`.

`ReactElement` / `VNode` and are compatible with TypeScript>=5.1. This is why those are used for public facing APIs, but `ReactNode` / `ComponentChildren` is used for internal components.

### DIFF
--- a/docs/_asset/editor.jsx
+++ b/docs/_asset/editor.jsx
@@ -2,20 +2,20 @@
 /* @jsxImportSource react */
 
 /**
- * @typedef {import('@wooorm/starry-night').Grammar} Grammar
- * @typedef {import('estree').Node} EstreeNode
- * @typedef {import('estree').Program} Program
- * @typedef {import('hast').Nodes} HastNodes
- * @typedef {import('hast').Root} HastRoot
- * @typedef {import('mdast').Nodes} MdastNodes
- * @typedef {import('mdast').Root} MdastRoot
- * @typedef {import('mdast-util-mdx-jsx').MdxJsxAttribute} MdxJsxAttribute
- * @typedef {import('mdast-util-mdx-jsx').MdxJsxAttributeValueExpression} MdxJsxAttributeValueExpression
- * @typedef {import('mdast-util-mdx-jsx').MdxJsxExpressionAttribute} MdxJsxExpressionAttribute
- * @typedef {import('mdx/types.js').MDXModule} MDXModule
- * @typedef {import('react-error-boundary').FallbackProps} FallbackProperties
- * @typedef {import('unified').PluggableList} PluggableList
- * @typedef {import('unist').Node} UnistNode
+ * @import {Grammar} from '@wooorm/starry-night'
+ * @import {Node as EstreeNode, Program} from 'estree'
+ * @import {Nodes as HastNodes, Root as HastRoot} from 'hast'
+ * @import {Nodes as MdastNodes, Root as MdastRoot} from 'mdast'
+ * @import {
+    MdxJsxAttribute,
+    MdxJsxAttributeValueExpression,
+    MdxJsxExpressionAttribute
+ * } from 'mdast-util-mdx-jsx'
+ * @import {MDXModule} from 'mdx/types.js'
+ * @import {ReactNode} from 'react'
+ * @import {FallbackProps} from 'react-error-boundary'
+ * @import {PluggableList} from 'unified'
+ * @import {Node as UnistNode} from 'unist'
  */
 
 /**
@@ -35,7 +35,7 @@
  *   OK.
  * @property {true} ok
  *   Whether OK.
- * @property {JSX.Element} value
+ * @property {ReactNode} value
  *   Result.
  *
  * @typedef {EvalNok | EvalOk} EvalResult
@@ -315,7 +315,7 @@ function Playground() {
   const scope = formatMarkdown ? 'text.md' : 'source.mdx'
   // Cast to actual value.
   const compiledResult = /** @type {EvalResult | undefined} */ (evalResult)
-  /** @type {JSX.Element | undefined} */
+  /** @type {ReactNode | undefined} */
   let display
 
   if (compiledResult) {
@@ -579,9 +579,9 @@ function Playground() {
 
 /**
  *
- * @param {Readonly<FallbackProperties>} properties
+ * @param {Readonly<FallbackProps>} properties
  *   Properties.
- * @returns {JSX.Element}
+ * @returns {ReactNode}
  *   Element.
  */
 function ErrorFallback(properties) {
@@ -601,7 +601,7 @@ function ErrorFallback(properties) {
 /**
  * @param {DisplayProperties} properties
  *   Properties.
- * @returns {JSX.Element}
+ * @returns {ReactNode}
  *   Element.
  */
 function DisplayError(properties) {
--- a/docs/blog/index.mdx
+++ b/docs/blog/index.mdx
@@ -1,3 +1,12 @@
+{/**
+  * @import {Item} from '../_component/sort.js'
+  */}
+
+{/**
+  * @typedef Props
+  * @property {Item} navigationTree
+  */}
+
 import assert from 'node:assert/strict'
 import {BlogGroup} from '../_component/blog.jsx'
 
@@ -16,11 +25,6 @@ The latest news about MDX.
 
 {
   (function () {
-    /**
-     * @typedef {import('../_component/sort.js').Item} Item
-     */
-
-    /** @type {Item} */
     const navigationTree = props.navigationTree
     const category = navigationTree.children.find(function (item) {
       return item.name === '/blog/'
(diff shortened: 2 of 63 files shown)


docs/_asset/editor.jsx
```diff
@@ -2,20 +2,20 @@
...
- * @typedef {import('unified').PluggableList} PluggableList
- * @typedef {import('unist').Node} UnistNode
+ * @import {Grammar} from '@wooorm/starry-night'
+ * @import {Node as EstreeNode, Program} from 'estree'
+ * @import {Nodes as HastNodes, Root as HastRoot} from 'hast'
+ * @import {Nodes as MdastNodes, Root as MdastRoot} from 'mdast'
+ * @import {
+    MdxJsxAttribute,
+    MdxJsxAttributeValueExpression,
+    MdxJsxExpressionAttribute
```
remcohaszing, line 12: Using asterisks for multiline imports caused problems, but not in the TypeScript playground. I’m looking into this.

remcohaszing: https://github.com/microsoft/TypeScript/issues/59011

wooorm: I’d work around it by not wrapping 🤷‍♂️

docs/_asset/editor.jsx
```diff
@@ -601,7 +601,7 @@ function ErrorFallback(properties) {
 /**
  * @param {DisplayProperties} properties
  *   Properties.
- * @returns {JSX.Element}
+ * @returns {ReactNode}
```
wooorm, line 604: better in separate PRs

docs/blog/index.mdx
```diff
@@ -1,3 +1,12 @@
+{/**
+  * @import {Item} from '../_component/sort.js'
+  */}
+
+{/**
+  * @typedef Props
+  * @property {Item} navigationTree
+  */}
```
wooorm, line 8: not 100% on this style? I’d lean to:

```
{
  /**
   * @import {Item} from '../_component/sort.js'
   */

  /**
   * @typedef Props
   * @property {Item} navigationTree
   */
}
```

packages/mdx/index.js
```diff
@@ -1,12 +1,12 @@
...
+ * @typedef {import('hast-util-to-jsx-runtime').Jsx} Jsx
+ * @typedef {import('hast-util-to-jsx-runtime').JsxDev} JsxDev
  * @typedef {import('./lib/util/resolve-evaluate-options.js').UseMdxComponents} UseMdxComponents
  * @typedef {import('./lib/compile.js').CompileOptions} CompileOptions
  * @typedef {import('./lib/core.js').ProcessorOptions} ProcessorOptions
- * @typedef {import('./lib/evaluate.js').EvaluateOptions} EvaluateOptions
- * @typedef {import('./lib/run.js').RunOptions} RunOptions
+ * @typedef {import('./lib/util/resolve-evaluate-options.js').EvaluateOptions} EvaluateOptions
+ * @typedef {import('./lib/util/resolve-evaluate-options.js').RunOptions} RunOptions
  */
```
wooorm, line 10: I like an `index.d.ts`, as it has `export type {} from`, here it would get rid of the typedef/type alias  https://github.com/rehypejs/rehype-document/blob/main/index.d.ts

remcohaszing: I’m working on a follow-up PR where I think this is better discussed. I do agree `export {} from` is an improvement over a typedef.

docs/blog/index.mdx
wooorm: Hmm, I feel quite strongly that entering/exiting JS each time is ugly.

And that if you’d put something in there (like a jsx element), that it would be indented 🤔

I’m fine to different typedefs in different comments.

remcohaszing: > Hmm, I feel quite strongly that entering/exiting JS each time is ugly.

I don’t really mind either style for typedefs. For function docs I would find it weird to combine though, but that’s not applicable in this particular case.

Another thing is that commenting is based on `{/*` and `*/}` delimiters. Though I don’t think it makes sense to uncomment jsdoc comments.

> And that if you’d put something in there (like a jsx element), that it would be indented 🤔

That makes perfect sense. This also reminds me that it’s possible to have multiple comments in one a flow expression, but just one JS expression. JSDoc has meaning, which makes me lean towards preferring one doc per flow expression again. But having the types grouped using one flow expression is also kind of nice.

---

Anyway, these are all just nitty details. I’ll adjust them to use one MDX flow expression.

wooorm: > For function docs I would find it weird to combine though, but that’s not applicable in this particular case.

Can you clarify what you dislike/like with examples?


> Though I don’t think it makes sense to uncomment jsdoc comments.

Agreed. These things are more like type annotations, which *currently* use special comments, because that’s what’s possible now.

> This also reminds me that it’s possible to have multiple comments in one a flow expression, but just one JS expression. JSDoc has meaning, which makes me lean towards preferring one doc per flow expression again. But having the types grouped using one flow expression is also kind of nice.

Right, right, I was thinking about that at first too, but then there’s a difference: these are allowed to be empty! And comments are deemed empty here too. 0 or more empty is still empty. So, zero-or-more empty, then 0-1 expression.

To clarify, some examples of what I think looks good.

Regular comment:

```mdx
# hi

{/* just a smol comment */}

hello
```

More complex expression:

```mdx
# hi

{
  (function () {
    return 1
  })()
}

hello
```

One JSDoc:

```mdx
{
  /**
   * @import {foo} from 'bar'
   */
}

# hi

hello
```

Multi JSDoc:

```mdx
{
  /**
   * @import {foo} from 'bar'
   */

  /**
   * @typedef Options
   * @property {number} something
   */
}

# hi

hello
```

Multi JSDoc and expression:

```mdx
# hi

{
  /**
   * @import {foo} from 'bar'
   */

  /**
   * @typedef Options
   * @property {number} something
   */

  (function () {
    return 1
  })()
}

hello
```

Also, was just thinking about how [do-expressions](https://github.com/tc39/proposal-do-expressions) will be nice!

```mdx
{
  do {
    if (loggedIn) {
      <LogoutButton />
    } else {
      <LoginButton />
    }
  }
}
```

docs/blog/index.mdx
```diff
@@ -1,3 +1,14 @@
+{
+  /**
+   * @import {Item} from '../_component/sort.js
+   */
+
+  /**
+   * @typedef Props
+   * @property {Item} navigationTree
+   */
+}
```
wooorm, line 10: <3

docs/community/index.mdx
```diff
@@ -1,3 +1,14 @@
+{
+  /**
+   * @import {Item} from '../_component/sort.js
+   */
+
+  /**
+   * @typedef Props
+   * @property {Item} navigationTree
+   */
+}
```
wooorm, line 10: Like this it’s also more similar to frontmatter 🤔

wooorm (approved):

wooorm:
In this PR, I saw the existing code:

```
// Register MDX nodes in mdast:
/// <reference types="remark-mdx" />
```

I think we can now change, and recommend `@import {} from 'remark-mdx`?

wooorm:
Thanks :)

docs/blog/index.mdx
remcohaszing: > Can you clarify what you dislike/like with examples?

If there are multiple exports with JSDoc annotations, It feels weird to me to merge the mdx flow expression the typedefs with that of the JSDoc of the first export.

```mdx
{
  /**
   * @typedef Props
   * @property {string} whatever
   */

  /**
   * @param {string} arg
   * @returns undefined
   */
}
export function someExport(arg) {}

{
  /**
   * @param {string} arg
   * @returns undefined
   */
}
export function anotherExport(arg) {}
```

> Also, was just thinking about how [do-expressions](https://github.com/tc39/proposal-do-expressions) will be nice!

Yes, that would be so nice!

wooorm: ok, I can see this, for this first example. But that feels like an edge case to me, induced because we allow `export` statements directly in markdown 🤔

I’m fine with breaking this like:

```mdx
{
  /**
   * @import {foo} from 'bar'
   */

  /**
   * @typedef Options
   * @property {number} something
   */
}

{
  /**
   * @param {string} arg
   * @returns {undefined}
   */
}
export function someExport(arg) {}
```

PR #2508, opened 2024-07-07
Fix link

artola:
### Initial checklist

### Description of changes

The documentation says:

> `@mdx-js/mdx` is a unified pipeline — wrapped so that most folks don’t need to know about unified: [core.js#L65](https://github.com/mdx-js/mdx/blob/main/packages/mdx/lib/core.js#L65).

and such line states:

> `@Property {string | null | undefined} [pragma='React.createElement']`

I would suggest to avoid linking to the source code, as it is likely to be broken again. Even a permalink will become obsolete overtime.

### DIFF
--- a/packages/mdx/readme.md
+++ b/packages/mdx/readme.md
@@ -1083,7 +1083,7 @@ what unified does: please read through the [`unifiedjs/unified`][unified] readme
 (the part until you hit the API section is required reading).
 
 `@mdx-js/mdx` is a unified pipeline — wrapped so that most folks don’t need to
-know about unified: [`core.js#L65`][core].
+know about unified, but can easily integrate their remark and rehype plugins.
 The processor goes through these steps:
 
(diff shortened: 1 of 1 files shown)


packages/mdx/readme.md
```diff
@@ -1083,7 +1083,7 @@ what unified does: please read through the [`unifiedjs/unified`][unified] readme
 (the part until you hit the API section is required reading).
 
 `@mdx-js/mdx` is a unified pipeline — wrapped so that most folks don’t need to
-know about unified: [`core.js#L65`][core].
+know about unified, but can easily integrate their remark and rehype plugins.
```
wooorm, line 1086:
```suggestion
know about unified.
```

I’d say just drop it instead of a new sentence. The next bullet points describe the unified processor!

wooorm (approved):

wooorm:
Thank you! :)

Issue #2503, opened 2024-07-03
Docs: Missing explicit docs and workaround for overriding literal HTML elements in MDX


karlhorky:
### Initial checklist

- [X] I read the [support docs](https://mdxjs.com/community/support/)
- [X] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [X] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/HEAD/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

@mdx-js/react@3.0.1, next@14.2.4, react@^18

### Steps to reproduce

1. Set up [the Next.js App Router + MDX `app-dir-mdx` example from Remco](https://codesandbox.io/p/sandbox/github/remcohaszing/next.js/tree/fix-mdx-components-examples/examples/app-dir-mdx)
2. Remove `experimental.mdxRs` from `next.config.js`
3. Add a line `video: () => <div>video replacement</div>` in `mdx-components.tsx`
4. Add a line `<video />` in `app/message.mdx`
5. Observe `<video />` appearing in HTML in DevTools, instead of the `<div>`
6. Research and find the following information in docs and GitHub issues / PRs
   1. https://github.com/mdx-js/mdx/issues/821
   2. https://github.com/kentcdodds/mdx-bundler/issues/160
   3. https://github.com/mdx-js/mdx/pull/2050
   4. Workaround: https://github.com/mdx-js/mdx/pull/2052#issuecomment-1140519087
   5. Workaround: https://github.com/kentcdodds/mdx-bundler/issues/160#issuecomment-1140526121
   7. https://github.com/remcohaszing/rehype-mdx-import-media/issues/2#issuecomment-2198361734
6. Attempt the workaround from 5iv. and 5v. (adding the `rehypePlugin` suggested by @wooorm)
7. Observe `<video />` still appearing in HTML in DevTools, instead of the `<div>`

![Screenshot 2024-07-03 at 13 38 04](https://github.com/mdx-js/mdx/assets/1935696/00d37418-e661-41d9-99c1-347fe02104c9)

Or, to skip steps 1-6, a reproduction repo:

- https://github.com/karlhorky/repro-mdx-literal-html-elements-overriding-with-custom-components

[...]

PR #2511, opened 2024-07-09
Improve scanning and visual structure of example

karlhorky:
### Initial checklist

### Description of changes

Riffing on 044e8b2a272186be04d16c270f5621718cc8a57e

### DIFF
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -350,7 +350,7 @@ import type {} from 'mdx'
 // ---cut---
 import Example from './example.mdx'
 
-console.log(
+export default function Component() {
   <Example
     components={{
       a(props) {
@@ -361,27 +361,27 @@ console.log(
       }
     }}
   />
-)
+}
 ```
 
 …we’d get:
 
-{
-  <div className="card">
-    <ul>
-      <li><a href="#alpha" style={{borderTop: '1px dotted', color: 'violet'}}>markdown syntax</a></li>
-      <li><a href="#bravo">JSX with a lowercase name</a></li>
-      <li><a href="#charlie" style={{borderTop: '1px dashed', color: 'tomato'}}>JSX with a capitalized name</a></li>
-    </ul>
-  </div>
-}
+```jsx
+<ul>
+  <li><a href="#alpha" style={{borderTop: '1px dotted', color: 'violet'}}>markdown syntax</a></li>
+  <li><a href="#bravo">JSX with a lowercase name</a></li>
+  <li><a href="#charlie" style={{borderTop: '1px dashed', color: 'tomato'}}>JSX with a capitalized name</a></li>
+</ul>
+```
 
-Observe that the first link (`#alpha`) is dotted and violet.
-That’s because `a` is the HTML equivalent for the markdown syntax being used.
-The second link (`#bravo`) remains unchanged,
-because in JSX syntax `a` is a literal tag name.
-The third link (`#charlie`) is dashed and tomato,
-as in JSX syntax `Link` is a reference.
+Observe that:
+
+1. The first link (`#alpha`) is dotted and violet, because `a` is the HTML
+   equivalent for the markdown syntax `[]()`.
+2. The second link (`#bravo`) remains unchanged, because in JSX syntax `a` is a
+   literal tag name, which [MDX does not modify][mdx-no-literals-modification].
+3. The third link (`#charlie`) is dashed and tomato, because in this JSX syntax
+   `Link` is a reference to a component.
 
 ### Layout
 
@@ -591,6 +591,8 @@ providers: pass components explicitly.
 [start]: /docs/getting-started/
 
 [what]: /docs/what-is-mdx/
+
+[mdx-no-literals-modification]: https://github.com/mdx-js/mdx/issues/821
 
 [mdx-react]: /packages/react/
 
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -209,7 +209,8 @@ console.log(<Example name="Mars" year={2022} />)
 ### Components
 
 There is one special prop: `components`.
-It takes an object mapping component names to components.
+It takes an object mapping component names to components, which MDX uses to
+replace default components with custom components.
 Take this example:
 
 ```mdx path="example.mdx"


docs/docs/using-mdx.mdx
```diff
@@ -209,7 +209,8 @@ console.log(<Example name="Mars" year={2022} />)
 ### Components
 
 There is one special prop: `components`.
-It takes an object mapping component names to components.
+It takes an object mapping component names to components, which MDX uses to
+replace default components with custom components.
```
ChristianMurphy, line 213: There are no default components. 🤔 
Perhaps wording like:

> It takes an object mapping component names to components. MDX adds the component names to the scope of the document to allow the component to be used.

docs/docs/using-mdx.mdx
```diff
@@ -350,7 +351,7 @@ import type {} from 'mdx'
 // ---cut---
 import Example from './example.mdx'
 
-console.log(
+export default function Component() {
```
ChristianMurphy, line 354: I'm not sure how this change relates to the goal of improving scanning or visual structure.

docs/docs/using-mdx.mdx
```diff
@@ -361,27 +362,27 @@ console.log(
...
-    </ul>
-  </div>
-}
+```jsx
+<ul>
+  <li><a href="#alpha" style={{borderTop: '1px dotted', color: 'violet'}}>markdown syntax</a></li>
+  <li><a href="#bravo">JSX with a lowercase name</a></li>
+  <li><a href="#charlie" style={{borderTop: '1px dashed', color: 'tomato'}}>JSX with a capitalized name</a></li>
+</ul>
+```
```
ChristianMurphy, line 376: This feels a bit like change for change sake.
The styles are there to make it clear visually what was rendered.
I'm not opposed to having a toggle between source and rendered.
But just source 🤷

docs/docs/using-mdx.mdx
```diff
@@ -361,27 +362,27 @@ console.log(
...
-Observe that the first link (`#alpha`) is dotted and violet.
-That’s because `a` is the HTML equivalent for the markdown syntax being used.
-The second link (`#bravo`) remains unchanged,
-because in JSX syntax `a` is a literal tag name.
-The third link (`#charlie`) is dashed and tomato,
-as in JSX syntax `Link` is a reference.
+1. The first link (`#alpha`) is dotted and violet, because `a` is the HTML
+   equivalent for the markdown syntax `[]()`.
+2. The second link (`#bravo`) remains unchanged, because in JSX syntax `a` is a
+   literal tag name, which [MDX does not modify][mdx-no-literals-modification].
```
ChristianMurphy, line 383: Explained more in https://github.com/mdx-js/mdx/issues/2503#issuecomment-2218365581
There's more value in describing what MDX _is_, I'd like to avoid the rabbit whole of trying to list all the things MDX isn't.

> The second link (`#bravo`) is rendered as-is, because in JSX syntax `a` is a
>   literal tag name.

docs/docs/using-mdx.mdx
karlhorky: I was looking to use a verb like "replace" or "override", to:
1. be explicit that it is transforming something (I'm ok with changing the "something" of "default components" - I agree that it's not good)
2. use terminology closer to what community members have done

docs/docs/using-mdx.mdx
karlhorky: Ah yeah sorry, that was me "fixing" what I thought was a mistake (console.log would not be as common to find wrapping around JSX in a .jsx document as a function)

docs/docs/using-mdx.mdx
karlhorky: How about default source code, with switcher to rendered?

docs/docs/using-mdx.mdx
karlhorky: My reasoning behind this and behind #2503 is that:

- I think it could be a misconception that should be broken
- on a docs page about MDX, I wouldn't find it unuseful or off-topic to connect a note about "what JSX does" with context about how that connects to MDX (instead of leaving it to the user to fill in the blanks implicitly by reading other docs) 

Just my opinion and suggestion though, if not accepted then I'll revert this

docs/docs/using-mdx.mdx
wooorm: You want to add one concrete sentence here.
I don’t think human language is expressive enough for that.
I think it needs to be illustrated.

there are no default components; a “custom” component implies non-custum components, what are those?
when you mention scope, I think it needs to be explained too, what does scope mean in MDX; it’s also not not *really* what happens.

> it is transforming something

It is *not* always transforming something: see the following lines. `Planet` is used and `Planet` is passed. That isn’t a transform.

> use terminology closer to what community members have done

Do you have sources for these?

docs/docs/using-mdx.mdx
wooorm: `console.log` is an intentional consistent choice across the website and the readmes

docs/docs/using-mdx.mdx
wooorm: Sorry, but I really do not see what you are seeing. What does this add for you? This document, especially this section on MDX content, is about how humans can *use* MDX to get results. It’s not about pseudocode JSX. I really have faith that humans can see something and then read the next paragraph explaining what they see and then grasp it.

docs/docs/using-mdx.mdx
wooorm: > I think it could be a misconception that should be broken

It is assumed knowledge from JSX. Then it is covered in-depth in “What is MDX”. Then it is explained in excruciating detail in the paragraphs above. **Then** this example covers it. No more words are needed.

You wrote a long issue about how you went down a rabbit hole by reading old issues. I do not want to send users down rabbit holes of 5 year old issues.

docs/docs/using-mdx.mdx
karlhorky: sources:

- https://github.com/mdx-js/mdx/issues/1385
- https://github.com/kentcdodds/mdx-bundler/issues/160
- https://github.com/mdx-js/mdx/issues/821
- i'm guessing there's more, with further research

docs/docs/using-mdx.mdx
karlhorky: 1. until this change, so far on https://mdxjs.com/docs/what-is-mdx/, https://mdxjs.com/docs/getting-started/ and https://mdxjs.com/docs/using-mdx/, there have been many code blocks with no results shown - all of a sudden, changing to show DOM without the user being able to see the resulting code seems unusual and jarring, and too large of a jump
2. can also turn this around and ask the other way: why not let the user see the resulting code if it's helpful? (in my experience very helpful, but again, maybe you disagree) why make them go to the DevTools to see the code?

wooorm: 1. indeed, but your point can be extended to *both* a) code *and* b) rendered result. Either is new. You wanted this new illustration as explanations in prose were not clear enough. I chose the rendered result (b) because the code (a) is not real even thought it implies that (we generate javascript code that is complex and different); the code relates to particular unexplained runtime options; actual code is not needed here, because it’s about whether things turn tomato or violet.
2. I don’t think it’s helpful (it provides detail that is irrelevant here, thus confusing the reader with unneeded info, which only should care about tomato vs violet); I don’t understand why someone would open devtools here

wooorm:
Thanks but no thanks. Appreciate you wanting to help. But I do not agree with several of the proposed changes (there are no default components; `console.log` is intentional; I prefer actual rendered things here over pseudocode; I don’t think linking to an issue is good here).

PR #2513, opened 2024-07-14
Refactor to improve wording

filippovd20:
importing mdx not md

### Initial checklist

### Description of changes

Fixing a typo in the code snipped importing from `mdx` not `md` directly.

### DIFF
--- a/docs/docs/using-mdx.mdx
+++ b/docs/docs/using-mdx.mdx
@@ -263,7 +263,7 @@ MDX using imported components!
 Because MDX files *are* components, they can also import each other:
 
 ```mdx path="example.mdx"
-import License from './license.mdx' // Assumes an integration is used to compile MDX -> JS.
+import License from './license.md' // Assumes an integration is used to compile markdown -> JS.
 import Contributing from './docs/contributing.mdx'
 
 # Hello world
@@ -468,7 +468,7 @@ cumbersome.
 Like so:
 
 ```mdx path="post.mdx"
-import License from './license.mdx' // Assumes an integration is used to compile MDX -> JS.
+import License from './license.md' // Assumes an integration is used to compile markdown -> JS.
 import Contributing from './docs/contributing.mdx'
 
 # Hello world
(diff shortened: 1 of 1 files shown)


wooorm:
Hi! Why do you think this is a mistake? I think this is intended!

remcohaszing:
I think the comment should say `Assumes an integration is used to compile markdown -> JS.`, not `Assumes an integration is used to compile MDX -> JS.`.

wooorm:
I kinda get what you're saying but  saying "md" there feels a bit more static. Like using marked/remark/markdown-it.
Perhaps the "to JS" part cancels that out enough.
The point, I think, when I added those comments, is that it doesn't work normally, to import these files. They need *some* integration, not one particular one. Likely one of ours, but not always.

filippovd20:
@wooorm @remcohaszing if you look into the code snippet from `post.mdx`, both License and Contributing are regular components with props. `Contributing `is imported from an `mdx `module assuming everything is configured as expected. License on the other hand is imported from `.md` which is definitely a typo. There's no way you can import a component from an md file directly. If you could do that then why bother about mdx then?

wooorm:
>  There's no way you can import a component from an md file directly. If you could do that then why bother about mdx then?

Yes you can, our packages can do that! Please see what MDX is, particularly this note: https://mdxjs.com/docs/what-is-mdx/#mdx-syntax.
Well, one reason where you go wrong: markdown doesn’t have imports so markdown couldn’t import markdown.

filippovd20:
@wooorm I'm wrong and you're right. Indeed with the bundler configured it also compiles regular markdown to JS which is also described in Quick Start. Do you think changing the comment would be beneficial to stress the fact that md is also compiled to JS? something like this:
`// Assumes a bundler is used to compile MD -> JS`

docs/docs/using-mdx.mdx
```diff
@@ -263,7 +263,7 @@ MDX using imported components!
 Because MDX files *are* components, they can also import each other:
 
 ```mdx path="example.mdx"
-import License from './license.md' // Assumes an integration is used to compile MDX -> JS.
+import License from './license.mdx' // Assumes an integration is used to compile MDX -> JS.
```
wooorm, line 266:
```suggestion
import License from './license.md' // Assumes an integration is used to compile markdown -> JS.
```

wooorm (approved):

wooorm:
thanks :)

PR #2517, opened 2024-07-22
Add Bun section to Getting started

karlhorky:
### Initial checklist

### Description of changes

Add footnote to `@mdx-js/esbuild` section that Bun is supported via the esbuild-compatible plugin API (I confirmed support and [Bun docs confirm it too](https://bun.sh/docs/runtime/plugins#third-party-plugins))

This actually also enables MDX eval support in scripts, which is not a straightforward thing to set up with `esbuild` (because [the Transform API, used by programs like `tsx`, does not support plugins](https://github.com/privatenumber/tsx/issues/491#issuecomment-1987041573))

### Alternatives considered

1. Don't include a code block, but rather link to https://bun.sh/docs/runtime/plugins#third-party-plugins. downsides:
   1. these docs are not dedicated to `@mdx-js/esbuild`, so it could disappear at any time
   2. it's not a full example to be used on its own
2. Add this instead to https://mdxjs.com/packages/esbuild/ . downsides:
   1. less discoverable (no ability to `cmd-f` -> search for "Bun" on "Getting Started" page)

### DIFF
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -817,9 +817,23 @@ See its readme on how to configure it.
 
 #### Bun
 
-MDX files can be imported in Bun by using
+MDX files can be imported in [Bun][] by using
 [`@mdx-js/esbuild`][mdx-esbuild].
-See [the section above][bundler-esbuild] on how to configure it.
+
+<details>
+  <summary>Expand example</summary>
+
+  ```toml path="bunfig.toml"
+  preload = ["./bunMdxEsbuild.ts"]
+  ```
+
+  ```js twoslash path="bunMdxEsbuild.ts"
+  import {plugin} from 'bun'
+  import mdx from '@mdx-js/esbuild'
+
+  plugin(mdx())
+  ```
+</details>
 
 ## Further reading
 
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -234,22 +234,9 @@ etc.) you use.
 To use more modern JavaScript features than what your users support,
 [configure esbuild’s `target`][esbuild-target].
 
-`@mdx-js/esbuild` can also be used with [Bun][]:
-
-<details>
-  <summary>Expand example</summary>
-
-  ```toml path="bunfig.toml"
-  preload = ["./bunMdxEsbuild.ts"]
-  ```
-
-  ```js twoslash path="bunMdxEsbuild.ts"
-  import {plugin} from 'bun'
-  import mdx from '@mdx-js/esbuild'
-
-  plugin(mdx())
-  ```
-</details>
+See also [¶ Bun][javascript-engines-bun],
+which you might be using,
+for more info.
 
 #### Rollup
 
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -935,7 +935,9 @@ MDX files can be imported in [Bun][] by using
 
 [build-system-vite]: #vite
 
-[bun]: https://bun.sh/
+[bun]: https://bun.sh
+
+[javascript-engines-bun]: #bun
 
 [bundler-esbuild]: #esbuild
 
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -811,10 +811,10 @@ MDX files can be imported in [Bun][] by using
   <summary>Expand example</summary>
 
   ```toml path="bunfig.toml"
-  preload = ["./bunMdxEsbuild.ts"]
+  preload = ["./bun-mdx.ts"]
   ```
 
-  ```js twoslash path="bunMdxEsbuild.ts"
+  ```js twoslash path="bun-mdx.ts"
   import {plugin} from 'bun'
   import mdx from '@mdx-js/esbuild'
 
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -234,6 +234,23 @@ etc.) you use.
 To use more modern JavaScript features than what your users support,
 [configure esbuild’s `target`][esbuild-target].
 
+`@mdx-js/esbuild` can also be used with [Bun][]:
+
+<details>
+  <summary>Expand example</summary>
+
+  ```toml path="bunfig.toml"
+  preload = ["./bunMdxEsbuild.ts"]
+  ```
+
+  ```js twoslash path="bunMdxEsbuild.ts"
+  import {plugin} from 'bun'
+  import mdx from '@mdx-js/esbuild'
+
+  plugin(mdx())
+  ```
+</details>
+
 #### Rollup
 
 <details>
@@ -910,6 +927,8 @@ See its readme on how to configure it.
 [api-processor-options]: /packages/mdx/#processoroptions
 
 [build-system-vite]: #vite
+
+[bun]: https://bun.sh/
 
 [bundler-esbuild]: #esbuild
 
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -62,7 +62,7 @@ MDX is a language that’s compiled to JavaScript.
 The easiest way to get started is to use an integration for your bundler if you
 have one:
 
-* if you use **esbuild**,
+* if you use **esbuild** (or Bun),
   install and configure [`@mdx-js/esbuild`][mdx-esbuild]
 * if you use **Rollup** (or Vite),
   install and configure [`@mdx-js/rollup`][mdx-rollup]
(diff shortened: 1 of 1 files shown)


remcohaszing (approved):

wooorm:
1. Oh cool that bun already mentions it!
2. I don’t think the esbuild plugin is currently used too much, I’d assume that bun would increase that usage; that could uncover some bugs/feature requests

ChristianMurphy (approved):

wooorm:
3. I do not understand what this has to do with eval? Scripts?

wooorm:
4. This seems similar to how bite compares to rollup, and next compares to webpack. See https://mdxjs.com/docs/getting-started/#bundler. I think you could change that to "esbuild (or Bun)" too
5. Also, in the same section under that heading but in the 2nd list, there's a mention of Node, you can also add a bullet after it talking about Bun. Either this point or 4. Probably, not both.
6. This should likely be a section under https://mdxjs.com/docs/getting-started/#javascript-engines?

wooorm:
Reading your alternatives section again, have you considered my point 6? It solves both the problems you highlight with the alternatives.
The main reason I wonder is: do bun users look for esbuild plugins? Or do they look for Bun?

karlhorky:
> 3\. I do not understand what this has to do with eval? Scripts?

Our use case was to run a Node.js script which imports `.mdx` files for their exports

We wanted to continue using `tsx script.ts` for this, but [`tsx` (and also the esbuild Transform API which `tsx` also relies upon) doesn't support esbuild plugins](https://github.com/privatenumber/tsx/issues/491#issuecomment-1987041573)

So instead of taking the long way around by first using the `esbuild` Build API and then executing the build output (or trying to do it in memory with [`write: false`](https://github.com/evanw/esbuild/issues/139)), we tried it with Bun, which worked pretty much out of the box

karlhorky:
> 4. This seems similar to how bite compares to rollup, and next compares to webpack. See [mdxjs.com/docs/getting-started](https://mdxjs.com/docs/getting-started/#bundler). I think you could change that to "esbuild (or Bun)" too
> 5. Also, in the same section under that heading but in the 2nd list, there's a mention of Node, you can also add a bullet after it talking about Bun. Either this point or 4. Probably, not both.
> 6. This should likely be a section under [mdxjs.com/docs/getting-started](https://mdxjs.com/docs/getting-started/#javascript-engines)?


Changes added for 4 and 6 in f188aa9eb0f462771c1b44f5c489847569485426

wooorm:
> script

I was confused by this word because it is often used to differentiate between scripts vs modules. Scripts in browsers not having a way to import/export things. Or in Node, scripts using CJS.

> eval

Similarly, `eval` is a javascript function. And MDX has an API for it ([`evaluate`](https://mdxjs.com/packages/mdx/#evaluatefile-options)).

> Our use case was to run a Node.js script which imports .mdx files for their exports (specifically JS exported metadata)

The document you are editing contains a section on Node. We have a specific package for Node ([`@mdx-js/node-loader`](https://mdxjs.com/packages/node-loader/)). Did you consider using it?

docs/docs/getting-started.mdx
```diff
@@ -911,6 +934,8 @@ See its readme on how to configure it.
 
 [build-system-vite]: #vite
 
+[bun]: https://bun.sh/
```
wooorm, line 937:
```suggestion
[bun]: https://bun.sh

[javascript-engines-bun]: #bun
```

- slash unneeded, rest of docs don’t use it either
- also adding link for bun section

docs/docs/getting-started.mdx
```diff
@@ -234,6 +234,23 @@ etc.) you use.
...
+  ```
+
+  ```js twoslash path="bunMdxEsbuild.ts"
+  import {plugin} from 'bun'
+  import mdx from '@mdx-js/esbuild'
+
+  plugin(mdx())
+  ```
+</details>
+
```
wooorm, line 253: I think it makes more sense to have the bun example in the bun section, and for consistency with the rest of the document, refer to it like in other places:

```suggestion
See also [¶ Bun][javascript-engines-bun],
which you might be using,
for more info.

```

docs/docs/getting-started.mdx
```diff
@@ -798,6 +815,12 @@ MDX files can be imported in Node by using
 [`@mdx-js/node-loader`][mdx-node-loader].
 See its readme on how to configure it.
 
+#### Bun
+
+MDX files can be imported in Bun by using
+[`@mdx-js/esbuild`][mdx-esbuild].
+See [the section above][bundler-esbuild] on how to configure it.
```
wooorm, line 822:
````suggestion
MDX files can be imported in [Bun][] by using
[`@mdx-js/esbuild`][mdx-esbuild].

<details>
  <summary>Expand example</summary>

  ```toml path="bunfig.toml"
  preload = ["./bunMdxEsbuild.ts"]
  ```

  ```js twoslash path="bunMdxEsbuild.ts"
  import {plugin} from 'bun'
  import mdx from '@mdx-js/esbuild'

  plugin(mdx())
  ```
</details>
````

**Review by wooorm — COMMENTED** [review]
I think it makes more sense to have the bun example in the bun section

**Comment by karlhorky** [comment]
> The document you are editing contains a section on Node. We have a specific package for Node ([`@mdx-js/node-loader`](https://mdxjs.com/packages/node-loader/)). Did you consider using it?

No, actually I was not aware of this package until now. But since we were using `tsx`, my entry point was "how can I bootstrap `tsx` / `esbuild` to understand `.mdx`?"

**Comment by wooorm** [comment]
> But since we were using tsx, my entry point was "how can I bootstrap tsx / esbuild to understand .mdx?"

Am I correct in understanding that currently you replaced Node/TSX with bun?

So, the earlier problem you had was how to use MDX with TSX but that was impossible?

Are you still interested in using MDX with TSX? Or is Bun enough?

**Comment by karlhorky** [comment]
> Am I correct in understanding that currently you replaced Node/TSX with bun?
> 
> So, the earlier problem you had was how to use MDX with TSX but that was impossible?

Yes, from [the reply in the tsx issue](https://github.com/privatenumber/tsx/issues/491#issuecomment-1987041573), I concluded that it is not possible to use `@mdx-js/esbuild` as an esbuild plugin with tsx.

For us, switching from tsx to Bun has no downsides and gives us a perf boost ([we've done this for other scripts using tsx before too](https://github.com/evanw/esbuild/issues/1492#issuecomment-2061825201) - for `import.meta.dirname` and `import.meta.filename` support)

> Are you still interested in using MDX with TSX? Or is Bun enough?

I'm interested in the option for other users though! It could potentially be an important thing for tsx users to have. Maybe with the Node.js loader it would be possible somehow?

**Comment by wooorm** [comment]
> for import.meta.dirname and import.meta.filename support

(Personally I don’t see the need for them, I believe URLs with `import.meta.url` is a very welcome improvement over paths, which have windows/unix compat differences, the possibility of paths being relative is also a downside, and URLs being specced by WHATWG is nice)

> Maybe with the Node.js loader it would be possible somehow?

TSX is a Node loader and we have a Node loader too. You can use two loaders together

File: docs/docs/getting-started.mdx [hunk 4]
```diff
@@ -798,6 +802,26 @@ MDX files can be imported in Node by using
...
+[`@mdx-js/esbuild`][mdx-esbuild].
+
+<details>
+  <summary>Expand example</summary>
+
+  ```toml path="bunfig.toml"
+  preload = ["./bunMdxEsbuild.ts"]
+  ```
+
+  ```js twoslash path="bunMdxEsbuild.ts"
```
wooorm, line 817: One thing I wonder about is this casing in the filename? Is this a very strong recommendation by bun, to use camelcasing? Seems like that would fail on some file systems?

Should the name include esbuild, or is that a side effect of what the focus is: Bun + MDX?

How about a short:

````suggestion
  preload = ["./mdx.ts"]
  ```

  ```js twoslash path="mdx.ts"
````

wooorm: `bun-preload.ts` also seems viable?

karlhorky: > Is this a very strong recommendation by bun, to use camelcasing?

It was my own choice. Camel case in filenames are pretty common in JavaScript and TypeScript ecosystems.

I think `mdx.ts` is not expressive enough. If it should be lowercase (kebab case?) then at least `bun-mdx.ts` or `bun-mdx-plugin.ts`

wooorm: OK, I’m very :-1: on camelcase in filenames, I only indeed sometimes see React people use it more recently. It’s not used anywhere in the things I work on.

Did you see my 2nd suggestion, `bun-preload.ts`?

My acceptable preferences in order are: `bun-preload.ts`, `mdx.ts`, `bun.ts`, `bun-mdx.ts`. All fine. They all answer “wait, why is this file here again?”.

I think the inclusion of “plugin” in `bun-mdx-plugin.ts` is unneeded. Similar to “esbuild” `bunMdxEsbuild.ts`. These terms have to do with how that file currently works internally.

karlhorky: Ok, switched to `bun-mdx.ts` in 3de41ee65c2527ae1f91aaf0c220ce97cc6f0fdf

wooorm (approved):

wooorm:
Thank you!

karlhorky:
Glad to help, thanks for the review and merge!

karlhorky:
cc @Jarred-Sumner Bun in MDX docs 😍

PR #2518, opened 2024-07-23
Fix types, lints in example

karlhorky:
### Initial checklist

### Description of changes

Small follow-up to #2517 to address the type checking and linting problems with the original code from the Bun docs

```
Argument of type 'Plugin' is not assignable to parameter of type 'BunPlugin'.
  Types of property 'setup' are incompatible.
    Type '(build: PluginBuild) => void | Promise<void>' is not assignable to type '(build: PluginBuilder) => void | Promise<void>'.
      Types of parameters 'build' and 'build' are incompatible.
        Type 'PluginBuilder' is missing the following properties from type 'PluginBuild': initialOptions, resolve, onStart, onEnd, and 2 more. ts (2345)
```

```
Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler. eslint (@typescript-eslint/no-floating-promises)
```

### DIFF
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -816,7 +816,7 @@ MDX files can be imported in [Bun][] by using
 
   ```js twoslash path="bun-mdx.ts"
   import mdx from '@mdx-js/esbuild'
-  import { BunPlugin, plugin } from 'bun'
+  import {type BunPlugin, plugin} from 'bun'
 
   await plugin(mdx() as unknown as BunPlugin)
   ```
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -815,10 +815,10 @@ MDX files can be imported in [Bun][] by using
   ```
 
   ```js twoslash path="bun-mdx.ts"
-  import {plugin} from 'bun'
-  import mdx from '@mdx-js/esbuild'
+  import mdx from '@mdx-js/esbuild';
+  import { BunPlugin, plugin } from 'bun';
 
-  plugin(mdx())
+  await plugin(mdx() as unknown as BunPlugin);
   ```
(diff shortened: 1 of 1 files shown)


docs/docs/getting-started.mdx
```diff
@@ -815,10 +815,10 @@ MDX files can be imported in [Bun][] by using
   ```
 
   ```js twoslash path="bun-mdx.ts"
-  import {plugin} from 'bun'
   import mdx from '@mdx-js/esbuild'
+  import { BunPlugin, plugin } from 'bun'
```
wooorm, line 819:
```suggestion
  import {type BunPlugin, plugin} from 'bun'
```

- match style of codebase
- it’s probably a type?

wooorm (approved):

wooorm:
thanks!

wooorm:
Can you try and build the website next time btw? That actually type checks. And finds other problems, such as that toml wasn’t registered, and that your example uses typescript instead of javascript! https://github.com/mdx-js/mdx/commit/05ecf65f1943e6810b21e40e19354210aa9abefb

karlhorky:
Ah, sorry for the multiple mistakes there 😬  I am actually working on GitHub only, I don't even have this cloned - just relying on GitHub Actions CI and my memory / reviewing to do the right thing.

Would it make sense to add a "smoke test" website build to the GitHub Actions CI?

For my own PRs for now, I'll keep in mind that things can break without being reported by CI.

wooorm:
It's intentional that things don't break, so newcomers can contribute, even when they're not as neat in their code as the maintainers! Or, a type error can exist for a bit in an example, while the website remains maintainable.
CI did probably show this in the logs btw; if you want to remain in GH but benefit from it, you can go through the logs.

PR #2523, opened 2024-08-06
docs: update Parcel MDX plugin for MDX 3

TechQuery:
### Initial checklist

### Description of changes

Parcel official MDX plugin is only support MDX 1, so I publish [a new plugin for MDX 3](https://github.com/EasyWebApp/Parcel-transformer-MDX).

### DIFF
--- a/docs/docs/getting-started.mdx
+++ b/docs/docs/getting-started.mdx
@@ -617,8 +617,8 @@ See [Configuring MDX on `nextjs.org`][next-configuring-mdx] for more info.
 
 #### Parcel
 
-[Parcel][] has its own plugin to support MDX.
-See [`@parcel/transformer-mdx`][parcel-mdx]
+[Parcel][] has its community plugin to support MDX 3.
+See [`parcel-transformer-mdx`][parcel-mdx]
 on how to use MDX with Parcel.
 
 ### JSX runtimes
@@ -877,7 +877,7 @@ MDX files can be imported in [Bun][] by using
 
 [parcel]: https://parceljs.org
 
-[parcel-mdx]: https://parceljs.org/languages/mdx/
+[parcel-mdx]: https://github.com/EasyWebApp/Parcel-transformer-MDX
 
 [rollup]: https://rollupjs.org
 


ChristianMurphy:
Thanks for sharing @TechQuery!

General feedback on what gets included in the plugins/guides list:

> Some criteria to include packages in this list.
>
> - Some documentation with at least some instructions on how to use the package.
> - A CI job to run tests. (The tests are already there.)
> - A `prepack` script and CI job to build the types. (The JSDoc types are already there.)
> - The package should be published to npm.

source: https://github.com/syntax-tree/hast/pull/24#issuecomment-2015384111

It looks like there aren't any types or tests included in the repository.
In addition:
* The repository does not include a LICENSE file, it should have a license. I'd recommend MIT.
* It looks like the repository uses a mix of `npm` and `pnpm` for dependencies and scripts. I'd recommend using only one. I'd recommend `npm`.
* The plugin hard codes some remark and rehype plugins, I'd recommend starting with vanilla MDX out of the box https://github.com/EasyWebApp/Parcel-transformer-MDX/blob/6380b072de8016c060f721afa3f5461f33c6b162/index.js#L53-L59
* The plugin does not allow users to specify their own remark and rehype plugins to include, I'd recommend adding this
* You are publishing from GHA, I'd recommend adding package provenance https://github.blog/changelog/2023-09-26-npm-provenance-general-availability/
* Parcel as of version 2 defaults to the faster SWC transpiler https://parceljs.org/blog/v2/, it looks like this forces MDX through the slower and non-default Babel pipeline https://github.com/EasyWebApp/Parcel-transformer-MDX/blob/6380b072de8016c060f721afa3f5461f33c6b162/index.js#L25-L29

TechQuery:
Thanks for quick reviewing @ChristianMurphy !
 
> It looks like there aren't any types or tests included in the repository.

This is my first Parcel plugin, I'm learning the best practice of developing Parcel plugins with ESM & TS:
- I use JSDoc & TS to [check JS codes internally][1], but users won't import it in their codes, so `.d.ts` isn't required.
- I plan to EasyWebApp/Parcel-transformer-MDX#1 to make test scripts easier.

> * The repository does not include a LICENSE file, it should have a license. I'd recommend MIT.

I added `package.json#license` field, and forgot to add a `LICENSE` file. (As an individual coder, I prefer GPLs.)

> * It looks like the repository uses a mix of `npm` and `pnpm` for dependencies and scripts. I'd recommend using only one. I'd recommend `npm`.

This package is designed for NPM compatibility, andI use PNPM for developing, so it's no mix-using for users.

> * The plugin hard codes some remark and rehype plugins, I'd recommend starting with vanilla MDX out of the box https://github.com/EasyWebApp/Parcel-transformer-MDX/blob/6380b072de8016c060f721afa3f5461f33c6b162/index.js#L53-L59

This plugin has the same goal of Parcel - Zero Configuration, so users can write Markdown files as what they do on GitHub immediately.

> * The plugin does not allow users to specify their own remark and rehype plugins to include, I'd recommend adding this

Because MDX 3 has no official configuration file, so I'm thinking about the configuration schema of my own plugin. I'm glad to have issue or PR submitted for this feature, and hoping a simple design to keep Parcel easy.

> * You are publishing from GHA, I'd recommend adding package provenance https://github.blog/changelog/2023-09-26-npm-provenance-general-availability/

Thanks for your suggestion! I noticed this update of NPM web-site, and was too busy to learn how to use it.

> * Parcel as of version 2 defaults to the faster SWC transpiler https://parceljs.org/blog/v2/, it looks like this forces MDX through the slower and non-default Babel pipeline https://github.com/EasyWebApp/Parcel-transformer-MDX/blob/6380b072de8016c060f721afa3f5461f33c6b162/index.js#L25-L29

These lines just aim to find JSX configuration from all kinds of possible files without extra configuration , which recommends TypeScript and is compatible with Babel, this plugin doesn't control the compiler.

[1]: https://github.com/EasyWebApp/Parcel-transformer-MDX/blob/6380b072de8016c060f721afa3f5461f33c6b162/index.js#L32-L36

ChristianMurphy:
> As an individual coder, I prefer GPLs.

I appreciate that, I don't think unified has a strict policy on license type for adoption (cc @wooorm @remcohaszing)
But for wider adoption, I generally recommend licenses that are compatible with Apache 2.0 https://www.apache.org/legal/resolved.html
The GPL and LGPL licenses exclude a lot of projects and organizations from adopting.

> This package is designed for NPM compatibility, and I use PNPM for developing, so it's no mix-using for users

I'd still recommend building, testing, and publishing with one.
They layout packages different and have different resolutions for peer and transitive dependencies.
Mixing the two during development could lead to unexpected errors.

> Because MDX 3 has no official configuration file, so I'm thinking about the configuration schema of my own plugin. I'm glad to have issue or PR submitted for this feature, and hoping a simple design to keep Parcel easy.

The plugins are remark and rehype based, you could leverage those files, similar to how eslint-mdx does
https://github.com/mdx-js/eslint-mdx?tab=readme-ov-file#mdxremark

> I use JSDoc & TS to [check JS codes internally](https://github.com/EasyWebApp/Parcel-transformer-MDX/blob/6380b072de8016c060f721afa3f5461f33c6b162/index.js#L32-L36), but users won't import it in their codes, so .d.ts isn't required.

Even if they aren't published I'd recommend running `tsc` to ensure they are valid.

> I plan to https://github.com/EasyWebApp/Parcel-transformer-MDX/issues/1 to make test scripts easier.

Sounds good

TechQuery:
> The GPL and LGPL licenses exclude a lot of projects and organizations from adopting.

In my opinion, LGPL is designed for libraries, why people keep worrying about it...

> I'd still recommend building, testing, and publishing with one.
> They layout packages different and have different resolutions for peer and transitive dependencies.
> Mixing the two during development could lead to unexpected errors.

PNPM has more strict Dependency Checking, all packages worked in PNPM also work well in NPM & Yarn.

In the other hand, NPM & Yarn have so bad performance, PNPM is saving my life. But I use it only for local Dependency Management & NPM scripts shorthands, all other things are compatible with NPM.

> Even if they aren't published I'd recommend running `tsc` to ensure they are valid.

When I resolve EasyWebApp/Parcel-transformer-MDX#1, I will rewrite the whole project with TypeScript to replace the quick & dirty version (which is made as the reason of long time waiting for Parcel MDX plugin updating.

TechQuery:
> Re license: I would very very strongly recommend going with MIT/ISC/similar. That’s what almost all packages in the JS ecosystem go with. Parcel too. And us. Only if you’d have a ton of knowledge about this, and thought this through, would I recommend going against 99% of the npm packages.

In my mind, GPLs are for powerless coders, and others are for powerful companies, universities or foundations. 

Until now, no one of open sources arounded give me a clear reason about this, just "most people use it, so you should use it"... But why? No one knows the logic in it, it doesn't make sense. I will do anything for what makes me believe.

> I am rather bummed that Parcel hasn’t solved this after a couple years. This reflects poorly on the state of Parcel.
> Still, this seems like a temporary solution for now, something that will be fixed in the future.

Yeah, as a long term user, I'm worrying about the team status of Parcel project. But after using webpack, Snowpack, MicroBundle & Vite, only Parcel make me,    teammates and students happy on coding with **Zero Configuration** design.

> Have you considered keeping the original text and link in place, but adding a note linking to your plugin for now, such as this [note for Vite](https://mdxjs.com/docs/getting-started/#vite)?

I thought about that, but it'll be confused for users to find a plugin only support MDX 1.x in MDX 3.x document.

ChristianMurphy:
> In my mind, GPLs are for powerless individual coders, and others are for powerful companies, universities or foundations.

Most software is from and for larger organizations and projects.
If your intent is for this to be a toy project for individual developers.
That is fine, but it doesn't need to be in the main documentation.

> Until now, no one of open sourcers arounded gives me a clear reason about this, just keeps saying: "Most people use it, so you should use it"... But why? No one knows the logic in it, it doesn't make sense. I will do anything for what makes me believe.

Concisely, Category X licenses, like LGPL, limit how software can be distributed https://www.apache.org/legal/resolved.html#prohibited
The terms get a bit fuzzier when used as a build tool, but it's pretty common for projects to bundle/build on top of build tools, and those would run into the distribution restrictions.

> Zero Configuration design

Presets can be created for other build tools and bundlers as well.

> confused for users to find a plugin only support MDX 1.x in MDX 3.x document.

I think they'll be more confused if they go out searching for a parcel plugin, see the official one, and don't see it listed in the MDX docs anywhere.

TechQuery:
> Most software is from and for larger organizations and projects.
> If your intent is for this to be a toy project for individual developers.
> That is fine, but it doesn't need to be in the main documentation.

This plugin is came from Individual coders, for but only for individuals.

And most great Open Source projects were _Toy projects_ at the beginning, such as Linux with GPL v2.

> Concisely, Category X licenses, like LGPL, limit how software can be distributed https://www.apache.org/legal/resolved.html#prohibited
> The terms get a bit fuzzier when used as a build tool, but it's pretty common for projects to bundle/build on top of build tools, and those would run into the distribution restrictions.

Ok, I will do more research.

> Presets can be created for other build tools and bundlers as well.

As I know, only Babel has presets, and other bundlers only have a bunch of configuration fields in their documents, which needs users to set them up manually.

> I think they'll be more confused if they go out searching for a parcel plugin, see the official one, and don't see it listed in the MDX docs anywhere.

This plugin is a result of my failed searching about MDX 3.x Parcel plugin, the Parcel official document will tell them that they only support MDX 1.x.

ChristianMurphy:
> most great Open Source projects were Toy projects at the beginning, such as Linux with GPL v2.

Sure, I agree.
My point is that the _MDX docs_ focuses on projects that have moved above toy project level.

> This plugin is [...] only for individuals.

This concerns me a bit.
Many of MDX's adopters are individuals working in/on-behalf of a larger organization.
I'm not saying you have to offer corporate enterprise support.
But excluding all organization based developers feels a step too far to me.

> As I know, only Babel has presets, and other bundlers only have a bunch of configuration fields in their documents, which needs users to set them up manually.

Any build tool can be wrapped to templated with defaults you like.
For example https://github.com/antfu-collective/vitesse or https://github.com/preactjs/preset-vite

wooorm:
> In my mind, GPLs are for powerless individual coders, and others are for powerful companies, universities or foundations.

I don’t think this understanding is correct; npm being 99% MIT shows that “powerless individual coders” are also using MIT.

> But why? No one knows the logic in it, it doesn't make sense. I will do anything for what makes me believe.

You can ask? :wink: What do you want to know?

> I thought about that, but it'll be confused for users to find a plugin only support MDX 1.x in MDX 3.x document.

I don’t think I see that? I don’t think so? Why would someone be confused with a note? You can show 2 things in that section?

I’d lean towards adding a note, as it’s something for now, and Parcel will solve it later?

TechQuery:
> > This plugin is [...] only for individuals.
> 
> This concerns me a bit.

I'm sorry, my Speak-to-Text Input Method lost the _not_, I mean "Not only for individuals"😂

> Any build tool can be wrapped to templated with defaults you like.
> For example https://github.com/antfu-collective/vitesse or https://github.com/preactjs/preset-vite

If a framework or bundler needs a bunch of scaffold codes to make it good to use, it's not good enough actually.

Zero Configuration design is beyond the presets, because the Web standards are the best presets. Parcel bundles projects based on How we refer assets (build dependency graph with `href`, `src`, `url()`, `import`, etc.) and What we write (auto install compilers and official plugins with `.html`, `.css`, `.less`, `.js`, `.ts`, `.tsx`, etc.).

If users have some Advanced requirements, Parcel's plugin system loads configurations of all the existed tool chains with the original ways (`.babelrc`, `tsconfig.json`, etc.), doesn't like others, such as: Vite's React & Vue plugins have totally different ways to set up Babel configuration with barely no detail document, it makes users to feel exhausted in the migration between frameworks and bundlers...

The web is simple before, and powerful but complex today. What Parcel does is to make the Web easier again.

wooorm:
> Zero Configuration [...]

There is a place for zero config stuff!

There is also a place for tools, the stuff we make here in unified, that make *all the other tools possible*. Some of those tools have different needs!

MDX, remark, unified, **are configurable**. And this is important. If you call your things MDX or unified, they must be configurable as well. If you want a zero-config tool, on top of MDX, you can make that, but it’s not MDX. I think you need a new name.

> The Web is simple before, and powerful but complex today. What Parcel does is to make the Web easier again.

I personally don’t think the web has ever been simple. The web is about humans. Not even just humans: human--computer interaction, computer--computer interaction, human--human interaction. That’s always going to be incredibly complex.

remcohaszing:
Re license: I recommend using MIT. It’s free, common, and simple. But you’re free to use whatever license you want. You can always relicense your own projects (but the old license applies to old versions). I don’t think we should block any (L)GPL licensed projects from being listed on the MDX website.

I agree with @wooorm that an MDX integration for a build tool should support MDX options. The current implementation picks an arbritrary list of plugins. Some of these are common, other’s I haven’t even seen before. I think that shows there’s the need for the user to specify their own plugin list.

wooorm:
> I agree with @wooorm that an MDX integration for a build tool should support MDX options. The current implementation picks an arbritrary list of plugins. Some of these are common, other’s I haven’t even seen before. I think that shows there’s the need for the user to specify their own plugin list.

Perhaps we should instead document to Parcel users how they can make their own plugin?

I think the core code is:

```js
import {Transformer} from '@parcel/plugin'

const transform = new Transformer({
  async transform({asset}) {
    const source = await asset.getCode()
    const file = await compile(source)
    asset.type = 'jsx' // Question: this is JS already, should it be `js`?
    asset.setCode(String(file))
    return [asset]
  }
})

export default transform
```

remcohaszing:
I haven’t used Parcel before, let alone dealt with plugins. I also haven’t looked in depth too much. Don’t they support any configuration options?

> Perhaps we should instead document to Parcel users how they can make their own plugin?

Or we could to it like `@mdx-js/node-loader`: expose a function to make it simple to create your own, and expose one with default configuration.

> I think the core code is:
> 
> ```js
> import {Transformer} from '@parcel/plugin'
> import {compile} from '@mdx-js/mdx'
> 
> const transform = new Transformer({
>   async transform({asset}) {
>     const source = await asset.getCode()
>     const file = await compile(source)
>     asset.type = 'jsx' // Question: this is JS already, should it be `js`?
>     asset.setCode(String(file))
>     return [asset]
>   }
> })
> 
> export default transform
> ```

We should pass the file path to `compile()` if we have access to it. The asset type should indeed match the output type. This means `jsx` if the `{jsx: true}` option is passed to `compile()` or `js` otherwise. I would also be surprised if Parcel doesn’t have source map support.

TechQuery:
@ChristianMurphy 

> Parcel and CRA are not good at configuration/customization.

Parcel's plugin system is different than others, but easy to configure & customize.

In the other hand, CRA's configuration/customization is suck.

> > Parcel bundles projects based on How we refer assets (build dependency graph with href, src, url(), import, etc.) and What we write (auto install compilers and official plugins with .html, .css, .less, .js, .ts, .tsx, etc.).
> 
> Right, and so does _every other_ bundler and build tool from the past 2-5 years 🙂 ESBuild, Vite, Rspack, etc all focus on aligning to web standards first and foremost, and have lots of official plugins.

webpack needs `<% something %>` to inject JS files, others load images, videos, etc by absolute paths from `public/`, which can't be hinted by your IDE.

But Parcel uses relative paths hinted by IDE, and auto install all compilers & official plugins, which others do not.

ChristianMurphy:
> In the other hand, CRA's configuration/customization is [bad].
> webpack needs <% something %> to inject JS files, others load images, videos, etc by absolute paths from public/, which can't be hinted by your IDE.

@TechQuery I agree, you're going after a straw-man, I'm not talking about webpack. 🙂 

> Parcel team has few action on MDX upgrading, "later" is undefined for years

Your code could also shared as a PR to Parcel itself to upgrade the official MDX integration, right?
And if it were updated in the official repo, documentation across GitHub and the internet wouldn't need to be updated.

TechQuery:
> > webpack needs <% something %> to inject JS files, others load images, videos, etc by absolute paths from public/, which can't be hinted by your IDE.
> 
> @TechQuery I agree, you're going after a straw-man, I'm not talking about webpack. 🙂

1. webpack is also active developing, not too old.
2. "others" means Vite, etc.

For now, No one latest bundlers implements all Zero Configuration features of Parcel, such as relative paths & auto installation.

> Your code could also shared as a PR to Parcel itself to upgrade the official MDX integration, right? And if it were updated in the official repo, documentation across GitHub and the internet wouldn't need to be updated.

I did, look up for the PR link.

TechQuery:
> The current implementation picks an arbritrary list of plugins. Some of these are common, other’s I haven’t even seen before.

@remcohaszing All preset MDX plugins come from official "Guides" document: https://mdxjs.com/guides/

TechQuery:
> Perhaps we should instead document to Parcel users how they can make their own plugin?
> 
> I think the core code is:
> 
> ```js
> import {Transformer} from '@parcel/plugin'
> import {compile} from '@mdx-js/mdx'
> 
> const transform = new Transformer({
>   async transform({asset}) {
>     const source = await asset.getCode()
>     const file = await compile(source)
>     asset.type = 'jsx' // Question: this is JS already, should it be `js`?
>     asset.setCode(String(file))
>     return [asset]
>   }
> })
> 
> export default transform
> ```

@wooorm My plugin came from https://github.com/parcel-bundler/parcel/pull/7922#issuecomment-1750704973, but we shouldn't make Parcel users to write it again & again.

ChristianMurphy:
> which should be common for most users.
> we shouldn't make Parcel users to write it again & again.

Users have a really wide variety of use cases for MDX.
Vanilla MDX is the most common use case.

If you want to avoid users having to write a custom code for when the do want a plugin.
Support the remark and rehype configuration files so users can add the plugins they want https://github.com/mdx-js/mdx/pull/2523#issuecomment-2272252267

TechQuery:
> If you want to avoid users having to write a custom code for when the do want a plugin.
> Support the remark and rehype configuration files so users can add the plugins they want [#2523 (comment)](https://github.com/mdx-js/mdx/pull/2523#issuecomment-2272252267)

@ChristianMurphy @wooorm @remcohaszing Custom Configuration support is released: https://github.com/EasyWebApp/Parcel-transformer-MDX/releases/tag/v0.4.0

TechQuery:
> Or we could to it like @mdx-js/node-loader: expose a function to make it simple to create your own, and expose one with default configuration.

@remcohaszing It would be nice if someone upgrades MDX Node.js loader to read configuration from a `mdx.config.js` file, then EasyWebApp/Parcel-transformer-MDX#1 may be no need to do.

wooorm:
> It would be nice if someone upgrades MDX Node.js loader to read configuration from a mdx.config.js file, then https://github.com/EasyWebApp/Parcel-transformer-MDX/issues/1 may be no need to do.

? I do not understand this issue you created on your repo?

I am not so sure I think `mdx.config.js` is a good idea?
There is no `html.config.js` either?

wooorm:
Thanks for your patience. I just landed this. Reworded it a bit tho.

PR #2593, opened 2025-02-21
esbuild plugin: use esbuild "jsx" loader if output is JSX

egnor:
### Initial checklist

* [x] I read the support docs 

### Description of changes

In the esbuild plugin, if the MDX pipeline is configured to output unprocessed JSX (via `jsx: true` in its options), have esbuild expect JSX (via `loader: "jsx"` instead of the default `"js"`). This allows users to freely choose whether the JSX-to-JS processing happens in the MDX pipeline or in esbuild (both are capable of doing it).

### DIFF
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -142,7 +142,7 @@ export function esbuild(options) {
       return {
         contents: value || '',
         errors,
-        loader: options.jsx ? 'jsx' : 'js',
+        loader: options?.jsx ? 'jsx' : 'js',
         resolveDir: path.resolve(file.cwd, file.dirname),
         warnings
       }
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -142,6 +142,7 @@ export function esbuild(options) {
       return {
         contents: value || '',
         errors,
+        loader: options.jsx ? 'jsx' : 'js',
         resolveDir: path.resolve(file.cwd, file.dirname),
         warnings
       }
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -66,10 +66,8 @@ const name = '@mdx-js/esbuild'
  *   Plugin.
  */
 export function esbuild(options) {
-  const {extnames, process} = createFormatAwareProcessors({
-    ...options,
-    SourceMapGenerator
-  })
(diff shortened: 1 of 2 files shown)


wooorm:
why do you want to use esbuild’s jsx loader?

egnor:
(updated with test & fix)

egnor:
> why do you want to use esbuild’s jsx loader?

somehow I knew you'd ask that!

pragmatically
- keeping JSX-to-JS translation in one place, the options can be a little subtle given the many flavors of JSX transform and various "automatic" transformers kicking around
- there may be subsequent / nonstandard things people want to do globally with JSX in their build system, that should apply to the JSX from MDX but also the JSX from (say) React components or whatnot
- in my particular specific case, I want my recma plugins to see untransformed JSX, that's a rather idiosyncratic case and perhaps best ignored, but `jsx: true` does exist and I felt like it ought to work

philosophically
- this feels like a basic correctness fix. esbuild requires onLoad callbacks to specify the loader (effectively content type), it defaults to `loader: "js"` but it's up to plugins to describe the output they're producing. in our case we know whether it's `"js"` or `"jsx"` and should set that value correctly I'd think
- it's a super simple change, easily tested, correct by inspection, unlikely to add maintenance burden, and removes surprise

packages/esbuild/lib/index.js
```diff
@@ -142,6 +142,7 @@ export function esbuild(options) {
       return {
         contents: value || '',
         errors,
+        loader: options?.jsx ? 'jsx' : 'js',
```
wooorm, line 145: I think it’s good to put a `const settings = {...options, SourceMapGenerator}` on L69, *use* that `settings` in what is now on L69, and then also use `settings.jsx` here. Instead of this optional chaining

wooorm:
It is probably good to add the same feature to other loaders. Though perhaps not in this PR.

I do think it’s better to not use this. That much of your troubles come from integrating so closely with JSX instead of the other ASTs.
But as you point out, this option exists.

packages/esbuild/lib/index.js
egnor: Done!

wooorm (approved):

wooorm:
Released in [`3.1.1`](https://github.com/mdx-js/mdx/releases/tag/3.1.1)! Thanks for your patience!

PR #2595, opened 2025-03-03
Error handling improvements to esbuild plugin

egnor:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This change improves(?) error conversion in the esbuild plugin, with no change to non-error function.

Previously, if a non-`VFileMessage` error was thrown by the MDX conversion process (eg. a simple nodejs runtime Error thrown by a plugin or the core), it would be wrapped as the `cause` in a `VFileMessage` with generic "Cannot process MDX file..." text, and re-thrown by the `onload` handler. This makes sense! However, when that message is converted in `vfileMessageToEsbuild`, the `cause` was completely ignored, so all information about that underlying error was lost, which makes debugging harder (especially debugging plugins).

Also previously, even when the original error thrown was a `VFileMessage`, if the optional `place.start.offset` (or `place.offset`) isn't defined, all place information was lost. This `.offset` field is optional (though set in most common paths).

This change revises things:

- In the `onload` handler, _always_ throw a new `VFileMessage` with the original exception attached. Hoist location information to the outer exception, if the inner exception appears to be a `VFileMessage` itself. Include the inner exception's text rendering in the outer's text (after a colon).

- In `vfileMessageToEsbuild`, handle the case where offset information isn't present, working with whatever information is present.

The changes to the unit test are a decent way to see the effect of all this.

### DIFF
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -117,12 +117,11 @@ export function esbuild(options) {
           '\n'
         messages = file.messages
       } catch (error_) {
-        const cause = /** @type (VFileMessage | Error) */ (error_)
         const message =
           new VFileMessage(
-            `Cannot process MDX file with esbuild:\n  ${error_}`, {
-            cause,
-            place: 'reason' in cause ? cause.place : undefined,
+            `Cannot process MDX file with esbuild`, {
+            cause: /** @type (VFileMessage | Error) */ (error_),
+            place: error_ instanceof VFileMessage ? error_.place : undefined,
             ruleId: 'process-error',
             source: '@mdx-js/esbuild'
           })
@@ -198,12 +197,20 @@ function vfileMessageToEsbuild(state, message) {
     location.length = Math.min(location.length, maxLength)
   }
 
+  /** @type {Error} */
+  var exc = message
+  var text = message.reason
+  while (exc.cause instanceof Error) {
+    exc = exc.cause
+    text = `${text}:\n  ${exc}`
+  }
+
   return {
     detail: message,
     id: '',
     location,
     notes: [],
     pluginName: state.name,
-    text: message.reason
+    text,
   }
 }
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -118,14 +118,15 @@ export function esbuild(options) {
         messages = file.messages
       } catch (error_) {
         const cause = /** @type {VFileMessage | Error} */ (error_)
-        const message =
-          new VFileMessage(
-            'Cannot process MDX file with esbuild', {
+        const message = new VFileMessage(
+          'Cannot process MDX file with esbuild',
+          {
             cause,
             place: 'reason' in cause ? cause.place : undefined,
             ruleId: 'process-error',
             source: '@mdx-js/esbuild'
-          })
+          }
+        )
         message.fatal = true
         messages.push(message)
       }
@@ -159,14 +160,14 @@ export function esbuild(options) {
  */
 function vfileMessageToEsbuild(state, message) {
   /** @type {Location} */
-  let location = {
+  const location = {
     column: 0,
     file: state.path,
     length: 0,
     line: 0,
     lineText: '',
     namespace: 'file',
-    suggestion: '',
+    suggestion: ''
   }
 
   const place = message.place
@@ -179,9 +180,9 @@ function vfileMessageToEsbuild(state, message) {
     const end = place && 'end' in place ? place.end : undefined
     if (end) {
       if (start.offset !== undefined && end.offset !== undefined) {
-        location.length = end.offset - start.offset;
+        location.length = end.offset - start.offset
       } else if (end.line === start.line) {
-        location.length = end.column - start.column;
+        location.length = end.column - start.column
       }
     }
 
@@ -199,8 +200,8 @@ function vfileMessageToEsbuild(state, message) {
   }
 
   /** @type {Error} */
-  var exc = message
-  var text = message.reason
+  let exc = message
+  let text = message.reason
   while (exc.cause instanceof Error) {
     exc = exc.cause
     text = `${text}:\n  ${exc}`
@@ -212,6 +213,6 @@ function vfileMessageToEsbuild(state, message) {
     location,
     notes: [],
     pluginName: state.name,
-    text,
+    text
   }
 }
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -122,7 +122,7 @@ export function esbuild(options) {
           new VFileMessage(
             `Cannot process MDX file with esbuild:\n  ${error_}`, {
             cause,
-            place: ('reason' in cause && cause.place) || undefined,
+            place: 'reason' in cause ? cause.place : undefined,
             ruleId: 'process-error',
             source: '@mdx-js/esbuild'
           })
--- a/packages/esbuild/lib/index.js
+++ b/packages/esbuild/lib/index.js
@@ -117,10 +117,10 @@ export function esbuild(options) {
           '\n'
         messages = file.messages
       } catch (error_) {
-        const cause = /** @type (VFileMessage | Error) */ error_
+        const cause = /** @type {VFileMessage | Error} */ (error_)
         const message =
           new VFileMessage(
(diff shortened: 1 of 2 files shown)


packages/esbuild/lib/index.js
```diff
@@ -116,15 +117,15 @@ export function esbuild(options) {
...
         const message =
-          'reason' in cause
-            ? cause
-            : new VFileMessage('Cannot process MDX file with esbuild', {
-                cause,
-                ruleId: 'process-error',
-                source: '@mdx-js/esbuild'
-              })
+          new VFileMessage(
+            `Cannot process MDX file with esbuild:\n  ${error_}`, {
```
wooorm, line 123:
```suggestion
            'Cannot process MDX file with esbuild', {
```

The point of cause, I believe, is to be an alternative to copying it into the reason.
I’m fine with always making a new error, a vfile message.

packages/esbuild/lib/index.js
```diff
@@ -116,15 +117,15 @@ export function esbuild(options) {
...
+            cause,
+            place: ('reason' in cause && cause.place) || undefined,
```
wooorm, line 125:
```suggestion
            place: 'reason' in cause ? cause.place : undefined,
```

I prefer ternaries over of conditionals, which can fall through

packages/esbuild/test/index.js
```diff
@@ -365,7 +366,7 @@ test('@mdx-js/esbuild', async function (t) {
               file: 'test/esbuild.mdx',
               length: 11,
               line: 3,
-              lineText: '# Hello, <Message />',
+              lineText: '',  // Not available when place.offset is deleted
```
wooorm, line 369:
```suggestion
              lineText: '',
```

No comment needed, the code below for `6` is clear that the test is about missing offsets

packages/esbuild/test/index.js
```diff
@@ -414,7 +415,8 @@ test('@mdx-js/esbuild', async function (t) {
         file.message('3', tree)
         file.message('4', esm)
         file.message('5', text)
-        file.message('6', jsx)
+        const m6 = file.message('6', jsx)
+        if (m6.place && 'start' in m6.place) delete m6.place.start.offset
```
wooorm, line 419:
```suggestion
        assert(m6.place)
        assert('start' in m6.place)
        delete m6.place.start.offset
```

Use assertions so this doesn’t accidentally silently fall through and test something else

packages/esbuild/lib/index.js
egnor: Okay, I changed it to not concatenate the message at this point, and instead walk the cause chain and concatenate the text when creating the esbuild message.

In theory the cause chain is available in the esbuild message via the `detail` field, but, that's a very generic field and basically nothing seems to look at it-- certainly esbuild itself doesn't print it out when reporting errors.

egnor:
Ready for another look!

packages/esbuild/lib/index.js
```diff
@@ -157,44 +158,60 @@ export function esbuild(options) {
...
-      line,
-      lineText: state.doc.slice(lineStart, lineEnd),
-      namespace: 'file',
-      suggestion: ''
-    },
+    location,
     notes: [],
     pluginName: state.name,
-    text: message.reason
+    text,
```
wooorm, line 215: Please use the npm scripts in this package: `npm test` is the main one. It uses `npm run format`. That formats things: it removes that `,`

packages/esbuild/lib/index.js
```diff
@@ -157,44 +158,60 @@ export function esbuild(options) {
...
+
+    const maxLength = state.doc.length - (start.offset || 0)
+    location.length = Math.min(location.length, maxLength)
+  }
 
-  const match = eol.exec(state.doc)
-  const lineEnd = match ? match.index : state.doc.length
+  /** @type {Error} */
+  var exc = message
+  var text = message.reason
```
wooorm, line 203: Please use the npm scripts in this package: `npm test` is the main one. It uses `npm run lint`. That checks things: it should warn about `var` and abbreviations

packages/esbuild/lib/index.js
```diff
@@ -157,44 +158,60 @@ export function esbuild(options) {
...
+  while (exc.cause instanceof Error) {
```
wooorm, line 204: If possible, do not use `instanceof`, but use duck typing.

`instanceof` has problems across realms: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms

The *point* of `cause` is that *other* values can be used, not just errors.

For one way how I handle `cause` chains, see: https://github.com/vfile/vfile-reporter/blob/main/lib/index.js#L242

packages/esbuild/lib/index.js
```diff
@@ -157,44 +158,60 @@ export function esbuild(options) {
...
+    exc = exc.cause
+    text = `${text}:\n  ${exc}`
+  }
```
wooorm, line 207: Are these lines because otherwise the “cause” is lost? That ESBuild doesn’t print them?

Is that something that ESbuild should fix? Upstream?

Or, if indeed needed here, perhaps you could use `vfile-reporter` I linked above, or, perhaps you could use `inspect` from `util`: https://nodejs.org/api/util.html#utilinspectobject-options. I believe that it serializes recursive `cause`s

packages/esbuild/test/index.js
```diff
@@ -231,7 +231,8 @@ test('@mdx-js/esbuild', async function (t) {
         },
         notes: [],
         pluginName: '@mdx-js/esbuild',
-        text: 'Unexpected character `/` (U+002F) before local name, expected a character that can start a name, such as a letter, `$`, or `_` (note: to create a link in MDX, use `[text](url)`)'
+        text:
+          'Cannot process MDX file with esbuild:\n  1:12: Unexpected character `/` (U+002F) before local name, expected a character that can start a name, such as a letter, `$`, or `_` (note: to create a link in MDX, use `[text](url)`)'
```
wooorm, line 235: I think this would be formatted by `npm run format` too :)

packages/esbuild/lib/index.js
egnor: (done)

packages/esbuild/lib/index.js
egnor: (done)

packages/esbuild/lib/index.js
egnor: (done)

In fact I learned this the hard way myself recently. No more `instanceof` here.

packages/esbuild/test/index.js
egnor: (done)

packages/esbuild/lib/index.js
egnor: Okay yes this is the tricky bit.

Error reporting in esbuild doesn't involve exceptions at all, but their own `Message` class which gets used for both errors and warnings. That class includes a `detail` field which can be an exception but could also be anything else, and in any case doesn't get printed in esbuild's output, it seems to be for custom API use.

Alternatives
- use `vfile-reporter` -- this does a fine job and does chase cause chains, unfortunately this wants an entire `VFile`, I don't see a way to format a single `VFileMessage` (though you could imagine exposing that)?
- use `util.inspect` -- this has a lovely formatter for `Error` objects, but even for subclasses like `VFileMessage` reverts to its default printer which just dumps the fields and does not print error messages or chase cause chains?
- just let the error get thrown rather than catching it and adding it to `messages` -- this does get handled by esbuild gracefully (including plugin attribution), but the extra location information in `VFileMessage` is lost
- add `VFileMessage`-shaped exceptions directly to `messages` (no wrapping), re-throw everything else, remove all cause-chain-processing -- this could work??

For now I simplified the cause processing to not try to walk the whole cause chain but just include the string representation of the cause if one exists.

wooorm (approved):

packages/esbuild/lib/index.js
wooorm: Good points, I appreciate that you came up with alternatives & discussed trade offs.

I could see an `reportError` in `vfile-reporter`, which can then be used here.

The no-wrapping idea could also perhaps work.

But this is probably fine to start with!

wooorm:
Thanks! :)

PR #2629, opened 2025-08-07
Handle query parameters in Vite

markdalgleish:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This fixes an issue where MDX imports with query parameters aren't picked up by the Rollup plugin in Vite. To fix this, this PR uses the `extnamesToRegex` util to create a regular expression for matching MDX extensions, and this util already handles the existence of query parameters correctly.

I ran into this in the context of React Router where I'm working on RSC support for Framework Mode which uses Vite. We're using query parameters to transform client and server versions of route files, but this approach means that MDX routes are no longer working since the query parameters cause the MDX plugin to skip over them.

### DIFF
--- a/packages/rollup/lib/index.js
+++ b/packages/rollup/lib/index.js
@@ -32,8 +32,8 @@
  *   Callback called by Rollup and Vite to transform.
  * @param {string} value
  *   File contents.
- * @param {string} path
- *   File path.
+ * @param {string} id
+ *   Module ID.
  * @returns {Promise<SourceDescription | undefined>}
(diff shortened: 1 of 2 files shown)


wooorm:
Cool, thanks!

markdalgleish:
@wooorm Sorry to ping you again, but do you know when this will be released?

wooorm:
Released in [`3.1.1`](https://github.com/mdx-js/mdx/releases/tag/3.1.1)! Thanks for your patience!

PR #2661, opened 2025-11-17
Document augmentation to make JSX available

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Previously frameworks used to define a global `JSX` namespace. This is no longer the case. Now users need to define it themselves using module augmentation.


wooorm:
thanks! :)

PR #2664, opened 2025-12-05
Add `remark-directive-mdx` to list of plugins

re-xyr:
### Initial checklist

* [x] I read the support docs 

### Description of changes

[`remark-directive-mdx`](https://github.com/re-xyr/remark-directive-mdx) is a remark plugin that transforms Markdown directives (`:directive[]{}`) parsed by [`remark-directive`](https://github.com/remarkjs/remark-directive) into MDX JSX elements. This change adds this plugin to the plugins list in the docs.

### DIFF
--- a/docs/docs/extending-mdx.mdx
+++ b/docs/docs/extending-mdx.mdx
@@ -97,6 +97,8 @@ See also the [list of remark plugins][remark-plugins],
   — expose the page title as a string
 * [`boning-w/rehype-mdx-toc`](https://github.com/boning-w/rehype-mdx-toc)
   — export the table of contents data into MDX module
+* [`re-xyr/remark-directive-mdx`](https://github.com/re-xyr/remark-directive-mdx)
+  — transform Markdown directives (`:directive[]`) to JSX elements
 * [`pangelani/remark-mdx-chartjs`](https://github.com/pangelani/remark-mdx-chartjs)
   — replace fenced code blocks with charts using [`react-chartjs-2`](https://react-chartjs-2.js.org/).
 * [`remcohaszing/remark-mdx-frontmatter`](https://github.com/remcohaszing/remark-mdx-frontmatter)


remcohaszing (approved):
Thanks! This project looks really cool. I do have some tips.

- I believe you don’t need to define `_mdxExplicitJsx` in any of the places where you do.
- Don’t use `Object.assign()` to modify nodes. Instead, replace them like this:
  ```js
  visit(tree, ['containerDirective', 'leafDirective', 'textDirective'], (node, index, parent) => {
    if(index == null || parent == null) {
      return
    }

    // …

    parent.children[index] = newNode
  })
  ```
- Avoid type casting using `as`.
- I think you may get better help from TypeScript if you use object literals instead of `unist-builder`. I may be wrong though.
- You perform some logic to convert directive attributes, interpreted as HTML attributes, to JSX attributes. Your approach is a bit naive. This makes sense, as `class` → `className` is a well-known example, but there’s more. You may be interested to have a look at or use [`hast-util-properties-to-mdx-jsx-attributes`](https://github.com/remcohaszing/hast-util-properties-to-mdx-jsx-attributes). On the other hand, users write those attributes without HTML. So maybe you don’t want to perform any conversion at all.
- If you map over an array, it’s generally better to specify the generic type on the `.map` method or the return type of the mapper function. This provides better TypeScript support.
- In `astroHandleLabel` you generate a `<Fragment />` element. However, you don’t insert a `Fragment` import.
- In your tests you build the AST manually and test on that. This way it’s fairly easy to make mistakes. I recommend testing your plugin with `compile()` and testing it against a string input as an actual user would use it.

wooorm:
> remarkMarkAndUnravel did this for all "explicit JSX elements", and it was called before external remark plugins. Maybe we can debate whether directives count as "explicit", or I missed something.

Depends what you want! If a user passes `h1: MyComponent`, I think they probably want to overwrite `# this` and not `<h1>this</h1>` or `::h1[this]`

> It would cause the old subtree to be traversed instead of the new one. Indeed, if I try to replace the node in the parent outright by parent!.children[index!] = newNode, the test suite fails:

Is this a good case for `[SKIP, currentIndex]`? Will the function be called again for the transformed node, ignore the old children, and then call for the new children?

> By default, the plugin performs no conversion on attributes; I just exposed a simple function that users can pass to the plugin to instruct it to do some basic transformations. If you think the way that function behaves is wrong or misleading, please elaborate and I'll mark the function as deprecated.

Generally I think there should be no conversion for things that users write themselves, and that MDX already has options for conversion for the stuff that users do not write.

wooorm:
Thanks!

PR #2680, opened 2026-05-29
Support Vite raw and url query parameters

### LINKED ISSUE #2657: Cannot use query parameters in Vite `import.meta.glob`
### Initial checklist

- [x] I read the [support docs](https://mdxjs.com/community/support/)
- [x] I read the [contributing guide](https://mdxjs.com/community/contribute/)
- [x] I agree to follow the [code of conduct](https://github.com/mdx-js/.github/blob/main/code-of-conduct.md)
- [x] I searched [issues](https://github.com/search?q=user%3Amdx-js&type=issues) and [discussions](https://github.com/orgs/mdx-js/discussions) and couldn’t find anything (or linked relevant results below)

### Affected package

@mdx-js/rollup@3.1.1

### Steps to reproduce

https://github.com/joshpensky/mdx-js-rollup-vite-glob-repro

1. Create a new Vite 7 project with `@mdx-js/rollup@3.1.1`
2. Define your `vite.config.ts` file with the `mdx` plugin
3. Create a few mdx files, then import with a raw query:
```
const docs = import.meta.glob("./docs/*.mdx", {
  query: "?raw",
  import: "default",
  eager: true,
});
```

### Actual behavior

The object values in the returned glob object are function components

### Expected behavior

The object values should be the raw strings, which is typical behavior for the `?raw` query param in Vite

### Runtime

node@24.11.0

### Package manager

pnpm@10.21.0

### Operating system

macOS Sequoia 15.3.2

### Build and bundle tools

Vite

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Vite has special treatment for the `raw` and `url` query parameters. Unfortunately, there’s not dedicated API to detect which file ID gets special treatment, and which does not. So we have to hard code it.

Closes #2657
Refs https://github.com/vitejs/vite/issues/22417

### DIFF
--- a/packages/rollup/lib/index.js
+++ b/packages/rollup/lib/index.js
@@ -90,6 +90,8 @@ export function rollup(options) {
 
       const [path, query] = id.split('?')
 
+      // Special case for Vite
+      // https://github.com/vitejs/vite/issues/22417
       if (query === 'raw' || query === 'url') {
         return
       }
--- a/packages/rollup/lib/index.js
+++ b/packages/rollup/lib/index.js
@@ -88,7 +88,12 @@ export function rollup(options) {
         })
       }
 
-      const [path] = id.split('?')
+      const [path, query] = id.split('?')
+
+      if (query === 'raw' || query === 'url') {
+        return
+      }
+
       const file = new VFile({path, value})
 
       if (
--- a/packages/rollup/lib/index.js
+++ b/packages/rollup/lib/index.js
@@ -90,8 +90,8 @@ export function rollup(options) {
 
       const [path, query] = id.split('?')
 
-      // Special case for Vite
-      // https://github.com/vitejs/vite/issues/22417
+      // Special case for Vite.
+      // <https://github.com/vitejs/vite/issues/22417>
       if (query === 'raw' || query === 'url') {
         return
       }
--- a/packages/rollup/test/index.js
+++ b/packages/rollup/test/index.js
@@ -124,4 +124,50 @@ test('@mdx-js/rollup', async function (t) {
     assert.match(code, /Hello Vite/)
     assert.match(code, /jsxs?\(/)
   })
+
+  await t.test('should support the ?raw query', async () => {
+    const result = /** @type {Array<RollupOutput>} */ (
+      await build({
+        build: {
+          lib: {
+            entry:
+              fileURLToPath(new URL('vite-entry.mdx?raw', import.meta.url)) +
+              '?raw',
+            name: 'query'
+          },
+          write: false
+        },
+        logLevel: 'silent',
+        plugins: [rollupMdx()]
+      })
+    )
+
+    const code = result[0].output[0].code
+
+    assert.match(code, /# Hello Vite/)
+    assert.doesNotMatch(code, /jsxs/)
+  })
+
+  await t.test('should support the ?url query', async () => {
+    const result = /** @type {Array<RollupOutput>} */ (
+      await build({
+        build: {
+          lib: {
+            entry:
+              fileURLToPath(new URL('vite-entry.mdx?url', import.meta.url)) +
+              '?url',
+            name: 'query'
+          },
+          write: false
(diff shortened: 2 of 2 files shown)


wooorm:
> Vite

What about non-vite rollup users?

---

What are the effects of this? (No tags applied). Breaking?

Remco, your last comment in the related issue is that you are not sure what to do.
Did that change? The Vite issue 22417 seems unresolved?

packages/rollup/lib/index.js
```diff
@@ -88,7 +88,12 @@ export function rollup(options) {
         })
       }
 
-      const [path] = id.split('?')
+      const [path, query] = id.split('?')
+
+      if (query === 'raw' || query === 'url') {
+        return
+      }
```
wooorm, line 97: please reference why this exist, with a URL to an issue or docs in comments or so, explaining why this seemingly magic behavior exists!

remcohaszing:
This is based on the discussion in https://github.com/vitejs/vite/issues/22417.

https://github.com/mdx-js/mdx/pull/2629 was a breaking change. Unfortunately, that means a breaking change was released in a patch version, as some users noticed. This change solves that real use case (https://github.com/mdx-js/mdx/issues/2657).

Unfortunately, Neither Rollup nor Vite has a proper API to deal with builtin query parameters. Reverting https://github.com/mdx-js/mdx/pull/2629 now would break MDX for React Router users, which is also not really bad. This PR is the best way to handle it.

I just tried and Rollup doesn’t support the `?raw` query param. It’s unaffected.

packages/rollup/test/index.js
```diff
@@ -124,4 +124,50 @@ test('@mdx-js/rollup', async function (t) {
     assert.match(code, /Hello Vite/)
     assert.match(code, /jsxs?\(/)
   })
+
+  await t.test('should support the ?raw query', async () => {
```
wooorm:
```suggestion
  await t.test('should support the `?raw` query', async () => {
```

packages/rollup/lib/index.js
```diff
@@ -88,7 +88,14 @@ export function rollup(options) {
         })
       }
 
-      const [path] = id.split('?')
+      const [path, query] = id.split('?')
+
+      // Special case for Vite
+      // https://github.com/vitejs/vite/issues/22417
```
wooorm:
```suggestion
      // Special case for Vite.
      // <https://github.com/vitejs/vite/issues/22417>
```

packages/rollup/test/index.js
```diff
@@ -124,4 +124,50 @@ test('@mdx-js/rollup', async function (t) {
...
+    const result = /** @type {Array<RollupOutput>} */ (
+      await build({
+        build: {
+          lib: {
+            entry:
+              fileURLToPath(new URL('vite-entry.mdx?raw', import.meta.url)) +
+              '?raw',
```
wooorm, line 135: Wait, this is adding `?raw` twice? In the param to `URL(x)`, and then concatenating it on again after `+`?

packages/rollup/test/index.js
```diff
@@ -124,4 +124,50 @@ test('@mdx-js/rollup', async function (t) {
...
+  })
+
+  await t.test('should support the ?url query', async () => {
+    const result = /** @type {Array<RollupOutput>} */ (
+      await build({
+        build: {
+          lib: {
+            entry:
+              fileURLToPath(new URL('vite-entry.mdx?url', import.meta.url)) +
+              '?url',
```
wooorm, line 158: Same?

wooorm (commented):
It looks like you made a mistake?
Then two minor nits.
Feel free to squash, merge, and release after!

packages/rollup/test/index.js
remcohaszing: The first query parameter is removed in the conversion from file URL to a path. So it’s useless. Nice catch.

PR #2681, opened 2026-05-30
Use unist-util-mdx-define

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This library abstracts away some of the logic needed to define a variable in mdast/hast/estree. In this case it saves us a small bit of boilerplate. We already have this dev dependency in our `node_modules`.


website/mdx-loader.js
```diff
@@ -238,23 +214,12 @@ function recmaInjectMeta(options) {
...
-      source: undefined,
-      attributes: [],
-      specifiers: []
-    })
+    define(
+      tree,
+      file,
+      {meta: valueToEstree(value, {instanceAsObject: true})},
+      {conflict: 'skip'}
+    )
```
wooorm, line 222: Nice, fun, looks clean!

This is just the website, local scripts. Have you considered it in other places?

One thing I have wondered before, related to that, is what to do with conflicts.
More for the internal `_helpers` that are injected in MDX content: what if a human writes a `_createMdxContent` themselves for some reason.

The “conflict resolve” mode would then likely be: give me the actual `_helper1` id back or so, which the user of `mdxDefine` then has to apply to every place where they wanted to reference `_helper`?

wooorm (approved):

PR #2682, opened 2026-05-31
Turn vfile messages into Rollup logs

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

During compilation, plugins may log vfile messages onto the vfile. Some integrations already handle these. The Rollup plugin did not. This change makes the Rollup plugin handle messages.

Fatal messages are unhandled on purpose. They are supposed to be thrown. Even if they are not thrown, Rollup’s `this.error()` throws. So instead, those are treated as warnings.


packages/rollup/lib/index.js
```diff
@@ -71,6 +71,47 @@ export function rollup(options) {
...
+            }
+
+            log.pluginCode = pluginCode
+          }
+
+          if (message.fatal === undefined || message.fatal === null) {
+            this.info(log)
+          } else {
+            this.warn(log)
+          }
```
wooorm, line 88: - `message: message.reason` probably?
- separate function?
- good to have a link to the rollup docs for `log`?
- have you considered other fields?

remcohaszing: I processed all feedback, here’s the link to the rollup docs: https://rollupjs.org/plugin-development/#this-warn

I have considered other fields. The other fields are more related to Rollup’s own API. It augments the logs. I did deliberately add `cause`, so that the vfile message is accessible from the programmatic API for more advanced users.

packages/rollup/lib/index.js
```diff
@@ -30,6 +31,50 @@ import {createFilter} from '@rollup/pluginutils'
...
+      }
+
+      pluginCode += message.ruleId
+    }
+
+    log.pluginCode = pluginCode
+  }
+
+  return log
+}
```
wooorm: util functions at the bottom?

wooorm (commented):
one minor nit, otherwise :+1:, feel free to squash and merge whenever!

PR #2683, opened 2026-05-31
Log vfile messages in Webpack loader

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Some of our integrations turn vfile messages into log messages for the tool they integrate with. This was missing from the Webpack loader. This change logs vfile messages in the Webpack loader.

For more information on Webpack logger severity, see https://webpack.js.org/api/logging/#logger-methods.

The Webpack logger allows you to log arbritrary values. These get stringified. However, this is extremely verbose. Instead, I decided to represent messaged with single-line strings. The logs contain a full file path with line number. I find this useful, because some terminal editors make this format clickable.

The following screenshot shows that it looks like

Webpack output with a warning and an info messages for 2 MDX files.

### DIFF
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -85,9 +85,42 @@ export function loader(value, callback) {
 
   const context = this.context
   const filePath = this.resourcePath
+  const logger = this.getLogger()
 
   process({value, path: filePath}).then(
     function (file) {
+      for (const message of file.messages) {
+        let log = filePath
+        if (message.line) {
+          log += ':' + message.line
+
+          if (message.column) {
+            log += ':' + message.column
+          }
+        }
+
+        log += ' '
+        if (message.source) {
+          log += message.source
+
+          if (message.ruleId) {
+            log += ':' + message.ruleId
+          }
+
+          log += ' '
+        } else if (message.ruleId) {
+          log += message.ruleId + ' '
+        }
+
+        log += message.message
+
+        if (message.fatal === undefined || message.fatal === null) {
+          logger.info(log)
+        } else {
+          logger.warn(log)
+        }
+      }
+
       callback(
         undefined,
         Buffer.from(file.value),
--- a/packages/loader/lib/index.js
+++ b/packages/loader/lib/index.js
@@ -90,34 +90,10 @@ export function loader(value, callback) {
   process({value, path: filePath}).then(
     function (file) {
       for (const message of file.messages) {
-        let log = filePath
-        if (message.line) {
-          log += ':' + message.line
-
-          if (message.column) {
-            log += ':' + message.column
-          }
-        }
-
-        log += ' '
-        if (message.source) {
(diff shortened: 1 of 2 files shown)


packages/loader/lib/index.js
```diff
@@ -85,9 +85,42 @@ export function loader(value, callback) {
...
+        }
+
+        log += message.message
+
+        if (message.fatal === undefined || message.fatal === null) {
+          logger.info(log)
+        } else {
+          logger.warn(log)
+        }
+      }
```
wooorm, line 98: Somewhat important:

- Maybe a separate function, to build a message?

More options:

- other than positional info (maybe also start/end), source/ruleId, maybe `cause` can be added? https://github.com/mdx-js/mdx/blob/c20db6abebf36a34e247283b1d55979ea5e8b98e/packages/esbuild/lib/index.js#L203-L206
- have you considered using `reason` instead of `message`? I think that’s the normal field in `VFileMessage`, IIRC `message` has a little more things (because it’s an `Error`)

There could of course be more (https://github.com/vfile/vfile-reporter/blob/270c90b69c836ed03aab5e49fc3fe93d372fe2cd/lib/index.js#L363), and that can be your answer to the above too, but things can be added too.

remcohaszing: From https://webpack.js.org/api/logging/:

> [!Warning]
> **Avoid noise in the log!** Keep in mind that multiple plugins and loaders are used together. Loaders are usually processing multiple files and are invoked for every file. Choose a logging level as low as possible to keep the log output informative.

So I decided to keep the logs minimal, but informative, on a single line.

The source and rule id are already logged.

It would be possible to log the cause, note, and url at a more verbose severity.

The `message` and `reason` properties of `VFileMessage` are assigned the same value, but I can use `reason` if your prefer.

wooorm: Yes, I think `reason` is better; that’s the one chosen for vfile-message; `message` is from `Error` / for `Error` compatibility

https://github.com/vfile/vfile-message/blob/2c782c8580c5b0a14baae19fa8badc628e567461/lib/index.js#L234

As for their warning: understandable, but I don’t think MDX should make that choice. I think plugins need to make that choice.

remcohaszing: > As for their warning: understandable, but I don’t think MDX should make that choice. I think plugins need to make that choice.

I don’t understand. This loader is where we do make the choice what to display to the user and how to format it. It’s up to the plugin what information to provide.

wooorm: In short:
a user can intentionally use a plugin that provides a lot of info in logs.
Neither MDX nor Webpack should remove all that important info.
Only if there are options

remcohaszing: Yes, the option is the logger verbosity. We can log other fields at a different verbosity.

PR #2695, opened 2026-09-24
Remove vm2 security recommendation

EoinTrial:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Removed the `vm2` recommendation from the security guidance in the getting started documentation.

`vm2` has a history of critical sandbox-escape vulnerabilities, including vulnerabilities that allow code running inside the sandbox to reach the host environment. Removing the recommendation avoids suggesting an in-process JavaScript sandbox as a sufficient security boundary for untrusted MDX.

The existing guidance instead recommends OS-level isolation using Docker or similar, along with rate limiting and process timeouts.

See [[CVE-2026-26332](https://github.com/advisories/GHSA-55hx-c926-fr95)](https://github.com/advisories/GHSA-55hx-c926-fr95).

`npm test` passes.


docs/docs/getting-started.mdx
```diff
@@ -207,8 +207,6 @@ If you do,
 you might want to look into using `<iframe>`s with `sandbox`,
 but security is hard,
 and that doesn’t seem to be 100%.
-For Node.js,
-[vm2][] sounds interesting.
 But you should probably also sandbox the whole OS using Docker or similar,
 perform rate limiting,
 and make sure processes can be killed when taking too long.
```
remcohaszing: I have no strong opinion on the suggestion of `vm2`. But if we do remove it, we should rewrite the sentence after it a bit.


```suggestion
For Node.js you should probably sandbox the whole OS using Docker or similar,
perform rate limiting,
and make sure processes can be killed when taking too long.
```

remcohaszing (changes requested):

docs/docs/getting-started.mdx
EoinTrial: @remcohaszing Thanks for taking a look!

ChristianMurphy:
> Removing the recommendation avoids suggesting an in-process JavaScript sandbox as a sufficient security boundary for untrusted MDX.

The purpose of this recommendation is to make a _very_ insecure approach many devs take, less insecure.
Yes, there are known sandbox escapes, but that feels less risky than no sandbox at all.
Do you have a concrete alternative you'd recommend instead?

